PR #6, opened 2020-06-01
Feature/chilgenb crt hit alg update

chilge:
some bug fixes in the CRTHit reco algo, improvements in the analyzer modules, added charge saturation effect to CRTDetSim (enabled by default)


SFBayLaser:
Including for LArSoft v08_54_00 release

PR #13, opened 2020-06-12
[wirecell] move experiment related configuration into icaruscode

wenqiang-gu:
I reorganized the wirecell configuration in several aspects.
1) Changed ups/product_deps, so that now it will install some icarus-related wirecell configuration in the `icaruscode` area, instead of `icarus_data`.
2) Now the lifetime and diffusion constants are loaded from the larsoft services. Slightly update the fhicl file: `wcls-multitpc-sim-drift-simchannel.fcl`, the usage doesn't change.


wenqiang-gu:
Hi @SFBayLaser, once this PR is merged. Could you help remove a directory from `icarus_data`? I moved some files into `icaruscode`, so I don't need to frequently change

PR #19, opened 2020-06-29
calorimetry module with sumadc, mixed deconv+raw

varanini:


varanini:
> In the updates to reco_icarus.fcl, did you mean to return the rawdigitfilter and deconvolution before "mixedHitTPCn"? If this is part of the normal reco1 to reco2 chain then those data products are meant to exist already, right? I am not sure what happens to things like associations if you add new data products to collection with the same names as before... plus, of course, adding yet another raw digit collection has implications for memory and disk usage.



> In the updates to reco_icarus.fcl, did you mean to return the rawdigitfilter and deconvolution before "mixedHitTPCn"? If this is part of the normal reco1 to reco2 chain then those data products are meant to exist already, right? I am not sure what happens to things like associations if you add new data products to collection with the same names as before... plus, of course, adding yet another raw digit collection has implications for memory and disk usage.

as I said on slack, we don't have to run it in the default chain. We can just keep it available there.
Moreover, i don't see where data products names are duplicated.
I am also changing the liftime to 10ms, please wait a moment while I do that

varanini:
> The definition of, for example, "icarus_reco_SigProcTPC0Mixed" includes "rawDigitFilterTPC0" and "decon1DroiTPC0" which are previously used when running the gaushit finder. So these modules will create duplicate data products with the same producer name. If run in a different stage then they will have a different process name - if you are using fully qualified identifiers looking up data that will be ok... but the question is still whether this is really what you intend here...

OK I canceled the replicas of rawdiigtfilter and decon1droi. icarus_reco_SigProcTPC0Mixed now only includes the hitfinding and is meant to run after reco1. Should be ok.

I will commit in a moment

PR #34, opened 2020-07-20
Feature/usher decodercleanup

SFBayLaser:
Decoder module cleanup primarily to remove/reduce the diagnositc output - have put this as fhicl controllable in case we need to do more looking at channel mapping


SFBayLaser:
Merging to start release

PR #46, opened 2020-08-20
Feature/howard pandora modular shower creation

brucehoward-physics:
changes in preparation for the move of shower reconstruction from LArReco TRACS to LArPandora PandoraModularShowerCreation


SFBayLaser:
Ok, will merge into develop for icaruscode v08_62_01

brucehoward-physics:
It's not clear to me if the PandoraModularShowerCreation has actually been merged into larsoft develop yet, do you know if it has?

I knew Ed Tyley had pull requests in, but am not sure the status... I can check if you're not sure though.

SFBayLaser:
Hmm... I do not know... ok, maybe we hold off just one more release?

SFBayLaser:
Ok, I checked the larsoft release notes and it is claimed that this was done several releases ago so you are back on the front burner.

PR #61, opened 2020-09-29
Updated CRT geometry sorting (and mapping)

PetrilloAtWork:
The adoption of the split wire geometry channel mapping has caused the auxiliary detector order in ICARUS to change.
The order was customised for ICARUS, but the new channel mapping included a new sorting which used the "standard" LArSoft sorting for `AuxDetGeo` and `AuxDetSensitiveGeo` objects (i.e. CRT modules and submodules).
This commit moves the sorting code of PMT and of CRT out of the geometry sorters, and makes both sorters call the old CRT sorting so refactored (the new sorter also calls the PMT new sorting algorithm).

### DIFF
--- a/icaruscode/Geometry/GeoObjectSorterICARUS.cxx
+++ b/icaruscode/Geometry/GeoObjectSorterICARUS.cxx
@@ -7,6 +7,7 @@
 ////////////////////////////////////////////////////////////////////////
 
 #include "icaruscode/Geometry/GeoObjectSorterICARUS.h"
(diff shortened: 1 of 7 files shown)


PetrilloAtWork:
**This pull request needs to be approved by @chilge** before merging.

chilge:
I have tested @PetrilloAtWork 's fix, and it seems to work. I approve :)

chilge (approved):

PR #64, opened 2020-10-04
Updated all default geometries to use split wires.

PetrilloAtWork:
This affects only configurations using the overrides
`use_[no]overburden_geometry.fcl` or similar.
The general default geometry was split wires already (and no overburden).


SFBayLaser:
I have by hand merged this into icarusalg so am going to close this PR

PR #86, opened 2020-10-19
Update anaraw_purity_icarus_DQM_multipletpc.fcl

wesketchum:
Update purity DQM files to run over both cryostats.


SFBayLaser:
Sorry, we had a discussion on this PR via slack yesterday but I think I'm not sure the conclusion. Should this PR be included as is or was there going to be an update?

wesketchum:
It is ok as is, but would be better if there’s an update … I’ll try to get that settled soon…

PR #89, opened 2020-10-22
sbndaq_artdaq_core direct dep removal, plus some c7 compile fixes

wesketchum:
Once the new larsoft release (and new sbncode release) comes out, we should have sbndaq_artaq_core build separately and setup via sbncode --> can remove the dependency in icaruscode.


wesketchum:
Related pull is https://github.com/SBNSoftware/sbncode/pull/34

SFBayLaser (commented):
Looks good, merging

PR #106, opened 2020-11-17
ICARUS fcl files for Simple Flash Matching

ikatza:
I'm migrating the Simple Flash Matching code to `sbncode`, in doing so removing all dependencies to `sbndcode`.

The fcl files under FlashMatch folder control the parameters for ICARUS.

This should be merged along with
https://github.com/SBNSoftware/sbncode/pull/59


SFBayLaser:
Any code that is meant to run needs to be in the "PMT" folder of icaruscode (meaning a subfolder of that folder). Thank you for moving.

SFBayLaser:
Has this PR and code been presented/reviewed to the ICARUS PMT group?

ikatza:
Hi Tracy,

Just inside PMT will do? I should mention that the module also uses pandora spacepoints, so I was a little reluctant to put it inside PMT. 

These fcl run the flash matcher code that was used for last year PAC. I have made presentations in the past, but not specifically on the ICARUS PMT group. I haven't presented anywhere on this last update.

SFBayLaser:
Yes, please move to PMT folder unless these are actually for running a job (i.e. not just includes). If for running then need to be added to the appropriate section of the fcl folder. It would be good to try to keep some form of organization rather than just adding more and more folders at the top level (a losing battle but one has to try to keep things orderly!).

brucehoward-physics:
@SFBayLaser By "not just includes" do you mean the difference between a fcl that is defining the module and the job fcl itself? In that case I think Iker has one of each.

Though @ikatza is there anything that ties the module definition fcl (flashmatch_icarus.fcl) to `icaruscode` specifically? I see a reference to `FlashMatch/fm_metrics_icarus.root` but that isn't in `icaruscode`. So possibly this module could live in `sbncode` too, with only the job fcl in `icaruscode`?

(but in any case, maybe that's something that could be figured out more long-term in the interest of having something sooner?)

ikatza:
I felt it was more natural to have the fcls on each experiment's main repositories so that users could modify these if they needed without having to pull `sbncode`. 

I believe I lean more towards making a subfolder under `icaruscode/fcl/reco`.

ikatza:
By the way I have now produced up to date metrics with the latest code. I've pushed a branch to `icarus_data` that will need to be merged as well with this PR:
https://cdcvs.fnal.gov/redmine/projects/icarus_data/repository?utf8=✓&rev=feature%2Ficaza_fm_move

brucehoward-physics:
Wanted to check in to see if this can be/is going to be merged? Thanks for any updates.

(Haven't looked at Iker's `icarus_data` feature branch but note to @SFBayLaser that he committed files there)

SFBayLaser:
I did merge these fhicl files but moved them around to put in the correct location... the "definition" file goes into the fcl configuration folder, the driver goes into a subfolder of fcl/reco.

Issue #113, opened 2021-01-07
decoder making negative waveform values


wesketchum:
https://github.com/SBNSoftware/icaruscode/blob/d507d0cdb54c64e99634b43967a0f0e89730fe1d/icaruscode/Decode/DecoderTools/TPCDecoderFilter1D_tool.cc#L404

There's a minus sign in the front that I think should not be there ... @SFBayLaser, can you confirm that's likely a bug?

Also, is there any reason not to use the defined function in sbndaq-artdaq-core:
https://github.com/SBNSoftware/sbndaq-artdaq-core/blob/d9058bf3185bd8f15803a53f64fba332bbc39d6e/sbndaq-artdaq-core/Overlays/ICARUS/PhysCrateFragment.cc#L68

Should all be the same of course, but good to keep things in one place? But maybe there's a multi-threading issue anticipated here?

I'll probably make a quick PR to attack to this.

SFBayLaser:
It turns out the waveforms coming from the electronics are inverted with respect to what the deconvolution and signal processing is expecting. The negative sign is intended to flip the polarity and then the (inverted) pedestal is subtracted from the waveform to center it at zero ADC counts. 
If you don't have the minus sign then you get waveforms on the collection plane that look like; 
![image](https://user-images.githubusercontent.com/10157034/103926522-e2e35880-50cd-11eb-96a8-40ae923ad39b.png)

As to the adc_val function... I'm not sure I knew it existed? It seems the only real difference is the bit mask, is that necessary?

SFBayLaser:
I think this has been resolved?

wesketchum:
The bit mask prevents potential issues about int vs. uint, but this may indeed not be necessary.

I think we can close this? Sorry, I'm actually not sure then what we see ... @cfarnese , this explains the negative baselines, but does this properly explain the negative going signals on the collection plan as you understand it?

SFBayLaser:
closing...

PR #114, opened 2021-01-07
fix negative TPC waveform bug, but also use adc_val function from Overlay class

wesketchum:
fix to issue #113


SFBayLaser:
Closing this as I think the problem is understood/resolved...

PR #117, opened 2021-01-28
Update to sliging window trigger simulation

PetrilloAtWork:
This set of commits supports trigger based on sliding windows, with only selected winodws enabled.
It may be used to simulate triggers on only the central part of the detector, or only a single side.

Also a bug fix is included for overlapping windows (which were supposed to be supported, but were not).


PetrilloAtWork:
Pinging @rhowell42 for his information.

PR #139, opened 2021-03-25
added opflash filter and opflash config for reconstructing opflash to be used w/ the filter

drinkingkazu:


SFBayLaser:
Merging now for testing

PR #140, opened 2021-03-26
Prepare for full multiple TPC readout at full rate using a PMT filter

SFBayLaser:
The main objective of this PR is to prepare for running full readout in multiple TPC mode at full rate using a PMT based filter. The stage 0 and stage 1 fhicl files have undergone substantial overhaul to be able to handle the case of "single TPC readout" (all of the TPC daq data in a single data product) vs "multi TPC readout" (the TPC daq data split by module). 
In addition we are now employing the PMT filter added by Kazu/Andrea in a previous PR into the workflow in order to reduce overall processing time.


SFBayLaser:
Ok, merge time!

Issue #147, opened 2021-04-09
Avoid art jobs with duplicate events


PetrilloAtWork:
I have received the n-th request for support that turned out to be caused by duplicate events, so I am triggering this.

## Definition: duplicate event

At the state of the _art_, a duplicate event is an event with the same ID (run, subrun and event number) as another event already processed.

## Duplicate events and _art_

The _art_ framework requires that a job process no duplicate events.
If that requirement is not met, several obscure consequences have been observed:

1. mismatch in tools like `TimeTracker` between the number of events processed by a module (including duplicates) and the total number of events (including one count for each unique event)
2. on processing an output file which merged two duplicate events, exceptions of less-than-trivial interpretation

#### How can duplicate events occur

The most common cause is the processing of two files at the end of a chain, which shared a parent file. This may happen because a file was processed by a job, the process was deemed to have failed, so the file was processed again by another "recovery" job. If the first job did actually not fail and the output files are just swept from a dropbox directory, at sweeping time both output files from the same input are collected and enter the final dataset.

Another cause may be an input file list with duplicate entries.

A third one is the mixing of different independent jobs. For example: I generate 50 events with `simulation_genie_bnb_icarus.fcl`,  which generates run 1, subrun 1, events 1 to 50, and then I run `standard_g4_icarus.fcl` on it and realise statistics is not enough. Then I run again `simulation_genie_bnb_icarus.fcl` to generate 50 more events, which will again be called run 1, subrun 1, events 1 to 50. Then I run `G4` on the two 50-event sample (and typically get weird error only at *`DetSim`* stage).

# The solution: no solution.

Given that this kind of mistakes are unavoidable, I point to prompt and clear feedback.
I suggest the service [`DuplicateEventTracker`](https://icarus-exp.fnal.gov/at_work/software/doc/icaruscode/latest/classsbn_1_1DuplicateEventTracker.html#details) to be added to the default service configuration (in the most basic table) with a configuration that terminates the job as soon as it discovers a duplicate event.

PR #148, opened 2021-04-09
Added DuplicateEventTracker to default services

PetrilloAtWork:
This is a proposed solution for issue #147.
It adds `DuplicateEventTracker` service to the `icarus_art_services`, which is included in any ICARUS configuration even vaguely official. The configuration of the service is such that an exception is thrown on the first duplicate event encountered.

This is a GitHub commit (i.e. made with GitHub web editor).

### DIFF
--- a/fcl/services/services_common_icarus.fcl
+++ b/fcl/services/services_common_icarus.fcl
@@ -81,6 +81,8 @@ icarus_art_services: {
     }
   }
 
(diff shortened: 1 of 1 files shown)


wesketchum (commented):
Without testing it at all, seems ok by me ... I'll see if @mayawos or @SFBayLaser want to approve.

Also, agree having it by default seems good.

Issue #151, opened 2021-04-12
PMT readout fragment ID change


PetrilloAtWork:
Recently the ID of the fragments from PMT readout boards (V1730) has changed (from `N` to `0x2000 + N`).
User code needs to mask the additional bits to access the PMT channel mapping database, which retains the old fragment ID values.

The standard interface to access that PMT channel mapping database from LArSoft environment is the service family `IICARUSChannelMap`, some methods of which accept a fragment ID as argument.
I suggest that this service automatically transform the PMT fragment ID, lifting the users from being aware of that implementation detail.

PetrilloAtWork:
This is a request for comments on the idea. Among the people involved: @wesketchum, @SFBayLaser, @ascarpel.

wesketchum:
This is resolved? If so, @PetrilloAtWork, @SFBayLaser , @ascarpel  or someone who can confirm, please close?

PetrilloAtWork:
I don't think it has been addressed.
I still see no masking in `ICARUSChannelMapProvider.cxx`.
@ascarpel, @animeshgit20, @biswaranjan91 , what is the fragment ID key in the PMT channel mapping database now? pattern `0x20xx` or `0x00xx`?

wesketchum:
@PetrilloAtWork @ascarpel @animeshgit20 @biswaranjan91 I think at this point this has been fixed? If so, let's close this out?

biswaranjan91:
FragmentID for PMT in the database is an integer from 0 to 23. 
Still it has not in the  pattern of 0x20xx or 0x00xx.
From my recollection, PMT decoder level it has changed, however we have not changed in the database level. May be Andrea(@ascarpel) can help on confirming this. I can check but not sure which fhicl file was used officially for PMT decoder.

PetrilloAtWork:
The issue was to elicit comments on encoding the database key (0-23) in the interface to the database.
Neither this has been implemented, nor opinions have been offered. So no, not done yet.

PetrilloAtWork:
I have opened a pull request with a proposed implementation.
Let's see if this can push the issue toward its conclusion.

wesketchum:
PR has been merged, so claiming this is done.

PR #152, opened 2021-04-12
Updated channel mapping access of PMT configuration extractor

PetrilloAtWork:
The fragment ID needs now to be masked.
This fix is not needed if the suggestion in issue #151 is adopted (but since I need this to work _now_ I am pushing it anyway).


SFBayLaser:
I am going to merge this now because I am not completely sure I understand issue #151 (and someone requested a release sooner than later)... So we will revisit...

PR #163, opened 2021-05-12
PMT decoding timestamps

PetrilloAtWork:
PMT decoding tool has been reworked to add a proper timestamp to the optical waveforms.
The timestamp should be compatible with the TPC timescale, although some tuning may still be needed.
Included so far is a small delay from the trigger signal to the PMT readout board trigger (which **does** need calibration) and the pre-trigger buffer size.

The tool coding style has been updated a lot, including the addition of extensive technical documentation in Doxygen format and the switch to configuration validation.

As brand feature of mine, the code is probably more complex that it needs to.

# Breaking changes

There are several, and of different severity.

## Breaking change: online monitor

Seriously. **The standard configuration for this tool will not work in online monitor environment.** That is because it needs information (i.e. the FHiCL configuration of the DAQ) that is not provided in the online monitor workflow. A special configuration is required for the online monitor to skip the features that require such information. I believe the set of parameters to the tool should be:
    
    physics.producers.daqPMT.DecoderTool.PMTconfigTag: @erase
    physics.producers.daqPMT.DecoderTool.RequireKnownBoards: false
    physics.producers.daqPMT.DecoderTool.RequireBoardConfig: false
    
(hard to say exactly since I don't have input files without FHiCL configuration in).

## Breaking change: decoding of special runs

The new configuration of the PMT decoding tool in `stage0_multiTPC_icarus.fcl` **requires** all PMT readout boards to be configured in the run. To decode runs with only a partial set of PMT readout boards requires a change in the configuration skipping the check therein:
    
    physics.producers.daqPMT.DecoderTool.RequireBoardConfig: false
    
(see e.g. `decodePMT_icarus_incomplete.fcl`).

## Breaking change: time stamps

This is the due to the new feature: previous tuning of algorithms based on the misalignment of TPC and PMT won't work any more (the misalignment should be ~much smaller~ gone).


PetrilloAtWork:
For @ascarpel: I would like you to take a look and possibly amend the per-board delays.
They are defined in [`icaruscode/Decode/fcl/CAEN_V1720_setup_icarus.fcl`](https://github.com/PetrilloAtWork/icaruscode/blob/2364c3bdd28314c17e7f36e0b69e5da5861a77b8/icaruscode/Decode/fcl/CAEN_V1720_setup_icarus.fcl) (at least until I fix the name that I just noticed to be typo'd).

ascarpel:
@PetrilloAtWork, the board delays are correct w/ the measure done with the oscilloscope. in the future we may want to move to a per-channel-correction, I believe. I am wondering if Animehs has informed you about the recent upgrade of the timestamps now and finally based on a TTT that resets with a known PPS output.

icaruscode/Decode/DecoderTools/PMTDecoder_tool.cc
```diff
@@ -161,67 +729,422 @@ void PMTDecoder::process_fragment(const artdaq::Fragment &artdaqFragment)
...
+            assert(boardInfo->fragmentID == fragment_id);
+            assert(boardInfo->setup);
+        }
+        
+        NeededBoardInfo_t const info = fetchNeededBoardInfo(boardInfo, fragment_id);
+        
+        nanoseconds const preTriggerTime = info.preTriggerTime;
+        nanoseconds const PMTtriggerDelay = info.PMTtriggerDelay;
+        auto const timeStamp
+          = fDetTimings.TriggerTime() - PMTtriggerDelay - preTriggerTime;
```
ascarpel, line 791: Just a comment here. In case you want to start preparing for the multi-acquisition readout you may start playing with the Withe Rabbit time from the Trigger Fragment class and the timeStampNSec from the Fragment overlay class. For legacy purposes it should be easy to apply a fix: the TTT doesn't mean anything if the reset is not sent with the PPS, but I think the timeStampSec method has the NPT server time stored regardless, so you'll lose nanosecond precision but still have something "meaningful" that appears and should be comparable with the Withe Rabbit time. Hope it makes sense.

ascarpel (approved):
I just ask @PetrilloAtWork if he wants to wait a little bit for this merge to set base for the decoder that uses the multi pmt readout window. otherwise is 👍 for me

icaruscode/Decode/DecoderTools/PMTDecoder_tool.cc
PetrilloAtWork: What do you mean with "legacy purposes" here?
For the rest, the only thing I need is a way to find out which waveform matches the global trigger.
I am not sure if the fragment timestamp (which I think is what you refer to with `timeStampSec`) helps there - I'll ask DAQ people.

PetrilloAtWork:
> @PetrilloAtWork, the board delays are correct w/ the measure done with the oscilloscope. in the future we may want to move to a per-channel-correction, I believe. I am wondering if Animehs has informed you about the recent upgrade of the timestamps now and finally based on a TTT that resets with a known PPS output.

Are they? I think I set them flat to 43 ns, which is intended to be the total delay. I would have thought that a more realistic setting would be something like:
```
  { Name: "icaruspmtwetop01"  TriggerDelay: "0 ns" },
  { Name: "icaruspmtwetop02"  TriggerDelay: "43 ns" },
  { Name: "icaruspmtwetop03"  TriggerDelay: "86 ns" },
```
(but I don't know the order and I don't know if that `0` is really 0 and if it was so linear and... in short, I don't know anything)

wesketchum:
Apologies for late response on this.

What you’ve done sounds fine, and I can update the appropriate config for the OM with the tools as you’ve done.

@SFBayLaser : is this going into the next icaruscode release (and, regardless, do you have an ETA for that?). I think it’s ok if so, but just need to know when/where.

I can do testing in the OM system on SBN-FD, so no issues. We can also make temporary local changes if we do find a major problem.

SFBayLaser (approved):
As usual, another tour de force in coding so can't complain about anything! 
I suppose one question, I've never understood the advantage to defining fhicl variables via "fhicl::atom" vs simply using the "get" method of the fhicl parameters object... I suppose there is a good reason but somehow it seems more complicated, both to write and to read.

SFBayLaser:
Ok, I will merge and include in the release to be built as soon as possible.

SFBayLaser:
Noting the breaking change above regarding processing... I'm unclear what this means for the current production running... are any changes necessary to the stage0 fhicl files?

PetrilloAtWork:
> Noting the breaking change above regarding processing... I'm unclear what this means for the current production running... are any changes necessary to the stage0 fhicl files?

It is worse than that: the changes are already in.

PetrilloAtWork:
> I suppose one question, I've never understood the advantage to defining fhicl variables via "fhicl::atom" vs simply using the "get" method of the fhicl parameters object... I suppose there is a good reason but somehow it seems more complicated, both to write and to read.

Two aspects (and a half):

1. `lar --print-description PMTDecoder` tells you the complete configuration, without you having to hunt it in potentially different parts of the code (constructor? `configure()`? `pset` saved as data member and used _anywhere_ in the code?). The half: it also offers you a place where you can describe what that parameter actually means (the `Comment` parameter).
2. if you type a parameter wrong, it tells you, as opposed as silently ignoring your "new" parameter and using the old one with the default value. It will tell for example that the configuration parameter `Delay` is unknown and, if it's mandatory, it will also tell you that the configuration parameter `delay` is missing.

The latter in particular is a very powerful feature.

It _is_ more complicate to write: firstly, it forces more planning than casually adding parameters anywhere; I consider this a feature. Secondly, it adds one additional step, that is the declaration of the parameter in a `struct`. This is not a big deal, but for some reason it does feel so (to me, as well). Also, there are cases where it is just not flexible enough (e.g. to match either a single atom or a sequence).

PR #165, opened 2021-05-19
Feature/jzettleNewFMObject

jzettle:
Pull request for the addition of a new sbn::SimpleFlashMatch data product replacing the anab::T0 object that was used previously. This is largely done to improve the amount of stored information from the simple flash match object in the CAF files. Tested on SBND and ICARUS simulations and shown to fill the correct object in the CAF files. The only changes to icaruscode are to .fcl files related to running the flash matching procedure.


ikatza:
This is all fine, but it depends on https://github.com/SBNSoftware/sbncode/pull/118

PR #178, opened 2021-06-10
Integration test fixes

PetrilloAtWork:
FHiCL configuration of the integration tests is updated to run and, in my hopes, to be correct.

It is not clear to me why the issues with reconstruction, which were present since many weeks, haven't been reported yet — possibly everybody is either using official simulated samples (the last production may predate the breaking changes) or running reconstruction on data (whose `Stage1` job configurations were updated).

I have also recalibrated the tests in the "sequence" suites: these tests run generation → detector simulation → digitisation → reconstruction jobs in chain, each step using as input the output from the previous one.
The "quick" tests, on the other end, use as input reference files. The most recent reference files are outdated and can't be used with the current `icaruscode`. The only exception is the generation stage, which can succeed and I have recalibrated as well.

These changes should allow for the regeneration of the reference files (which is in fact performed with the sequence suite), after which the quick test suite calibration may happen. Quick tests will indeed need their own specific recalibration, because they may process fewer events than the equivalent sequence test suite jobs.

## Known issues

Reconstruction stage seems to use a lot of memory (2.5 GB) for the _initialisation_ (of vertex reconstruction Boosted Decision Tree in Pandora): it appears the XML parser loads the whole content of the data file containing the 10,000 BDT parameters, and in a non-optimal way too. This affects mostly the single particle job reconstruction, where this initialisation sets the peak memory of the whole job about 2 GB higher than the rest of the job needs. Samples including cosmic rays will likely not be dominated by it, but neutrino-only samples still may.
I consider this an issue to be solved, and therefore I haven't increased the expected memory usage of the reconstruction job.
As a consequence, the integration test infrastructure will report a memory out of range warning for this test.
I suggest the same criterion to be used when calibrating the neutrino + cosmic rays reconstruction quick test.


PetrilloAtWork:
Tagging @SFBayLaser, @vitodb and Sergey Martynenko (as soon as I learn his user tag in GitHub).

PetrilloAtWork:
trigger build

PetrilloAtWork:
Ok, I was dreaming. @vitodb?

vitodb:
No you were not dreaming, that was right, just need to wait 10 minutes for the comment to get the CI built to start, it is in the queue.

PetrilloAtWork:
trigger build

PetrilloAtWork:
Yes, again: because I can.

PR #179, opened 2021-06-11
Removed unnecessary compiler warning setting

PetrilloAtWork:
GCC 8.3 (`e19`) wrongly issued a warning that therefore I disabled.
GCC 9.3 (`e20`) appears to have that issue solved.


PetrilloAtWork:
Giving up. There are four places where this option is used, at least in one of them it is still needed with `e20`.
It will need more care to solve this problem.

PR #181, opened 2021-06-12
PMT decoder support for multi-window PMT readout

PetrilloAtWork:
This requests contains changes to the trigger and PMT decoding softare.

## Trigger decoding

* bug fix: a default value was wrong (2 ns instead of 2 s)
* added some optional debugging information: dump the whole fragment content

## PMT decoding

The main feature being pursued here is the support for the "overlapping window" readout mode (the name is not official), which has been pioneered in run `6029`. In that mode, waveforms can come in packets shorter than the configured value (e.g. 30 µs) and they may come adjacent.

* major feature: support for waveforms of any size: timestamp is correctly computed including the actual waveform size
* adjacent waveforms (whose end and start match at less than 1 ns) are merged into a longer waveform
* more information in the optional debugging tree
* minor technical changes here and there

## Other features

* an analyzer module has been added to dump on screen the content of _artdaq_ data fragments

### DIFF
--- a/icaruscode/Decode/CMakeLists.txt
+++ b/icaruscode/Decode/CMakeLists.txt
@@ -4,6 +4,7 @@ art_make(
           EXCLUDE
                         PMTconfigurationExtraction_module.cc
                         DumpPMTconfiguration_module.cc
+                        DumpArtDAQfragments_module.cc
                         DaqDecoderICARUSPMT_module.cc
           MODULE_LIBRARIES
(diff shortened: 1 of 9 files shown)


PetrilloAtWork:
Does it compile? having made a partial merge (`git add --patch`) I haven't tested _exactly_ the version in the head ofthis published version. How do I find out?...

PetrilloAtWork:
trigger build

PR #195, opened 2021-07-06
Add bit to ignore the DuplicateEventTracker for the event display (th…

brucehoward-physics:
…anks Gianluca!). Note I am only adding these to the EVD fcls in the main directory and not the overburden directory.

NOTE: I only tested this out on a custom event display fcl using version v09_22_01 with data and not a checked out version of the default ones for simulation, but this seems like a rather minor change. Let me know if more formal testing/etc. is desired before merging.


PetrilloAtWork (approved):

SFBayLaser (approved):

SFBayLaser:
Will include this in this Friday's (July 9) release.

Issue #191, opened 2021-06-29
Space Charge Service Coordinate Restriction Fails in Edge Case 


gputnam:
At: https://github.com/SBNSoftware/icaruscode/blob/190b11d2965547f7218a4fd2b532e6cb0bc80b65/icaruscode/TPC/Simulation/SpaceCharge/SpaceChargeICARUS.cxx#L229
And: https://github.com/SBNSoftware/icaruscode/blob/190b11d2965547f7218a4fd2b532e6cb0bc80b65/icaruscode/TPC/Simulation/SpaceCharge/SpaceChargeICARUS.cxx#L243
The checking of the coordinate boundary is slightly different from the value it is set to. This means that for a point with (e.g.) x==368.4895, the Space Charge service throws an exception.

PR #199, opened 2021-07-09
Update Space Charge Service in ICARUS to split-wire geometry. Fix typos in coordinate fixing code.

gputnam:
Title. Also resolves https://github.com/SBNSoftware/icaruscode/issues/191.


gputnam:
The Space Charge service currently handles coordinate boundaries using hard-coded values taken from the Geometry.

The hard-coded TPC ID's are out-of-date as of the split wire Geometry. This PR resolves that issue. It does not try to fix the deeper problems with the use of hard coded values.

Issue #11, opened 2020-06-04
Make `icaruscode` depend on `sbncode`.


PetrilloAtWork:
`icaruscode` is going to depend on `sbncode` (and indirectly on `sbnobj`).

PetrilloAtWork:
Depends on SBNSoftware/sbncode#6.

PetrilloAtWork:
Branch SBNSoftware/icaruscode#feature/repositoryRestructuring now contains a version of `icaruscode` that is compatible with `sbncode` branch SBNSoftware/sbncode#feature/repositoryRestructuring. Similar branches are available for SBNSoftware/sbnobj and SBNSoftware/sbndcode.
How does GitHub references work here? Obviously I don't understand it yet.

Anyway: the adaptation work has been much less than trivial for the CRT data, which got merged, split and changed at the same time. I have started a document to be kept in `sbncode` repository to document all changes in one place.
When the actual merge ends, most of this work will need to be repeated if any change in the moved code has occurred in the meanwhile.

PR #203, opened 2021-07-15
c7:prof build cleanup

rennney:
1. Change in all `CMakeLists.txt` files is the same. Since `-Wno-maybe-uninitialized` is not supported by clang, it is made exclusive for GNU compiler
2. In `icaruscode/Analysis/overburden/OBAnaICARUS_module.cc` added override to funcion declaration
3. In `icaruscode/Decode/BeamBits.h` put names template in namespace
4. In `icaruscode/Decode/DecoderTools/PMTconfigurationExtractor.h` deleted one return statement since it is declared in cxx
file
5. In `icaruscode/Decode/DecoderTools/PMTconfigurationExtractor.cxx` and `icaruscode/PMT/Trigger/Algorithms/SlidingWindowCombinerAlg.cxx` removed std::move as redundand use
6. In `icaruscode/PMT/Algorithms/PMTsimulationAlg.cxx` added template statemens to std::function around lumbda functions to make it work (review?)
7. In `icaruscode/Decode/DaqDecoderICARUSPMT_module.cc` remove brackets around scalar initializer
8. In `icaruscode/PMT/Trigger/MajorityTriggerEfficiencyPlots_module.cc` commented out unused variable
9. In `icaruscode/Decode/DecoderTools/PMTDecoder_tool.cc` mark function as override and moved function setBitIndices defenition prior to its use
10. In `icaruscode/Decode/DecoderTools/TriggerDecoder_tool.cc` commented unused privet variable
11. In `icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc` since to add `int` and `" "` changed `" "` to `std::string(" ")` as a solution (please review?)
12. In `icaruscode/TPC/SignalProcessing/RecoWire/Decon1DROI_module.cc` removed `beginJob()` and `endJob()` public functions since they override one's from initial class. The private field `fEvent` declaration is moved from `beginJob()` to constructor (review?)
13. In `icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc` commented out unused variables

### DIFF
--- a/icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
+++ b/icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
@@ -199,7 +199,7 @@ private:
     unsigned int                                                fCoherentNoiseGrouping;      ///< Number of consecutive channels in coherent noise subtraction
     unsigned int                                                fCoherentNoiseOffset;        ///< Offset for the midplane...
     unsigned int                                                fMorphologicalWindow;        ///< Window size for filter
-    bool                                                        fOutputStats;                ///< Output of timiing statistics?
+//    bool                                                        fOutputStats;                ///< Output of timiing statistics?
     float                                                       fCoherentThresholdFactor;    ///< Threshold factor for coherent noise removal
 
     // Parameters for the ROI finding
@@ -593,7 +593,7 @@ void DaqDecoderICARUSTPCwROI::processSingleFragment(size_t                      
 
     std::string boardIDs = "";
 
-    for(const auto& id : boardIDVec) boardIDs += id + " ";
+    for(const auto& id : boardIDVec) boardIDs += id + std::string(" ");
 
     mf::LogDebug(fLogCategory) << "   - # boards: " << boardIDVec.size() << ", boards: " << boardIDs;
 
--- a/icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
+++ b/icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
@@ -593,7 +593,7 @@ void DaqDecoderICARUSTPCwROI::processSingleFragment(size_t                      
 
     std::string boardIDs = "";
 
-    for(const auto& id : boardIDVec) boardIDs += id + std::string(" ");
+    for(const auto& id : boardIDVec) boardIDs += std::to_string(id) + " ";
 
     mf::LogDebug(fLogCategory) << "   - # boards: " << boardIDVec.size() << ", boards: " << boardIDs;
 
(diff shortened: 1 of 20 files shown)


rennney:
This is the main part of some cleanup across icarusalg, icarus_signal_processing, and icaruscode to make it build with c7:prof  

Bullets 6), 11), and 12) are the most "questionable" for me

icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
```diff
@@ -593,7 +593,7 @@ void DaqDecoderICARUSTPCwROI::processSingleFragment(size_t
 
     std::string boardIDs = "";
 
-    for(const auto& id : boardIDVec) boardIDs += id + " ";
+    for(const auto& id : boardIDVec) boardIDs += id + std::string(" ");
```
PetrilloAtWork, line 596: I _request_ a different solution here, e.g.
```suggestion
    for(const auto& id : boardIDVec) boardIDs += std::to_string(id) + " ";
```
The accident why `(int + std::string)` works is that `std::string` is converted to ROOT `TString`.
It's better to avoid such accident.

icaruscode/PMT/Trigger/Algorithms/SlidingWindowCombinerAlg.cxx
```diff
@@ -55,7 +55,7 @@ auto icarus::trigger::SlidingWindowCombinerAlg::sortChannels
   (std::vector<raw::Channel_t> channels) -> std::vector<raw::Channel_t>
 {
   std::sort(channels.begin(), channels.end());
-  return std::move(channels); // is move() needed?
+  return channels; // is move() needed?
```
PetrilloAtWork, line 58:
```suggestion
  return channels;
```
The compiler clearly claims a move is _not_ needed.

PetrilloAtWork (changes requested):
I have left a couple of comments, and one of them is a request of change.
Some of Clang's complains are about code I wrote, and I would like to check it more extensively and give an alternative solution (I'll go with a branch for that).

PetrilloAtWork (commented):
@rennney: as anticipated, I have published a branch in my fork with the proposed changes covering some of the code I wrote, as pull request #204 (pull request into _your branch_). Let me know your comments.

rennney:
@PetrilloAtWork I looked through the changes on your branch and merged it. All looks good to me (sorry for removing the "prevent narrowing" part comlitely, should've thought this one through more carefully)

PetrilloAtWork (approved):
I have no pending comments left.

SFBayLaser (approved):
As I see the diffs here I don't have any issues with the two changes. In change #11 it is probably a leftover variable from when that module was cloned and then updated, in #12 the origins of that module date back to the early days of LArSoft so those two functions are remnant from before the days of recorded history.

SFBayLaser:
I will merge this for v09_26_01, probably this evening.

SFBayLaser (approved):
Ok, I think we are good to go here

PR #206, opened 2021-07-19
Small fix

rennney:
Error: suggest braces around initialization of subobject
Suggested fix: adding extra pair of {}


rennney:
solved in #207

PR #208, opened 2021-07-19
Ascarpel simple ophit flash study

ascarpel:
The following pull request proposes some NTuples for the analysis of the light, in particular of the OpFlash and OpHit objects. They are intended as a first step towards a more elaborate work for the PMT calibration. This PR: 
- Add a simple module inside icaruscode/PMT/OpReco/ to dump the association between OpFlashes and OpHits with some more handly quantities at the end of the light reconstruction. 
- updates the relevant configuration files inside icaruscode/PMT/OpReco/fcl and add an example to run the new analyzer as a stand alone called anaPMT_icarus.fcl 
- updates the relevant configuration files for the stage1 processing 

Updates tested locally with version v09_26_01 of icaruscode


icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -0,0 +1,457 @@
...
+
+    m_pmt_x.push_back(PMTxyz[0]);
+    m_pmt_y.push_back(PMTxyz[1]);
+    m_pmt_z.push_back(PMTxyz[2]);
+
+  }
+
+  fGeoTree->Fill();
+
+  fEventTree = tfs->make<TTree>("eventstree", "higher level information on the event" );
```
gputnam, line 192: Could this have a better name like "pmtTree"? Since it's just optical information.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -0,0 +1,457 @@
...
+  if ( !fFlashLabels.empty() ) {
+
+    for ( size_t i=0; i<fFlashLabels.size(); i++  ) {
+
+      auto const label = fFlashLabels[i];
+
+      art::Handle<std::vector<recob::OpFlash>> flash_handle;
+      e.getByLabel( label, flash_handle );
+
+      if( flash_handle.isValid() ) {
```
gputnam, line 391: I have been told in the past that these modules should not be flexible if input data is not present. So this should crash if the handle is not valid.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -0,0 +1,457 @@
...
+          auto const & flash = (*flash_handle)[idx];
+
+          m_flash_time = flash.Time();
+          m_sum_pe = flash.TotalPE();
+            
+          auto const & ophits = ophitsPtr.at(idx);
+
+          // Get the multiplicity, the position and the number of PE per Side
+          float xyz[3] = {0.0, 0.0, 0.0};
+          processOpHits( ophits,
```
gputnam, line 407: What about OpHits that aren't in a flash? Should those not be saved?

Also, it looks like the hit information does not include the data you would need to match hits to flashes in the TTree? I believe that would also be useful.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -0,0 +1,457 @@
...
+        mf::LogError("ICARUSFlashAssAna")
+           << "Invalid recob::OpFlash with label"+label.label()+"\n";
+      }
+
+    } // end for on flash input tags
+
+  }
+
+  else {
+    mf::LogError("ICARUSFlashAssAna")
```
gputnam, line 447: It's weird that a mis-configuration would log an error for every single event -- maybe move this to the constructor?

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -0,0 +1,457 @@
...
+
+  if ( !fFlashLabels.empty() ) {
+
+    for( auto const & label : fFlashLabels ) {
+
+        // TTree for the flash in a given cryostat
+        std::string name = label.label()+"_flashtree";
+        std::string info = "Three for the recob::Flashes with label "+label.label();
+
+        TTree* ttree = tfs->make<TTree>(name.c_str(), info.c_str() );
```
gputnam, line 210: So when I make a TTree like this I always make the branch definition as a C++ class with a ROOT dictionary. Is there a preference in style between doing that / making custom TBranch's as is done here?

gputnam (commented):
Made comments! Nothing that should stop this from being merged if it needs to be today, but maybe stuff to implement going forward.

Also, some if this stuff is _style_ -- and I don't know what the "preferred" choices are. It might be helpful to have an explicit discussion on these things, but that is probably longer term.

SFBayLaser (approved):

PR #214, opened 2021-07-28
Don't keep shower daughters

gputnam:
Change to Drop EM Shower daughters for the ICARUS G4 configuration.


gputnam:
As a note -- this doesn't actually change the G4 simulation. Just how the bookkeeping is done for truth information downstream of G4.

SFBayLaser:
Sorry, are we going ahead and merging this now or is there meant to be a discussion with SBN AI group first?

jzennamo:
This seems like it is a good idea and since it doesn't change the underlying physics and is consistent with what SBND does, I think it is good to go!

PR #216, opened 2021-07-29
Trigger simulation support using energy deposition summary

PetrilloAtWork:
The trigger simulation modules have been extended to support alternative ways to provide energy deposition information.
The motivation is that the source files might have dropped the information needed for the energy deposition because of data size; at the time of dropping, if a summary is created, the necessary information stays available for the next stages.
Two sources are now supported:
* `sim::SimEnergyDeposits` collections: all energy depositions, one by one;
* `sim::EnergyDepositionSummary`: summary of the information from above.

Summary objects may be produced with the provided module, `ExtractEnergyDepositionSummary`, which supports the concepts of beam gate, pre-spill window and active volume. While the intended source to extract the summary information from is energy depositions, an alternative, lesser way is also provided, from `sim::SimChannel`. The account of the total energy in the active volume may be slightly different here (for example, some charge — and energy — may get lost because of diffusion simulation). The time determination of the energy depositions is, on the other end, at the verge of the useless: the time recorded in `sim::SimChannel` is the time of arrival of the drifting charge, _quantised_ by the sampling period (400 ns), and subject to all physics effects from transportation (non-uniform electric field, diffusion, ...). Only the drift time is "reverted" in this module, and in the end the resolution is on the same order of magnitude as the BNB beam gate.
In short: do not use `sim::SimChannel` as source unless you don't have alternatives, and do not use it if the energy deposition falling into a narrow beam gate is important.

### DIFF
--- a/icaruscode/CMakeLists.txt
+++ b/icaruscode/CMakeLists.txt
@@ -1,6 +1,8 @@
 
 set( IFDH_ART_DIR $ENV{IFDH_ART_DIR} )
 
+add_subdirectory(IcarusObj) 
+
 add_subdirectory(Analysis)
 add_subdirectory(CRT)
 add_subdirectory(Decode)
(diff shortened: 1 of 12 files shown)


SFBayLaser:
trigger build

PR #224, opened 2021-08-03
extend decon2droi to full cryostats

wenqiang-gu:
I also added a configuration parameter `use_roi_refinement`. If it's set to `false`, it skips the ROI refinement in wirecell toolkit and reduces the memory usage. Note, our current plan is to use Tracy's and Dae's ROI, so it's fine for wirecell to skip its own ROI finding and refinement.


wenqiang-gu:
Although it doesn't screw up the configuration for including the new option `use_roi_refinement`, this option only works along with a new wirecell UPS. WireCell team is working on that.

PR #230, opened 2021-08-10
Updated ICARUS position based on SBN-doc-22998

aaduszki:
Updated ICARUS position in NuMI coordinates. @betan009 suggested to merge this soon so that the new values can be used in the new MC production.


betan009 (approved):
I reviewed the changes and everything looks good to me

PR #234, opened 2021-08-11
Fix of the PMT gain fluctuations.

PetrilloAtWork:
@fjnicolas has pointed out that the statistics used for PMT gain fluctuations in ICARUS is not correct.
Fluctuations are overestimated, effectively describing them as correlated between all photoelectrons at a given time.

This pull request aims to address the issue.

In short, instead of scaling the fluctuation from one Poisson-distributed photoelectron gain by a factor of _N_ photoelectrons, the fluctuation is extracted from a Poisson gain distribution for _N_ photoelectrons.

Review is appreciated.

@rennney , @SFBayLaser : this is a **breaking change** in the sense that it changes the result of the optical detector digitisation, and potentially of optical hit finding etc.
Reference files will need to be updated after this pull request is accepted and merged.

P.S. This pull request also contains a style update for a configuration file I used to test the change. That is not particularly relevant and reviewers can ignore it if they so desire.


fjnicolas (approved):
Looks fine!

PR #239, opened 2021-08-18
Split the reco test

rennney:
The reco test is split into two tests.
reco0 is run with detsim input via stage0_multiTPC_icarus_MC.fcl
reco1 is run with reco0 input via stage1_multiTPC_icarus_gauss_MC.fcl

preliminary reference files are generated but need to be updated through Jenkins after merging  

old reco test is preserved, but disabled

### DIFF
--- a/test/ci/ci_tests.cfg
+++ b/test/ci/ci_tests.cfg
@@ -352,6 +352,49 @@ parse_art_output=True
 output1=%(TFILENAME)s
 
 
+[test single_reco0_quick_test_icaruscode]
+STAGE_NAME=reco0
+INPUT_STAGE_NAME=detsim
+NEVENTS=2
+# calibrated with [e20] on icarusbuild02.fnal.gov
+cpu_usage_range=1650:2250
+mem_usage_range=4000000:5000000
+
+script=%(EXPSCRIPT_ICARUSCODE)s
+FHiCL_FILE=%(CI_FHICL_PREFIX_ICARUSCODE)ssingle_%(STAGE_NAME)s_quick_test_icaruscode.fcl
+INPUT_BASE_FILE_NAME=single_%(INPUT_STAGE_NAME)s_test_icaruscode
+OUTPUT_BASE_FILE_NAME=single_%(STAGE_NAME)s_test_icaruscode
+EXTRA_DIR=%(SINGLE_INPUT_DIRNAME_ICARUSCODE)s
+INPUT_STREAM=%(XROOTD_INPUTFILEDIR_ICARUSCODE)s/%(EXTRA_DIR)s/%(INPUT_STAGE_NAME)s/%(INPUT_BASE_FILE_NAME)s_%(REF_ICARUSCODE)s.root
+#outBNB is added to specify the output and might need to be removed later
+OUTPUT_STREAM=outBNB:%(OUTPUT_BASE_FILE_NAME)s_%(CUR_ICARUSCODE)s.root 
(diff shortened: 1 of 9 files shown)


rennney:
trigger build

test/ci/ci_tests.cfg
```diff
@@ -502,14 +545,61 @@ args=%(STDARGS_ICARUSCODE)s --input-file %(INPUT_STREAM)s --reference-files %(RE
...
+
+[test single_reco1_seq_test_icaruscode]
+STAGE_NAME=reco1
+INPUT_STAGE_NAME=reco0
+NEVENTS=%(NEVENTS_SEQ_SINGLE_ICARUSCODE)s
+# calibrated on icarusbuild01.fnal.gov on 20180325 (same as quick test because detsim has reduced number of events)
+# calibrated with [e19] on icarusbuild02.fnal.gov on 20200921
+# calibrated with [e20] on icarusbuild02.fnal.gov on 20210607: 800 MB -> 3.8 GB!!!
+cpu_usage_range=100:402
+mem_usage_range=700000:1200002
```
PetrilloAtWork, line 579: Consider using a marking different than `2`:
```suggestion
cpu_usage_range=100:406
mem_usage_range=700000:1200006
```
or whatever is available.
As a reminder, those weird numbers are there just to facilitate lookup in case of failure: the message tells you that resource usage was outside the limit `700000:1200006` and you can look up the configuration easily by searching for `0006` instead of the test name (which you may need to copy and paste).

test/ci/ci_tests.cfg
```diff
@@ -795,14 +928,63 @@ args=%(STDARGS_ICARUSCODE)s --input-file %(INPUT_STREAM)s --reference-files %(RE
 parse_art_output=True
 output1=%(TFILENAME)s
 
+[test nucosmics_reco0_seq_test_icaruscode]
+STAGE_NAME=reco0
+INPUT_STAGE_NAME=detsim
+NEVENTS=%(NEVENTS_SEQ_NUCOSMICS_ICARUSCODE)s
+
+# calibrated with [e19] on icarusbuild02.fnal.gov
```
PetrilloAtWork, line 936: This message needs to be updated... I suggest adding the date of calibration too (see some other messages).

test/ci/ci_tests.cfg
```diff
@@ -649,6 +739,49 @@ args=%(STDARGS_ICARUSCODE)s --input-file %(INPUT_STREAM)s --reference-files %(RE
...
+TFILENAME=hist-%(OUTPUT_BASE_FILE_NAME)s.root
+args=%(STDARGS_ICARUSCODE)s --input-file %(INPUT_STREAM)s --reference-files %(REFERENCE_FILES)s --extra-options --TFileName,%(TFILENAME)s
+parse_art_output=True
+output1=%(TFILENAME)s
+
+[test nucosmics_reco1_quick_test_icaruscode]
+STAGE_NAME=reco1
+INPUT_STAGE_NAME=reco0
+NEVENTS=2
+# calibrated with [e19] on icarusbuild02.fnal.gov
```
PetrilloAtWork, line 768: Message needs update.

PetrilloAtWork (approved):
Question: what is the parameter `output1` for? what does it add? would we gain anything by a `output2` which features the NuMI _art_/ROOT file?

The other comment is that there are messages about the calibration that need to be updated (they still report `e19` qualifiers, which is surely not what you used to calibrate the new jobs).

Anyway there are no show-stoppers to the inclusion as is anyway.
This also sets us one step closer to have tests for ICARUS data files too.

PR #244, opened 2021-08-22
change output tag for wirecell sim rawdigits

wenqiang-gu:
change the output tag of simulated RawDigits from "daq:daq[0,1,2,3]" to "daq:TPC[EE,EW,WE,WW]".


SFBayLaser:
We need this for testing so am including now

PR #242, opened 2021-08-19
G4 job configurations to follow a standard naming pattern

PetrilloAtWork:
To avoid the proliferation of naming schemes, I am proposing and applying a pattern for naming job configuration files.

This change is breaking: private configurations extending the previously hosted ones will find with a missing `#include`, and scripts and workflows relying on the old names will have to be updated as well.

At the same time, a few job configurations have been _added_, modifications of existing ones (esp. with space charge effects and overburden flipping).

Note that this pull request _includes_ #241 (which is aimed for production special branches).

## Rationale/ruleset

* base pattern: `<main flavour>_<stage>_<experiment>[_modifier[...]].fcl`
* main flavours:
    * `standard`
    * `cosmics` (somehow arbitrary, set this way by "tradition")
    * `intime` (implied `cosmics`)
* modifiers (in priority order: highest is earlier):
    * `sce`: adds `enable_spacecharge_icarus.fcl`
    * `volCryostat`/`volDetEnc`: single line changes
    * `overburden`/`nooverburden` (`use_overburden_geometry_icarus.fcl`/`use_nooverburden_geometry_icarus.fcl`)
* implementation:
    * `<main flavour>_<stage>_<experiment>.fcl` specialises
      `standard_<stage>_<experiment>.fcl` (or another `<main flavour>_<stage>_<experiment>.fcl`)
    * `<config>_modifier.fcl` specialises `<config>.fcl`
    * modifier drop-in configurations are kept in `modifiers` directory
        * exception: overburden-related drop-in _and_ configurations are kept in
          `overburden` directory

## Change map

| old                                                   | replacement                                       |
| ----------------------------------------------------- | ------------------------------------------------- |
| `enable_spacecharge.fcl`                              | `enable_spacecharge_icarus.fcl`                   |
| `cosmics_g4_enable_spacecharge.fcl`                   | `cosmics_g4_icarus_sce.fcl`                       |
| `cosmics_nooverburden_g4_icarus_volCryostat.fcl`      | `cosmics_g4_icarus_volCryostat_nooverburden.fcl`  |
| `cosmics_nooverburden_g4_icarus_volDetEnc.fcl`        | `cosmics_g4_icarus_volDetEnc_nooverburden.fcl`    |
| `cosmics_overburden_g4_icarus_volCryostat.fcl`        | `cosmics_g4_icarus_volCryostat_overburden.fcl`    |
[...]

PR #247, opened 2021-08-24
Fixes to FHiCL configurations

PetrilloAtWork:
Mending pull request #242: two configuration files were left behind, and another two were not updated to internally adopt the new names.
The fixing table on pull request #242 has been updated accordingly.


SFBayLaser:
trigger build

PR #252, opened 2021-08-31
New Noise Model Integration With WireCell

rennney:
Change in config/param files to incorporate a new noise model (per plane per TPC).
Add converter script to produce WireCell compatible .json.bz2 noise files from .root files

Simulation works with through : wirecell_multitpc_detsim_icarus.fcl (in icaruscode/fcl/detsim/)

PS. The extra commit is to restore the TPC volume names that occasionally got deleted in my branch


SFBayLaser:
I have uploaded the noise json files to icarus_data and pushed to develop there so will now merge this for testing locally. And launch a CI build too...

SFBayLaser:
trigger build

PR #254, opened 2021-09-13
pointed vertex MVA to new xml file

rhowell42:
needs to have feature/rh_BDTs from the icarus_data repository incorporated as well


brucehoward-physics (approved):
I can review from a code standpoint: it's a 1 line change here and then adding the XML to the appropriate area in icarus_data, so that looks okay.

Tracy and/or others should probably comment if there are needs for further tests of the performance etc. or if folks are happy with what Ryan showed recently.

PR #259, opened 2021-09-21
FluxFile name in NuMI MC

brucehoward-physics:
Change the flux file name in the NuMI fcl to remove the "dk2nu" bit.

Also I noticed that the NuEElastic fcl had this parameter too but was mostly a copy. I think you can edit the fcls like here to achieve the same effect, no?

However, I also see `interaction mode unknown!!` near the end of the job when trying to run this module. Not sure if this is known/anticipated or something to be aware of.

I'm including a change to these fcls in this pull request, but we can back out if needed. I've included @PetrilloAtWork on the PR review as he is the one I usually see condensing fcls. I would include Nicholas Suarez (who committed the NuEElastic fcls originally but I can't find his username to tag him. I'll send him a message.)


nbs25:
I'm fine with updating editing the fcls to remove the redundancy of the mostly-repeated file (I'll just have to let the production team know).

The`interaction mode unknown!!` code appeared when I changed the EventGeneratorList from default, but when I test the output to the G4 level and upwards, and when I plot the MC truth values, everything looks ok to me.

SFBayLaser:
I am assuming from a SLACK thread that Minerba is approving this so will go ahead and merge.

Issue #278, opened 2021-11-02
Use LArQL in ICARUS


wesketchum:
Per discussion in SBN AI meeting today (2 Nov), we want to update ICARUS to using LArQL for 2021C.

@marcodeltutto can help advise on the one line fcl change.

PR #281, opened 2021-11-02
Enabling LArQL for ICARUS 

jzennamo:
This enables the LArQL (addressing https://github.com/SBNSoftware/icaruscode/issues/278) and makes the ICARUS simulation consistent with SBND (see PR https://github.com/SBNSoftware/sbndcode/pull/194)

I'll tag @ggamezdiego to review since I don't see Alessandro M on github

### DIFF
--- a/fcl/services/simulationservices_icarus.fcl
+++ b/fcl/services/simulationservices_icarus.fcl
@@ -40,7 +40,7 @@ icarus_largeantparameters: {
     SkipWireSignalInTPCs:     []     # put here TPC id's which should not receive ionization electrons - used to simulate TPC geom volumes which are actually dead LAr volumes in protoDUNE
     FillSimEnergyDeposits:    true
     UseModBoxRecomb:          true   # use Modified Box recombination instead of Birks
-    UseModLarqlRecomb:        false   # use LArQL recombination corrections (dependence on EF)
+    UseModLarqlRecomb:        true   # use LArQL recombination corrections (dependence on EF)
    
     #* Recombination factor coefficients come from Nucl.Instrum.Meth.A523:275-286,2004
     #* * @f$ dE/dx @f$ is given by the voxel energy deposition, but have to convert it to MeV/cm from GeV/voxel width


jzennamo:
trigger build

jzennamo:
This is mostly important for light production outside of the TPC where the electric field strength is low, but it does have an impact on the high dE/dx region too. I have checked the dQ/dx vs. dE/dx for stopping protons with LArQL (red) and without LArQL (blue):

image

When taking the ratio there is a bit of statistical jitter since these aren't the same exact protons but you can see that it is a percent-level effect at very high dE/dx:

image

ggamezdiego:
For completeness, this is what we get for the case of the light in SBND with and without the LArQL model
Captura de pantalla 2021-11-03 a las 16 14 09

ggamezdiego:
BTW, I spoke today with Alessandro and told him about this PR

jzennamo:
@ggamezdiego excellent, thanks for sharing these! It is my understanding that until ICARUS moves to a model similar to SBND this effect won't be seen, is that true?

ggamezdiego:
@jzennamo, the function EFieldAtStep was modified in the ISCalcCorrelated class to be able to distinguish between IN/OUT active volume, by the ISTPC new class. My understanding is that everyone using this class should be sensitive to that

marcodeltutto (approved):

varanini:
I would approve. How do I do that?

varanini (approved):

ggamezdiego (approved):

mrmooney (approved):
This is fine.  However, in the future, please better communicate proposed changes that impact charge scale (even a small amount) to those leading relevant working groups.  Originally, we were informed that charge scale would not be impacted by this change at all, which was not true.

PetrilloAtWork:
Reference: [SBN DocDB 21270](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=21270).
Is there any comparison on the _light_ in ICARUS?

jzennamo:
I have not, but it modifies the recombination factor, I think in its current implementation this won't have a strong impact on ICARUS, how would you like me to explore this?

jzennamo:
My simulation is fairly low statistics but I'll plot the ratio of LArQL/no LArQL for abs(true X) and dE/dx... plots inbound...

jzennamo:
image

jzennamo:
Here is the light ratio vs dEdx:
image

PetrilloAtWork:
I also suppose there is not a large effect, but I think it's wise to verify.
Maybe my understanding is wrong... here it is.
The new algorithm modifies the recombination factor in regions with low electric field; it is asymptotically matching the old factor for high fields, and the plots in Laura Paulucci presentation show that above 300 V/cm the correction is indistinguishable by eye.
This saves us from issues in between the wire planes, where the field gets to ~900 V/cm.
On the other end, the difference should be large out of the field cage (vanishing electric field).
The recombination factor affects the light in that the recombined electrons produce scintillation light, which should then be quite large _outside_ of the field cage (which your _x_ plot does not cover). Possibly averaging inside and outside of the active volume may wash out the effect.

PetrilloAtWork:
trigger build

PetrilloAtWork:
I would try a C.I. test to get a feeling...

PR #282, opened 2021-11-03
Update calorimetry constants with comment

gputnam:
As discussed at the Tuesday Calibration meeting -- update the calorimetry constants in ICARUS MC.


SFBayLaser:
It seems that when people get tagged to be a reviewer they do get an email but it is non descriptive enough to get ignored... so I think one needs to be proactive to reach out to them and make sure they know to do the review.

mrmooney (approved):

PR #284, opened 2021-11-07
Updates for coherent noise simulation

SFBayLaser:
Changes enabling the TPC detector simulation to output "all" channels associated to "boards" to enable the coherent noise simulation across 64 channels.


SFBayLaser:
trigger build

PR #286, opened 2021-11-09
Encode PMT channel mapping DB key in the interface to DB

PetrilloAtWork:
This is an implementation of the proposal in icaruscode issue #151.

### DIFF
--- a/icaruscode/Decode/ChannelMapping/ICARUSChannelMapProvider.cxx
+++ b/icaruscode/Decode/ChannelMapping/ICARUSChannelMapProvider.cxx
@@ -19,6 +19,7 @@
 
 #include <string>
 #include <iostream>
+#include <cassert>
 
 namespace icarusDB
 {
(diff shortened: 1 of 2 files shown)


wesketchum (approved):
This looks fine to me. One thing is that I think asserts will be ignored in prof builds? If we want those in place, we may want to consider throwing exceptions instead? Could be wrong there though.

PetrilloAtWork:
You are correct that the `assert()` will be evaluated only in `debug` builds.
That is a situation that is never expected to happen by logic design, a prerequisite. In some sense the assert is just the documentation of the expectation. With the bonus that in some limited circumstances that expectation is actually tested.
So basically it boils down to the attitude "if this ID is not for PMT fragment, it's your fault and I don't even care what this code is going to do". I express this attitude in the documentation and also, as a courtesy, in the assertion.

SFBayLaser (commented):
This looks to be fine to me too.

SFBayLaser (approved):

SFBayLaser:
AAARGHH!! This went to the master branch.

PetrilloAtWork:
Ahahah! you fell for my trick!!

I.e.: sorry, I did not check and `hub` still does that sometimes/always/when it feels like.

PR #289, opened 2021-11-15
Offbeam triggers in the trigger decoding

ascarpel:
Pull request to add handling Offbeam triggers in the trigger decoding and further propagated to the downstream processing where relevant.


icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -199,14 +240,74 @@ void opana::ICARUSFlashAssAna::beginJob() {
...
+
+    }
+
+  }
+
+
+  // This ttree will hold the ophit information when a flash is not found in the event
+  // NB: information of the optical hits in events where flashes are present are lost
+  
+  if ( !fOpHitLabels.empty() ) {
```
PetrilloAtWork, line 277: It turns out this `if` branch is not necessary. Just saying.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -350,36 +537,97 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
...
-          m_beam_type = beamgate.BeamType() ;
+          m_beam_type = beamgate.BeamType() ; 
 
         }
 
       }
 
       else {
-        std::cout << "Invalid Trigger Data product " << fTriggerLabel.label() << "\n" ;
+        if( fDebug) { std::cout << "No sim::BeamGateInfo associated to label: " << fTriggerLabel.label() << "\n" ; }
```
PetrilloAtWork, line 565: I recommend either `MF_LOGTRACE()` or, if you really want to bypass message facility management of the debugging, `mf::LogVerbatim` (which outputs `INFO` verbosity level).

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -350,36 +537,97 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
...
+        m_gate_name = bitName(bit);
+        m_trigger_timestamp = trigger_handle->triggerTimestamp;
+        m_gate_start_timestamp =  trigger_handle->beamGateTimestamp;
+        m_trigger_gate_diff = m_trigger_timestamp-m_gate_start_timestamp;
+
+        std::cout << m_trigger_timestamp << "; " << m_gate_start_timestamp << " ;" << m_trigger_gate_diff << std::endl;
+
+      }
+      else{
+        if( fDebug ){ std::cout << "No raw::Trigger associated to label: " << fTriggerLabel.label() << "\n" ; }
```
PetrilloAtWork, line 586: Same comment as about re the output channel.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -350,36 +537,97 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
...
+
+        sbn::triggerSource bit = trigger_handle->sourceType;
+
+        m_gate_type = (unsigned int)bit; 
+        m_gate_name = bitName(bit);
+        m_trigger_timestamp = trigger_handle->triggerTimestamp;
+        m_gate_start_timestamp =  trigger_handle->beamGateTimestamp;
+        m_trigger_gate_diff = m_trigger_timestamp-m_gate_start_timestamp;
+
+        std::cout << m_trigger_timestamp << "; " << m_gate_start_timestamp << " ;" << m_trigger_gate_diff << std::endl;
```
PetrilloAtWork, line 582: This looks like a development message that needs to be removed or flagged.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -107,23 +131,37 @@ class opana::ICARUSFlashAssAna : public art::EDAnalyzer {
...
     //int m_nophit;
+    short m_baseline;
+    short m_chargesum;
+    int m_nticks;
     float m_beam_gate_start=-99999;
     float m_beam_gate_width=-99999;
     int m_beam_type=-1;
+    unsigned int m_gate_type;
+    std::string m_gate_name;
+    int m_trigger_timestamp;
```
PetrilloAtWork, line 161: The trigger timestamp from [`sbn::ExtraTriggerInfo::triggerTimestamp`](https://icarus-exp.fnal.gov/at_work/software/doc/icaruscode/latest/structsbn_1_1ExtraTriggerInfo.html#acbbec1e38c4afbcc587f2c2332668dd0) is absolute in nanoseconds from The Epoch. The only data type able to contain all that information is `std::uint64_t`.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -350,36 +537,97 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
...
+
+      if( trigger_handle.isValid() ) {
+
+        sbn::triggerSource bit = trigger_handle->sourceType;
+
+        m_gate_type = (unsigned int)bit; 
+        m_gate_name = bitName(bit);
+        m_trigger_timestamp = trigger_handle->triggerTimestamp;
+        m_gate_start_timestamp =  trigger_handle->beamGateTimestamp;
+        m_trigger_gate_diff = m_trigger_timestamp-m_gate_start_timestamp;
```
PetrilloAtWork, line 580: To make sure that you are not affected by any mistake in the conversion, I suggest you use the original timestamps in this formula:
```suggestion
        m_trigger_gate_diff = trigger_handle->triggerTimestamp - trigger_handle->beamGateTimestamp;
```

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -389,10 +637,13 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
       art::Handle<std::vector<recob::OpFlash>> flash_handle;
       e.getByLabel( label, flash_handle );
 
-      if( flash_handle.isValid() ) {
+
+      // We want our flashes to be valid and not empty
+      if( flash_handle.isValid() && !flash_handle->empty() && flash_handle->size() > 0 ) {
```
PetrilloAtWork, line 642: Double check...
```suggestion
      if( flash_handle.isValid() && !flash_handle->empty()) {
```

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -350,36 +537,97 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
...
+    }
+
+
+  /*
+    Now we take care of the flashes: we separate the case where we have a flash and the case where we have not a flash
+  */
+
   if ( !fFlashLabels.empty() ) {
 
     for ( size_t i=0; i<fFlashLabels.size(); i++  ) {
```
PetrilloAtWork, line 633: Same comment about the name of the index `i` as above (although you inherited this).

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -436,17 +687,27 @@ void opana::ICARUSFlashAssAna::analyze(art::Event const& e) {
 
       else {
 
-        mf::LogError("ICARUSFlashAssAna")
-           << "Invalid recob::OpFlash with label"+label.label()+"\n";
-      }
+        if( fDebug) { mf::LogError("ICARUSFlashAssAna")
+           << "Invalid recob::OpFlash with label"+label.label()+"\n"; }
```
PetrilloAtWork, line 691: Use `art::inputTag::encode()` to get the full label string, which might have critical hints.
```suggestion
           << "Invalid recob::OpFlash with label '" << label.encode() << "'"; }
```

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
PetrilloAtWork, line 633: Another comment... and another thing that you inherited: it appears that there is an implicit binding between the flash label and the cryostat it works on.
This should be checked at run time, since using the labels `[ 'E', 'W' ]` will produce different results than `[ 'W', 'E' ]` and than `[ 'E' ]` or `[ 'W' ]`.

PetrilloAtWork (changes requested):
There is a critical flaw in the data type of the trigger timestamp, that needs to be addressed somehow.
Another important aspect, that should be addressed but is inherited and not introduced with this pull request, is the undocumented and unchecked (unless I missed that!) relation between the flash data product tags in the configuration and the two cryostats. I won't hold the pull request because of this, since it is inherited, but this is a big liability to this module.
For the rest: it would be good to uniform the output to console (my recommendation: `mf::LogVerbatim`/`mf::LogTrace`), but again I will not stop the merge because of that if the author does not want to do that.

PetrilloAtWork:
Also, if in a rush, we can consider to merge the changes in the decoder first (they have no issue) and gain some more time to address the comments for the analyzer module.

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
ascarpel: The binding between label and cryostat is inherited from the reconstruction of Flashes.  I think checks of some sort should be made at that level. When hits are not assigned to flashes in events with flashes the information is lost ( this is a limitation ). When flashes are not found, hits are saved on both cryostats as the ophit reconstruction is agnostic to  the cryostat where it is applied.

ascarpel: I have however created a simple function to solve the only implicit binding native of this analyzer in line 713

PetrilloAtWork: Yes, that was the point in question. Although rather than relying on a naming convention I would ask the data.
The protocol prescribes that only one cryostat is considered per input data product, so we can reasonably assume that all the hits in the flash collection belong to that cryostat (interestingly enough, a `recob::OpFlash` is not directly aware of which cryostat it is in). So:
```cpp
assert(!ophits.empty());
geo::CryostatID const cid = fGeom.OpDetGeoFromOpChannel(ophits.front()->OpChannel());
```
gets the ID of the cryostat the data product covers (and `cid.Cryostat` is `0` or `1` in ICARUS) from the first optical detector hit in the flash (assuming, as documented in the assertion, that a flash is associated with at least one hit). In case of error (definitely unexpected), [`geo::GeometryCore::OpDetGeoFromOpChannel()`](https://nusoft.fnal.gov/larsoft/doxsvn/html/classgeo_1_1GeometryCore.html#a4ed940e72b58f91ab8c48aafc4a3a351) throws an exception.
Here `fGeom` is an hypothetical `geo::GeometryCore const& fGeom` data member initialised in the constructor with the traditional `fGeom{ lar::providerFrom<geo::Geometry>() }` (like `geop` in some other method of the module).

ascarpel: I have implemented this suggestion and removed the cryostat attribution via label

ascarpel:
Addressed also the second round of comments

icaruscode/PMT/OpReco/ICARUSFlashAssAna_module.cc
```diff
@@ -152,16 +191,23 @@ class opana::ICARUSFlashAssAna : public art::EDAnalyzer {
...
-  , fOpHitLabel( config().OpHitLabels() )
+  , fSaveWaveformInfo( config().DumpWaveformsInfo() )
+  , fOpDetWaveformLabels( config().OpDetWaveformLabels() )
+  , fOpHitLabels( config().OpHitLabels() )
   , fFlashLabels( config().FlashLabels() )
   , fPEOpHitThreshold( config().PEOpHitThreshold() )
-{ }
+  , fDebug( config().Debug() )
+{ 
+  fGeom=lar::providerFrom<geo::Geometry>();
```
PetrilloAtWork, line 209: For consistency:
```suggestion
  , fGeom( lar::providerFrom<geo::Geometry>() )
{ 
```

PetrilloAtWork (changes requested):
One issue: an object from geometry is copied instead of referenced.
Copies of geometry objects should have been forbidden in LArSoft; for some reason `geo::OpDetGeo` does not forbid it, but it's just by chance (and by bug).
There is also an optimisation suggested (avoiding a lookup of an information we already have).

PetrilloAtWork (approved):

SFBayLaser:
trigger build

SFBayLaser:
Ok, this looks like it is ready to go! Will start the merge process

SFBayLaser (approved):

PR #260, opened 2021-09-27
Feature/ac ldmgen

animeshgit20:
Please add the ldm generator in the icaruscode.

PR #295, opened 2021-11-19
new ldm branch for November production

animeshgit20:
The data files added to the icarus_data ups product area and linked with the module. ldm_g4 stage also added.


jzennamo:
@animeshgit20 how does this compare to https://github.com/SBNSoftware/icaruscode/pull/260, are they the same PR or should we disregard https://github.com/SBNSoftware/icaruscode/pull/260?

animeshgit20:
Please disregard #260.
________________________________

jzennamo:
One thing I would say before my review is I think a lot of this generator code would be better placed in SBNCode since it doesn't seem to be experiment specific, is there a reason these modules wouldn't be useful to SBND?

jzennamo:
also should `icaruscode/Generators/ldm-test.root_ldm.hepmc` have been included in this PR? It looks like a accidental commit to me

animeshgit20:
yes, this should not be in this PR, this should go to the icarus_data area, with a directory name Darkmatterfile. Just inform, I have kept this file inside icarus_data area with the folder name Darkmatterfile, the path of the file is included within the module.

/srcs/icarus_data/icarus_data
________________________________

SFBayLaser:
trigger build

betan009 (approved):

icaruscode/Generators/ldm_g4_icarus.fcl
```diff
@@ -0,0 +1,86 @@
+#include "services_icarus_simulation.fcl"
```
PetrilloAtWork, line 1: What is this configuration adding/changing to the functionality of the standard G4 (`standard_g4_icarus.fcl`)?
If it is only the `mcreco` module, and that one is intentionally desired, I _recommend_ that this configuration inherits the defaults from the experiment and just add it:
```
#include "mcreco.fcl"
#include "standard_g4_icarus.fcl"

physics.producers.mcreco: @local::standard_mcreco  # from mcreco.fcl
physics.producers.simulate: [ @local::physics.producers.simulate, mcreco ]
```

icaruscode/Generators/prodhepmc_icarus.fcl
```diff
@@ -0,0 +1,60 @@
+#include "simulationservices_icarus.fcl"
+#include "hepmcfilegen.fcl"
+#include "geometry_icarus.fcl"
+process_name: HepMCNuWro
+services:
+{
+  scheduler:    { defaultExceptions: false }    # Make all uncaught exceptions fatal.
+  # Load the service that manages root files for histograms.
+  TFileService:          {fileName: "MC_ldm_Aug27.root"}
```
PetrilloAtWork, line 9: Do we really want an output file with a date hard-coded in?
```suggestion
  TFileService:          {fileName: "MC_ldm_Aug27.root"}
```

icaruscode/Generators/prodhepmc_icarus.fcl
```diff
@@ -0,0 +1,60 @@
...
+ #filters reject all following items.  see lines starting physics.producers below
+ simulate: [ rns, generator ]
+ #define the output stream, there could be more than one if using filters 
+ stream1:  [ out1 ]
+ #trigger_paths is a keyword and contains the paths that modify the art::event, 
+ #ie filters and producers
+ trigger_paths: [simulate] 
+ #end_paths is a keyword and contains the paths that do not modify the art::Event, 
+ #ie analyzers and output streams.  these all run simultaneously
+ end_paths:     [stream1]
```
PetrilloAtWork, line 45: These are actually redundant, and you may choose to omit them (matter of style).
```suggestion
```

icaruscode/Generators/HepMCFileGen_module.cc
```diff
@@ -0,0 +1,457 @@
...
+//------------------------------------------------------------------------------
+
+//************** Animesh modified it to include file from ICARUS_data area ********************//
+
+void evgen::HepMCFileGen::open_file()
+{
+
+   std::string fullFileName;
+    cet::search_path searchPath("FW_SEARCH_PATH");
+    searchPath.find_file("Darkmatterfile/ldm-test.root_ldm.hepmc",fullFileName);
```
PetrilloAtWork, line 158: I have a problem with this approach: we have a fixed set of randomly generated events, stored in a standard repository (`icarus_data`) and we have a module to process those events.
I would *greatly* prefer the input file path to be passed as a parameter of the module.

Also I discourage the naming of a file as something containing `.root` as some code will find `.root` in the name and assume this is a ROOT file. I _recommend_ changing that `.` into something else (underscore maybe `_`?)

icaruscode/Generators/HepMCFileGen_module.cc
```diff
@@ -0,0 +1,457 @@
...
+//     #include "IFDH_service.h"
+//   #else
+//     // bare IFDHC
+//     #include "ifdh.h"
+//   #endif
+// #else
+//   #undef USE_IFDH_SERVICE
+//   // nothing doing ... use ifdef to hide any reference that might need header
+//   #include <cassert>
+// #endif
```
PetrilloAtWork, line 109: Not clear to me if this needs to stay. It smells outdated, although I am not 100% sure.
```suggestion
```
(right _now_ this module does not use IFDH at all, but that's a slightly different point)

PetrilloAtWork (commented):
As it is now, I fail to understand the point of adding the module `HepMCFileGen` into `icaruscode`: it converts a single, hard-coded generated file, into LArSoft format. Can you explain:
1. if this module would be compatible with any/many HepMC files (the module name suggests so); in that case, it should be changed to allow to receive a user-specified input file;
2. if there are reasons why the `TextGen` module already in LArSoft would not do the job (e.g. different input format?);
3. what is the expected future of this module: there is _a lot_ of code commented out, and unless there is the plan to add those are functionalities back, that code is better deleted.

As @jzennamo has noticed, `icaruscode/Generators/ldm-test.root_ldm.hepmc` needs to be removed from the pull request.

PetrilloAtWork (changes requested):
Sorry, I slipped on GitHub. My previous "comments" should have read as a request for changes.

icaruscode/Generators/CMakeLists.txt
```diff
@@ -1,2 +1,102 @@
+find_ups_product(ifdhc)
+#include_directories( $ENV{IFDHC_FQ_DIR}/inc )
+#set( IFDH_AS_SERVICE_LIB_LIST
+#     art_Framework_Services_Registry
+#     ${Boost_SYSTEM_LIBRARY}
+#     ${IFDH_SERVICE} )
+#include_directories ( . )
```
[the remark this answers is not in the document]
animeshgit20: To answer your question 
1. "if this module would be compatible with any/many HepMC files (the module name suggests so); in that case, it should be changed to allow to receive a user-specified input file;" That is correct it should be user specific. if there is already a module which can integrate hepmc file format, i think i will be happy to use that (but I am not sure about that). Only modification needed to include dark matter as the parent particle.
2. "if there are reasons why the TextGen module already in LArSoft would not do the job (e.g. different input format?);" Yes, any textGen module will do that that are already in LArsoft, but I am not aware if there is already one in the stable version, only changes require to add dark matter as the input to the file.
3." what is the expected future of this module: there is a lot of code commented out, and unless there is the plan to add those are functionalities back, that code is better deleted." I agree, previously this code taking input from local working area, so few functions are added, but after adding the file within icarus_data area, i have not removed those, i can remove those and update the branch if required.

icaruscode/Generators/HepMCFileGen_module.cc
animeshgit20: I can change the name of the file and remove .root extension. but i not sure I have understood the difference between calling the file from icarus_data area and passing the name as a parameter of the module, will that be much faster to access the file? Can you please explain little more?

mroda88 (approved):
So we had a chat with @animeshgit20  and @PetrilloAtWork . 
From the generator point of view there is not really much to say as the generation is done outside this code. 
With respect to the available TextFileGen module we have in larsoft a generation for BSM physics has the additional requirement that we should be able to import the parameters used for the generation inside LArSoft. This is the real issue, which is not available here and in the already available code. 

With @PetrilloAtWork we thought this to require additional data products and a dedicated way to take that information and add it into a CAF file metadata to be used for the analysis. 
At the time being we decided to postpone this activity as the production will consist only of a sample obtained with the same generator configuration.

Documentation on the expected structure of the file is still missing.

SFBayLaser:
So what is the status of this Pull Request? Are we postponing while issues above are worked out? Also I note one pending change request and one pending review... So I can't merge until those are dealt with...

animeshgit20:
I am going to work in next few hours on the updates that suggested by Marco and Gianluca and update the feature branch.
________________________________

icaruscode/Generators/HepMCFileGen_module.cc
```diff
@@ -0,0 +1,457 @@
...
+  explicit HepMCFileGen(fhicl::ParameterSet const & p);
+  void produce(art::Event & e)        override;
+  void beginJob()                                 override;
+  void beginRun(art::Run & run)                   override;
+  void endSubRun(art::SubRun& sr)     override;
+private:
+//  void ExpandInputFilePatternsDirect();
+//  void ExpandInputFilePatternsIFDH();
+  void open_file();
+  std::ifstream* fInputFile;
```
PetrilloAtWork, line 124: To have the file name as parameter (1):
```suggestion
  std::string fInputFilePath; ///< Path to the HEPMC input file, relative to `FW_SEARCH_PATH`.
  std::ifstream* fInputFile;
```
```suggestion
  std::ifstream* fInputFile;
```

icaruscode/Generators/HepMCFileGen_module.cc
```diff
@@ -0,0 +1,457 @@
...
+//  std::string              fFileCopyMethod;  ///< "DIRECT" = old direct access method, otherwise = ifdh approach schema ("" okay)
+  
+  double         fEventsPerPOT;     ///< Number of events per POT (to be set)
+  int            fEventsPerSubRun;  ///< Keeps track of the number of processed events per subrun
+  ifdh_ns::ifdh* fIFDH;             ///< (optional) flux file handling
+};
+//------------------------------------------------------------------------------
+evgen::HepMCFileGen::HepMCFileGen(fhicl::ParameterSet const & p)
+  : EDProducer{p}
+  , fInputFile(0)
```
PetrilloAtWork, line 137: To have the file name as parameter (2):
```suggestion
  , fInputFilePath(p.get<std::string>("InputFilePath"))
  , fInputFile(nullptr)
```
```suggestion
  , fInputFile(0)
```

icaruscode/Generators/HepMCFileGen_module.cc
PetrilloAtWork: To have the file name as parameter (3):
```suggestion
    cet::search_path searchPath("FW_SEARCH_PATH");
    searchPath.find_file(fInputFilePath,fullFileName);
```
```suggestion
    cet::search_path searchPath("FW_SEARCH_PATH");
    searchPath.find_file("Darkmatterfile/ldm-test.root_ldm.hepmc",fullFileName);
```

PetrilloAtWork: To have the file name as parameter (3):
```suggestion
    cet::search_path searchPath("FW_SEARCH_PATH");
    searchPath.find_file(fInputFilePath,fullFileName);
```

PetrilloAtWork:
To add to @mroda88 comment about the difference with `TextGen`: this module requires a specific non-standard format where some entries are reinterpreted (e.g. the first daughter of particle PDGID=`52` is actually whether this interaction is charged or neutral, etc.), and also the filling of `simb::MCNeutrino`, which as far as I remember `TextGen` does not attempt as it requires specific knowledge of the roles of the particles in the interaction.
While the addition of generic metadata is relatively easy, to abstract this last part is complicate and indeed it requires some dedicated effort (LArSoft should also be involved).
A third aspect is the population of the PDG database as known by ROOT (and eventually by GEANT?) to cover BSM particles specific to the model being generated.

animeshgit20:
All the changes suggested by Marco and Gianluca completed, added g4 fcl file for the ldm inside g4 configuration, many thanks to Gianluca and Marco,

PetrilloAtWork (approved):

jzennamo:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
Note the c7 build failure above (see https://dbweb8.fnal.gov:8443/LarCI/app/ns:ICARUS/build_detail/phase_details?build_id=icarus_ci/3306&platform=Linux%203.10.0-1160.45.1.el7.x86_64&phase=build&buildtype=slf7%20c7:prof). Please fix before I can merge in...

SFBayLaser:
trigger build

animeshgit20:
trigger build

animeshgit20:
trigger build

animeshgit20:
trigger build

animeshgit20:
trigger build

PR #296, opened 2021-11-19
Added fcl files for kdar higgs portal production

lcjrice:


jzennamo:
@mroda88 This got merged in by accident but I thought I'd bring it to your attention to review

mroda88:
@jzennamo I think that's ok. these are just technical files, there is no physics in it, apart from the selection of the Higgs mass. 

I don't think these values, simply because there are just 4 of them, can cover a phase space. So I would have not put it into the code itself. But I guess we need to keep track of them for production purposes. 
So for me they are ok. thanks for pointing it out to me.

PR #297, opened 2021-11-19
Feature/howard nov21 freeze

brucehoward-physics:
This has several changes, I'll give a short list and if I think there is something important to note about the status I'll put it here too:

* Added a fcl meant to be analogous to the BNB in-time sce on fcl but for NuMI. As it uses pre-existing fcls and just is a 2-line fhicl combining them via include (as in the BNB case) and the dump of fcl configuration seemed okay, I think this is _probably_ okay. HOWEVER, it would be good to check the output objects in some way. **Such a validation is still a work in progress!!**

* Pandora behavior changes: (A) I have switched Pandora to the "ActiveBoundingBox" mode (fhicl switch) and (B) removed a few algorithms.

A. In thinking further about some things, I wonder if the volume should be made to include the wires. I probably need to think more about this, and I also reached out to the Pandora folks for comment. In any case, even if this isn't the final configuration, I contend this configuration should be better around the cathode than the default.

B. I removed a few unnecessary algs (`LArTrackParticleBuilding`/`LArPcaShowerParticleBuilding`) from the "Master Settings" that seemed to be at play in one of the tracks-getting-called-showers pathologies found before.

* NuMI-specific configurations: Added NuMI-version of the "Master Settings" XML that sets the MaxX0 parameter to 2.0 (default 1.0 is not big enough for the NuMI beam window). Correspondingly, there is a set of 3 stage1 fcls that inherit the "non-NuMI" parent and then change the Pandora configuration to use this XML. The `nofilter` and `_MC` ones were tested on a simulated particle gun event to see that this had the desired effect; the third wasn't tested but is essentially the same. (If someone has an idea for an event to test let me know!) NOTE: I did _not_ make a "RawICARUS" version of this. If we want to make some files using the Raw Hits chain, one could be added.

I tag @jzennamo as a reviewer on this PR especially since the overall workflow/requests are involved in this part of the PR, though of course I welcome any comments from anyone on any point.

* Some changes to the stage0 fcls. The more recent changes were only present in the latest release I think (the `daqTPCROI`). I fixed up where I believe the fixes are necessary, but it's possible I missed something. I did run a few of them after the change (I think it was the `multitpc_icarus` and `splitstream_nofilter`) but didn't test every fcl.

Probably @SFBayLaser is best poised to know if I missed something here.

### DIFF
--- a/fcl/gen/corsika/prodcorsika_proton_intime_icarus_numi_sce_on.fcl
+++ b/fcl/gen/corsika/prodcorsika_proton_intime_icarus_numi_sce_on.fcl
@@ -1,0 +1,3 @@
+#include "prodcorsika_proton_intime_icarus_numi.fcl"
+
+#include "enable_spacecharge_icarus.fcl"
--- a/icaruscode/Decode/fcl/stage0_multiTPC_icarus.fcl
+++ b/icaruscode/Decode/fcl/stage0_multiTPC_icarus.fcl
@@ -16,11 +16,11 @@ physics.reco: [daqTPCROI, @sequence::icarus_stage0_multiTPC]
 
 ## boiler plate...
 physics.trigger_paths: [ reco ]
-outputs.out1.fileName: "%ifb_%tc-%p.root"
-outputs.out1.dataTier: "reconstructed"
+outputs.outBNB.fileName: "%ifb_%tc-%p.root"
+outputs.outBNB.dataTier: "reconstructed"
 
 # Drop the artdaq format files on output
-outputs.out1.outputCommands: ["keep *_*_*_*", "drop *_*_*_DAQ*", "drop *_decon1droi_*_*", "drop *_decon1DroiTPC*_*_*" ]
+outputs.outBNB.outputCommands: ["keep *_*_*_*", "drop *_*_*_DAQ*", "drop *_decon1droi_*_*", "drop *_decon1DroiTPC*_*_*" ]
 
 # Set up for the single module mutliple TPC mode...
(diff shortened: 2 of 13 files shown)


SFBayLaser:
trigger build

fcl/gen/corsika/prodcorsika_proton_intime_icarus_numi_sce_on.fcl
```diff
@@ -0,0 +1,3 @@
+#include "prodcorsika_proton_intime_icarus_numi.fcl"
+
+#include "enable_spacecharge_icarus.fcl"
```
PetrilloAtWork, line 3: I am trying to move away from `sce_on` in favour of just `sce`; but since we are mimicking `prodcorsika_proton_intime_icarus_bnb_sce_on.fcl` here the name you gave makes more sense.

PetrilloAtWork (approved):
Technically I do not see any issue (the module label names should be cross-checked by @SFBayLaser).
In terms of actual effects on reconstruction, I can't judge.

jzennamo (approved):
Looks good on my side, should we drop those two algorithms for the BNB stream too?

brucehoward-physics:
> Looks good on my side, should we drop those two algorithms for the BNB stream too?

These ones `I removed a few unnecessary algs (LArTrackParticleBuilding/LArPcaShowerParticleBuilding)`? The change was made in `PandoraSettings_Master_ICARUS.xml` which is used in the default BNB case!

jzennamo:
Ha, good eyes, I wonder if that calls into question the quality of my review ;)

SFBayLaser (approved):
I don't see any issues here, it passes the CI tests (though they may not directly test these changes) and it fixes some faux pas's I left behind.

PR #299, opened 2021-11-23
Update the Canny Edge Filter

SFBayLaser:
Currently the ROI finding is using the "old" 2D morphological filter. These updates allow use of the Canny Edge filter for ROI finding. The updates remove pedestal corrections and coherent noise removal and now expect this will have already been performed on the input waveforms. Some tuning of threshold parameters is still necessary but code is functional and being included to aid those optimization studies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 
 cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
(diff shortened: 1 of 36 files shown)


SFBayLaser:
trigger build

SFBayLaser:
And of course the CI tests will fail with these updates because I need to update the underlying dependent packages. Argh... Ok, it builds for me on SLF7 so proceeding...

Issue #293, opened 2021-11-17
Updated ICARUS Data Workflows For OffBeam streams (BNB and NuMI)


wesketchum:
There's been a bit of conversation ongoing on slacks and emails on getting workflows going for offbeam streams that are separated to be BNB-like and NuMI-like. To better track this leading into any ICARUS needed updates, I'm making an issue here and we should lead to appropriate PRs.

wesketchum:
From the online file management side:

* I added `offbeamnumi` and `offbeambnb` to `data_streams`
* I updated the assignment for new files to have beam type 'none'

Looks to me from FTS logs that SAM information is being properly declared, though still awaiting the first files to be transferred properly.

PR #300, opened 2021-11-23
Update trigger handling for off beam streams

wesketchum:
Added in OffbeamBNB and OffbeamNuMI filter streams/made the rest of the underlying fixes to get the filters to work.

(Also removed single quote deliminators in the trigger decoder tool because my emacs syntax highlighting couldn't handle it and it cost me 30 minutes figuring out that large chunks of code weren't actually commented out/inside quotes. Sorry/not sorry.)

### DIFF
--- a/icaruscode/Decode/DecoderTools/TriggerDecoder_tool.cc
+++ b/icaruscode/Decode/DecoderTools/TriggerDecoder_tool.cc
@@ -60,7 +60,7 @@ namespace daq
     if(info.wr_seconds >= 1483228800)
       correction = 37;
     uint64_t const corrected_ts
-      { (info.wr_seconds-correction)*1'000'000'000ULL + info.wr_nanoseconds };
+      { (info.wr_seconds-correction)*1000000000ULL + info.wr_nanoseconds };
     return corrected_ts;
 
   } // getNanoseconds_since_UTC_epoch_fix()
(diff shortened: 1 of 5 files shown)


wesketchum:
This should at least partially resolve #293.

SFBayLaser:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
trigger build

mayawos:
Should L.60-61 in icaruscode/Decode/fcl/stage0_multiTPC_splitstream_nofilter_icarus.fcl have both pathOffbeamBNB, pathOffbeamNUMI in the array?

wesketchum:
I was thinking to include it in a separate release? But I may have misunderstood our readiness for that from the discussion on Monday.

wesketchum:
> Should L.60-61 in icaruscode/Decode/fcl/stage0_multiTPC_splitstream_nofilter_icarus.fcl have both pathOffbeamBNB, pathOffbeamNUMI in the array?

Good catch! Just edited that ...

mayawos:
> Looking at the differences I see the addition of the offbeam BNB and NuMi triggers... but are we not also including min bias for both BNB and NuMI as well? Also, if I understand correctly, there is no "unknown" category anymore? Or do we want to deal with those in a later release?

I think we can deal with the unknown category in a later release. I don't think it's usable for production but it can be cleaned up after.
As for min bias, my understanding is that we use the same trigger bits for the NUMI and BNB portion of the min bias run, so currently there's no need to add the min bias BNB and NuMI yet.

mayawos (approved):

SFBayLaser (approved):
Ok, I will merge!

PR #306, opened 2021-12-01
Adding cafmaker for NuMI MC which only runs GENIE uncerts

jzennamo:
This is needed for NuMI caf making where we don't want to run the BNB flux uncertainties, go figure!


jacoblarkin:
So that is not actually going to get rid of the fluxweight. It looks like the definition of `caf_ana_sce_evtw_sequence` has changed to include the fluxweight already. There is actually a redundancy in the runprod for this file and `cafmakerjob_icarus_sce_genie_and_fluxwgt.fcl` since the event weight generators are included in the sequence of producers twice. You can run only genieweight by creating a new sequence or using `caf_ana_sce_sequence` and adding `rns`, `genieweight`, and `cafmaker` afterwards.

fcl/caf/cafmakerjob_icarus_sce_genie.fcl
```diff
@@ -0,0 +1,7 @@
+#include "cafmakerjob_icarus_sce.fcl"
+
+services.RandomNumberGenerator: {}
+
+physics.runprod: [ @sequence::caf_ana_sce_evtw_sequence, genieweight, cafmaker ]
```
jacoblarkin, line 5: This is the line I was talking about.

Here is a quick fix, but I don't know if this is best or if you want to define a new sequence in `cafmaker_defs.fcl`.
```suggestion
physics.runprod: [ @sequence::caf_ana_sce_sequence, rns, genieweight, cafmaker ]
```

jacoblarkin (approved):

jzennamo:
I assume a similar change should be made in `cafmakerjob_icarus_sce_genie_and_fluxwgt.fcl`?

jacoblarkin:
Yes. I'm not actually sure if the producers will run twice if they are included in the sequence twice, but if they are then it would be good to remove the redundant call.

I can make that change here or in a new pull request if you want me to.

jzennamo:
Feel free to push that change here, thanks so much!!!!!

cjbackhouse (commented):
~deleted~

cjbackhouse (approved):
Sad that we have to add yet more fcls. Probably a better method is to make the numi reweight module do nothing if the neutrinos aren't coming from the NuMI beam (I don't know how it would tell, but we should find or add a way) but this is OK for now.

PR #308, opened 2021-12-05
Update cal constants for 2021C data processing

gputnam:
title.


gputnam:
I'll be mentioning this in the ICARUS Software meeting tomorrow.

SFBayLaser (approved):
This looks good based on your presentations last week. We may need to update these one more time with changes coming with the patch release... while some bug fixes may affect this the main change is a change in the baseline of ROIs which probably impacts first and middle induction planes.

mrmooney (approved):
An improvement!  I echo Tracy here.  This will also need to be changed after channel-specific gains are tuned.

PR #317, opened 2022-01-06
Updating cathode location in SpaceCharge module

mrmooney:
This update should fix SCE simulation for ICARUS following a shift in the cathode location for both cryostats.  It should be verified with a MC sample of > 20k cosmic events, which I can carry out once the sample is available.


SFBayLaser (approved):
We need this fix straight away so am approving. We should put on the list fixing this properly using the geometry service so we don't encounter this problem again when the geometry invariably changes again.

SFBayLaser:
trigger build

PR #318, opened 2022-01-06
Updates to enable running of 2D deconvolution

SFBayLaser:
The ROI finder needed one more bug fix in order to work correctly, after that the scale factor needed an adjustment so that the "standard" parameters in the ROI finder and gauss hit finder would not result in too many hits. Finally adjust the definitions in the fcl files so allow switching to 2D deconvolution if desired.


rennney (approved):
I approve changes. 

Some comments (mostly for myself):
1)Fram_scale parameter might need a bit more work
2)Eventually we want a separate fcl file to run it with changed input for roifinder.

PR #325, opened 2022-01-14
Update WireCell Garfield simulation

rennney:
Update Garfield simulation in WireCell with new bias voltages

The change comes with updated file in icarus_data


rennney:
set it to develop branch, but needs to go to production release with the new patch if possible

rennney:
trigger build

PR #327, opened 2022-01-15
Refurbished PMT readout configuration discovery code.

PetrilloAtWork:
It should be more resilient to art changes, and might also work in the online monitor.

In particular, I have added a utility to extract the configuration from `art::Run` (or `art::Event`, or `canvas::Event`) instead of directly from the _art_/ROOT input file, and the module uses this new utility.

The issue with the online monitor was that it gets its data from a stream so it does not have a input file; this new approach _might_ work since the input stream does have _art_ runs (whether they hold configuration or not is a different question).


rennney:
trigger build

PR #331, opened 2022-02-04
Updates to decoder for 2d simulation and introduction of purity monitor module

SFBayLaser:
Sorry to combine two relatively independent things! 
1) The 1d drift simulation outputs all channels to its RawDigits, including those that are not physical. This matches the data and enables the coherent noise subtraction to work without having to otherwise know that there are only 32 channels that are real. Probably this is something that needs to be addressed... Meanwhile, the 2d simulation only outputs "live" channels with the result that the process of decoding and coherent noise subtraction was causing some channels to be "lost". This PR introduces a way to recognize when this is happening and to patch those channels back in. Note this required a bit of interface modifications in order to allow to change the number of channels to use in the coherent noise subtraction on the fly. 
2) This PR also introduces a new TPC purity monitoring module meant to enable switching the current 2D approach to one utilizing reconstructed objects starting with pandora tracks and using gauss hits.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 
 cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
-project(icaruscode VERSION 09.42.01 LANGUAGES CXX)
+project(icaruscode VERSION 09.42.02 LANGUAGES CXX)
 
 message(STATUS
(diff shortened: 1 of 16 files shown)


SFBayLaser:
trigger build

rennney:
trigger build

PR #333, opened 2022-02-07
Explicit setting for PMT gain

PetrilloAtWork:
The interface of the current single photon response (SPR) has been updated to receive the PMT gain as a parameter to determine the expected amplitude of the signal.
At the same time, the parameter is now also shared (via common configuration definition) with the gain fluctuation simulation.

Breaking change
================

There is no really breaking change in physics, but there will be changes.
So ok, yes, it is a breaking change.
The amplitude of the single photon response is slightly changed (O(0.1%)) because of rounding.
Also the gain used as a reference to compute the fluctuations is changed (sensibly: from 10^7^ to 7 10^6^), which scales down all the gains from all PMT stages down. As the gain of the first stage, leading the fluctuation, decreases, its _relative_ Poisson uncertainty increases. In other words, larger fluctuations will occur (note that the gain used in the creation of the waveforms has effectively _not_ changed, since it's taken from the SPR which already was, and still is, 7 10^6^, while the fluctuation is changed from 10^7^). Anyway, each stage gain is decreased by only 3.5%, so the increase in fluctuation won't likely be noticeable.

### DIFF
--- a/icaruscode/PMT/Algorithms/pmtsimulation_icarus.fcl
+++ b/icaruscode/PMT/Algorithms/pmtsimulation_icarus.fcl
@@ -68,18 +68,22 @@ icarus_photoelectronresponse_gaussian: {
 # of April 2020), which come from data analysis from the BNL group.
 # 
 # In particular, raise and fall times are reported to be 1.4 and 3.2 ns for
-# the PMT, and 1.8 and 4.2 ns including the 37 m cable distortion.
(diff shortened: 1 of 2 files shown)


ascarpel (approved):
I have verified the information written and the `peakCurrentFromGain` function giving results that match with the expected peak current at different choices of gain. All appears correct. 

Thanks for this edit, it is going to simplify the code and make it more transparent to not familiar users ( like me :P )

rennney:
trigger build

PR #335, opened 2022-02-09
Port decoder updates and purity monitor to production branch

SFBayLaser:
This PR should port the necessary updates to the MC decoder to handle the wirecell 2D simulation. Riding along for free is the purity monitor...

### DIFF
--- a/icaruscode/Decode/DecoderTools/CMakeLists.txt
+++ b/icaruscode/Decode/DecoderTools/CMakeLists.txt
@@ -8,6 +8,7 @@ art_make( SUBDIRS details
           LIB_LIBRARIES
                         icaruscode_Utilities
                         sbnobj_Common_PMT_Data
+                        canvas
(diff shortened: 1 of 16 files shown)


icaruscode/Decode/DecoderTools/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ art_make( SUBDIRS details
           LIB_LIBRARIES
                         icaruscode_Utilities
                         sbnobj_Common_PMT_Data
+                        canvas
```
rennney, line 11: This is from "Refurbished PMT readout configuration discovery code" 

would it work without actual file change?

SFBayLaser: Oh, I see. I don't think this is going to be an issue for anything.

PR #342, opened 2022-02-22
Trigger response simulation updates

PetrilloAtWork:
I am dumping my trigger working area into `develop`... the main update is the possibility of running most of the trigger emulation without the presence of the optical waveforms.
Trigger simulation is modularised; the first module discriminates the PMT waveforms against a threshold, and the following modules rely on that information.
These commits introduce a new small data product `sbn::OpDetWaveformMeta` that includes pretty much everything from `raw::OpDetWaveform` except for the waveform sample values. Such data products are produced either by a specific module or also automatically by the discrimination modules used in the first trigger simulation stage. The rest of the modules have been updated to rely on these small data object rather than to the waveform.
This should allow all trigger simulation which does not change the PMT discrimination threshold to be run on an input file without the original `raw::OpDetWaveform` data product.

The following list was created by `gh` command... not too bad.

- Added reduced optical detector data product without waveform.
- Removed analysis objects now promoted to reconstruction
- Renamed sbn::PMTcoverageInfo into sbn::OpDetWaveformMeta
- LVDSgates module now requires waveform metadata instead of waveforms.
- SlidingWindowTrigger module now requires waveform metadata instead of waveforms.
- Migrated all the main trigger modules to use tracking trigger gates
- fixup
- Removed deprecated data structures from trigger code.

**Depends on SBNSoftware/icarusalg#45.**


mmrosenberg:
trigger build

mmrosenberg:
Hi PetrilloAtWork - it looks like the CI build failed for this PR. When you get a chance, could you take a look at the logs (links above) for the slf7 e20:prof build? (The c7 build was expected to fail due to unrelated issues.)

PetrilloAtWork:
I have forgotten to mention it depends on a quite old `icarusalg` PR. Added to the description above.

PR #344, opened 2022-03-03
Fhicl files that work for the 330 and 300 MeV Portal mass points.

gputnam:
More fhicl files for the MeVPrtl Higgs Scalar generator -- more model points to generate.


rennney:
trigger build

PR #358, opened 2022-03-11
clearing std::cout prints from the crthit module and some correction …

biswaranjan91:
…to cable delay, noise monitoring module and data analysis module


rennney:
trigger build

PR #364, opened 2022-03-17
This is meant to reconcile TPC related changes on the production branch with the develop branch

SFBayLaser:
Christian Farnese has updated the purity monitor code to remove failures due to fitting problems, update the noise scaling factor for the detector simulation as per numbers determined by Justin Mueller.


SFBayLaser:
trigger build

SFBayLaser:
Oh, and this also adjusts the electronics gain used in the detector simulation as determined by Mike Mooney

PR #372, opened 2022-03-23
Changed misleading message in ICARUSFlashAssAna

PetrilloAtWork:
A message suggesting that the optical flash data product was not found was printed even when the data product was there but empty. Example:
```
%MSG-e ICARUSFlashAssAna:  ICARUSFlashAssAna:simpleLightAna@BeginModule  23-Mar-2022 13:14:22 CDT
 run: 7033 subRun: 1 event: 192799
Not found a recob::OpFlash with label 'opflashCryoE'
%MSG
%MSG-e ICARUSFlashAssAna:  ICARUSFlashAssAna:simpleLightAna@BeginModule  23-Mar-2022 13:14:22 CDT
 run: 7033 subRun: 1 event: 192799
Not found a recob::OpFlash with label 'opflashCryoW'
%MSG
```
This fix prints a different message in case the data product is not found and in case it is empty.

This is a GitHub change, I did not compile it and count on the integration test to do it for me.


rennney:
trigger build

PR #373, opened 2022-03-23
Fixed absolute trigger timestamp in trigger decoder [SBN2022A]

PetrilloAtWork:
The trigger timestamp ultimately comes from the White Rabbit system, which has its time in [International Atomic Time (TAI) standard](https://en.wikipedia.org/wiki/International_Atomic_Time). To properly correct it into the UTC standard we adopted for timestamps some second offsets are necessary.
These are applied the first time by the trigger "board reader" during event building, and the result is currently used to attribute the trigger fragment and the _art_ event an absolute time. The trigger decoder in `icaruscode` also performed (presumably) the same correction as a workaround for a past issue, but a few months ago it went off-sync with the board reader.

This change ultimately trusts the board reader time and rather than attempting to apply the same correction, it learns it from the  time already corrected by the board reader.
This correction offset is also added as a new data member of the `sbn::ExtraTriggerInfo` data product, so that the "raw" value can be determined off-line if needed.

Note that there is a offset parameter which is now unused. I have deliberately changed the name of the parameter since it does not do what the old name suggested (ok, it does nothing at all...).
The intention is to remove that parameter completely in the parallel pull request against `develop`.

### DIFF
--- a/icaruscode/Decode/DataProducts/ExtraTriggerInfo.cxx
+++ b/icaruscode/Decode/DataProducts/ExtraTriggerInfo.cxx
@@ -19,12 +19,15 @@
 namespace {
   
   // ---------------------------------------------------------------------------
-  struct TimestampDumper { std::uint64_t timestamp; };
+  template <typename T = std::uint64_t>
+  struct TimestampDumper { T timestamp; };
   
-  TimestampDumper dumpTimestamp(std::uint64_t timestamp)
+  template <typename T>
+  TimestampDumper<T> dumpTimestamp(T timestamp)
     { return { timestamp }; }
   
-  std::ostream& operator<< (std::ostream& out, TimestampDumper wrapper) {
+  template <typename T>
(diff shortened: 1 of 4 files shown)


jzettle:
I think this makes sense. Thanks Gianluca for finally correcting this. I must admit I hadn't followed this closely enough I thought we had resolved this difference already.

jzettle (approved):

wesketchum (approved):

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_37_01_02 LArSoft/webevd@LARSOFT_SUITE_v09_37_01_02 SBNSoftware/sbndaq-artdaq-core@v1_00_00of2 SBNSoftware/sbncode@v09_37_02_01 SBNSoftware/sbnobj@v09_12_12 SBNSoftware/sbnanaobj@v09_17_06_04 SBNSoftware/icarus_data@v09_44_00 SBNSoftware/icarusalg@release/SBN2022A SBNSoftware/icarus*@v09_37_01

PR #377, opened 2022-03-30
1sec shift is now removed between gate_start_timestamp and CRTHit T0

jedori0228:
There used to be a 1sec difference between the gate_start_timestamp and T0 of CRTHit, but this was updated in https://github.com/SBNSoftware/icaruscode/pull/373.
We now don't need manual shifts when applying the timing cuts on CRTData. This should be merged before the production starts.

### DIFF
--- a/icaruscode/CRT/CRTDataAnalysis_module.cc
+++ b/icaruscode/CRT/CRTDataAnalysis_module.cc
@@ -403,8 +403,8 @@ namespace crt {
       }      
       */
       /// Looking for data within +/- 3ms within trigger time stamp
-      /// Here t0 - trigger time -ve, only adding 1s makes the value +ve or -ve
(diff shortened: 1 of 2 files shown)


biswaranjan91 (approved):

PetrilloAtWork (approved):
Looks good.

A word of warning on the old code: adding `1e9` to an `int64_t` makes it become a `double` and lose precision (7-8 bits lost, becomes ~100 ns resolution). An integer like `1'000'000'000` should have been used instead (in this case the needed precision was microseconds, so it did not matter anyway).

jzettle (approved):

jedori0228:
This was accidentally merged to release/SBN2022A, but reverted back. A new PR for this is https://github.com/SBNSoftware/icaruscode/pull/384.

PR #382, opened 2022-04-01
Updates to the space point building

SFBayLaser:
These are primarily aimed at solving issues seen by the ML reconstruction effort:
1) Import the Cluster3D module to modify so it does not overwrite the 2 hit space point "chi2" with a set number
2) Remove the restriction that a hit cannot have been used to make a 3 hit space point. This was killing exactly the space points we were trying to recover! 
  The 2 hit space point volume is a bit concerning and still to be watched.


rennney:
trigger build

rennney:
trigger build

rennney:
trigger build

PR #383, opened 2022-04-02
Fixed absolute trigger timestamp in trigger decoder

PetrilloAtWork:
This is the twin pull request of #373, this one against `develop`.
This pull request includes those same changes (in fact, same commit), plus the removal of the vestigial `Offset` parameter, as announced in that very pull request.

Commit list, courtesy of `gh`:
- Fixed absolute trigger timestamp in trigger decoder
- Removed "Offset" configuration parameter from trigger decoder

### DIFF
--- a/icaruscode/Decode/DataProducts/ExtraTriggerInfo.cxx
+++ b/icaruscode/Decode/DataProducts/ExtraTriggerInfo.cxx
@@ -19,12 +19,15 @@
 namespace {
   
   // ---------------------------------------------------------------------------
-  struct TimestampDumper { std::uint64_t timestamp; };
+  template <typename T = std::uint64_t>
+  struct TimestampDumper { T timestamp; };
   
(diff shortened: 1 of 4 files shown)


PetrilloAtWork:
Reviewer guide: not much to say. The changes are the same as the ones already accepted on `SBN2022A` branch.
So beside possible technical error, the only question is whether that now vestigial `Offset` parameter should have been left there.
My own opinion is well represented by this PR...

jzettle (approved):
I agree with this again. I feel we should handle this in the online to avoid any issue where timestamps are displaced due to needing to make changes in multiple places, which is what happened here now that the precision has been corrected everywhere needed. With that philosophy the offset parameter is unnecessary. It looks correct to me.

rennney:
trigger build

SFBayLaser (approved):

PR #389, opened 2022-04-06
Fix on key-value CSV file parser

PetrilloAtWork:
The key-value parser can be used to parse a string in the type of format that the trigger hardware uses to communicate its data.
The format is a comma-separated list of words, some of them are keys, and then their values follow.

The parser is intended to be generic and as such it needs some heuristic to figure out whether a work is a key or not. The heuristic currently implemented fails miserably when the value is a hexadecimal number starting with a letter (e.g. `deadbeef`), and to address this the parser allows to specify custom patterns declared to be keys, and how many values such key have.
For example, a string `Hex1, deadbeef, Hex2, a10e, Value, 9` can be parsed specifying that both `Hex1` and `Hex2` (regex pattern `Hex[12]`) are key requiring one value. If this step is omitted, `Hex1`, `deadbeef`, `Hex2` and `cafe` would all be processed as keys (with no associated value).

Now the bug: after the end of a special key and values (in that example, `Hex1` and `deadbeed`), an immediately following special key would not have its number of values honoured (in the example, `Hex2` would not be interpreted as requiring a mandatory value, and `a10e` would be interpreted as another key).

This commit fixes that bug.

Review notes: I recommend skipping the review, since a real review of the change requires detailed understanding of the algorithm and the algorithm itself is not that interesting to learn.

### DIFF
--- a/icaruscode/Decode/DecoderTools/details/KeyedCSVparser.cxx
+++ b/icaruscode/Decode/DecoderTools/details/KeyedCSVparser.cxx
@@ -103,17 +103,21 @@ void icarus::details::KeyedCSVparser::parse
     
     std::string tokenStr { cbegin(token), cend(token) };
     
-    // if there are values pending, this is not a key, period.
     bool bKey = false;
     do {
       
+      // if there are values pending, this is not a key, period;
+      // if all required values have been assigned, the next token is a key.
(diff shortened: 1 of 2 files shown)


rennney:
trigger build

PR #390, opened 2022-04-06
runFilesFromSAM.sh can now translate only part of a SAM definition

PetrilloAtWork:
Feature added to `runFilesFromSAM.sh`.
This script queries SAM database for a run or a definition and fills a file list with paths to the files in that run or definition.
This change allows to limit the number of files put in the file list even when querying a SAM definition (before only the run queries supported this).

Review notes: I don't think this requires a review.


rennney:
trigger build

PR #391, opened 2022-04-06
Greenlee ph lite develop

hgreenlee:
Update modules SimPMTIcarus and ICARUSMCOpHit to be able to process either SimPhotons or SimPhotonsLites.


rennney:
trigger build

PR #393, opened 2022-04-12
Fixed faulty printout of sbn::ExtraTrigInfo

PetrilloAtWork:
Very minor fix (the time correction was 2-complemented when dumping the object to an output stream, instead of 36.0 it looked like 2⁶⁴/1e9 - 36).
Review not really needed.


mmrosenberg:
trigger build

PR #384, opened 2022-04-02
Feature/jskim crt hit reco timing shift

jedori0228:
This is a combined one of #377 and #378.
- There used to be a 1sec difference between the gate_start_timestamp and T0 of CRTHit, but this was updated in #373.
We now don't need manual shifts when applying the timing cuts on CRTData. This should be merged before the production starts.
- This is related to https://github.com/SBNSoftware/sbncode/pull/251. Using trigger timing information to convert CRT T0 timestamps to relative time. For this, trigger label is added, and also a MC default T0offset is set by a parameter.

PR #396, opened 2022-05-05
Update CRTHit timing structure

jedori0228:
This PR is related to https://github.com/SBNSoftware/sbnobj/pull/52.
- When looking at https://github.com/SBNSoftware/sbnobj/blob/v09_13_12/sbnobj/Common/CRT/CRTHit.hh#L49-L50, `sbn::crt::CRTHit::ts0_ns` and `sbn::crt::CRTHit::ts1_ns` should be the last 9 digits of the full timestamp in ns, but we've been saving the entire timestamp to them. This is fixed in this PR.
- Also, the 1 second shift in the trigger time is (or will be? @jzettle) fixed in the recent data, so the corresponding manual-shift is removed in the CRTHit reconstruction. This part has been already applied in the [release/SBN2022A](https://github.com/SBNSoftware/icaruscode/tree/release/SBN2022A) branch by https://github.com/SBNSoftware/icaruscode/pull/384.

### DIFF
--- a/icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
+++ b/icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
@@ -49,7 +49,7 @@ vector<art::Ptr<CRTData>> CRTHitRecoAlg::PreselectCRTData(vector<art::Ptr<CRTDat
     char type   = fCrtutils->GetAuxDetType(adid);
 
     /// Looking for data within +/- 3ms within trigger time stamp
-    /// Here t0 - trigger time -ve, only adding 1s makes the value +ve or -ve
+    /// Here t0 - trigger time -ve
     if (fData && (std::fabs(int64_t(crtList[febdat_i]->fTs0 - trigger_timestamp)) > fCrtWindow)) continue;
     
     if ( type == 'm'){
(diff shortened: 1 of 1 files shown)


biswaranjan91 (approved):

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
```diff
@@ -260,9 +260,9 @@ sbn::crt::CRTHit CRTHitRecoAlg::FillCRTHit(vector<uint8_t> tfeb_id, map<uint8_t,
     crtHit.pesmap      = tpesmap;
     crtHit.peshit      = peshit;
     crtHit.ts0_s_corr  = time0 / 1'000'000'000; 
-    crtHit.ts0_ns      = time0;
+    crtHit.ts0_ns      = time0 % 1'000'000'000;
     crtHit.ts0_ns_corr = time0; 
-    crtHit.ts1_ns      = time1;
+    crtHit.ts1_ns      = time1 % 1'000'000'000;
```
jzettle, line 265: I'm trying to understand this better and I'm guessing I'm missing background information about the construction of the timestamps. 

Are the ts0 and ts1 represented as second+nanosecond timestamps and you are trying to only extract the last nine digits from the nanosecond component in this effort? This what this change represents to me and I wanted to understand if I have the interpretation correct

jedori0228: That is correct. `ts0` is a 19-digit  integer which represent the time stamp in nanosecond, and we are trying to save the last nine digits-only to `ts1_ns`, and the second-only part (the rest 10 digits) to `ts0_s`. It looks to me that this was the original `intention` of introducing the lines https://github.com/SBNSoftware/sbnobj/blob/v09_13_12/sbnobj/Common/CRT/CRTHit.hh#L49-L50

jzettle: Thanks Jaesung, I thought that was the case. Thanks for confirming.

mmrosenberg:
trigger build

PR #394, opened 2022-04-29
Updated trigger decoding scheme

jzettle:
Updated trigger decoding scheme to work with runs taken with new trigger system. This includes additional information added to the online data fragment but this was tested at every step on data runs that do not have this additional fragment information. It has successfully decoded several runs with the new trigger system along with one run I took specifically (8103) in order to test the additional information. This is likely not 100% complete but it is operational and my understanding is this is an urgent need so this should remedy this need.

PR #398, opened 2022-05-13
Additional adjustments to trigger decoding

jzettle:
This is a partner to https://github.com/SBNSoftware/icaruscode/pull/394 as the trigger decoder work has greatly increased in scope. Now the idea is to make it backwards compatible with the old trigger data, which I did not envision as something happening immediately. In order to do that it requires a new trigger fragment type defined and the old fragment definition kept around. This allows to run over data collected with both the old and new trigger systems, which was not possible. That work is not complete, but this is something we can now use to run over any new data taken (as this is a new fragment it will not work with data taken before the fragment's existence) and then data with the old trigger system (before ~Mar. 2022) would be decoded with the old fragment implementation (a "stage0_defs_old.fcl" is included for this purpose).  

I have run this over files I took with the new "V2" fragment and it successfully decodes them. Run usable for this test is 8254 and this will also need to be included in the new DAQ release. Dependent on https://github.com/SBNSoftware/sbndaq-artdaq-core/pull/51 and the online version https://github.com/SBNSoftware/sbndaq-artdaq-core/pull/52 and https://github.com/SBNSoftware/sbndaq-artdaq/pull/65


PetrilloAtWork (commented):
I don't understand the need of a new _module_.
It appears to me that we can use the old module, and in the Stage0 configuration when we select the tool V2 or legacy we can set the fragment input tag accordingly. Am I missing anything?

jzettle:
That's largely the idea of this PR, only creating a new tool V2 (which is essentially a copy of the tool in the other PR on this) and then selecting based on the tool we want to use in the stage0 processing. There are other PRs supporting the creation and definition of the new fragment type. I think that is what I have done here. I might be missing something in your comment though. If I directly used the word module it was a mistake.

PetrilloAtWork:
In the pull request, there is a new _art_ module: `icaruscode/Decode/DaqDecoderICARUSTriggerV2_module.cc`. The configuration `icaruscode/Decode/fcl/decoderdefs_icarus.fcl` refers to it (let's see if I can still add a comment inline...).
There is also a new _art_ tool, and that's fine with me.

icaruscode/Decode/fcl/decoderdefs_icarus.fcl
```diff
@@ -55,4 +55,10 @@ decodeTrigger: {
                     DecoderTool:        @local::TriggerDecoderTool
 }
 
+decodeTriggerV2: {
+		   module_type:	        DaqDecoderICARUSTriggerV2
```
PetrilloAtWork, line 59: I would have thought that the old module `DaqDecoderICARUSTrigger` chould serve both the old "legacy" tool and the "V2" tool.

jzettle: Okay I see, this is now removed and there is one module to handle both cases

PetrilloAtWork (approved):
It looks ok to me (I did not test it, I trust Jacob's).

wesketchum (approved):

SFBayLaser (approved):
Approving

PR #404, opened 2022-05-27
Update detector geometry (icarus_v4) [2/2]

PetrilloAtWork:
This is the main part of a multi-repository pull request. See also: SBNSoftware/icarusalg#54.
Discussion should happen there.


SFBayLaser:
In looking at the CI test failures it seems to me that these are independent of the geometry change? So most are failing with memory issues, two because there are a different number of generated particles. So I am not sure that can be counted against this PR...

PetrilloAtWork:
Yes, C.I. tests are going to fail because of the "breaking" change.

PetrilloAtWork:
Let me clarify:
* this request updates geometry to a new formal version, from `icarus_v3` to `icarus_v4`
* every job is required a geometry that is consistent between the one used in the input file of the job and the one configured in the job itself
* C.I. uses as input files some reference files produced with an older version, which used `icarus_v3`
* the only exception is generation jobs, which are safe since their input includes no geometry at all

Given this, it is _expected_ that the standard C.I. test will fail on a formal geometry change.
I have run local C.I. jobs overriding the geometry check. The G4 stage still presents differences, which I attribute to some particles (e.g. neutrons) reaching the parts of CRT that are changed; at that moment, the random number sequence used by the old and the new job diverge, as do the job results. Statistically, though, the numbers appear consistent (as much as one can tell with a sample of just 2 events).

PR #407, opened 2022-06-06
Included new version of trigger decoder framework in development release

jzettle:
This PR adds the new trigger decoding framework to the develop branch. Same caveats as for the original PR to the production branch: https://github.com/SBNSoftware/icaruscode/pull/398 that it only works on data taken with the latest DAQ release or before the implementation of the new trigger system. In the near future I will create a separate branch to handle this data.

Tested with all 4 existing data streams with only the trigger decoder and also the full stage0 processing for run 8353

### DIFF
--- a/icaruscode/Decode/BeamBits.h
+++ b/icaruscode/Decode/BeamBits.h
@@ -47,8 +47,16 @@ namespace sbn {
     // --- BEGIN -- Generic bit functions --------------------------------------
     /// @name Generic bit functions
     /// @{
+    //template <typename EnumType>
+    //using mask_t = std::underlying_type_t<EnumType>;
+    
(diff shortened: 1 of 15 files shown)


rennney:
trigger build

rennney:
@jzettle Hi, it looks like CI unit test filed. Please check CI dashboard for the details. From the first look, it seems that file "stage0_icarus_defs_olddata.fcl" is missing

jzettle:
Thanks Sergey @rennney. You are right I thought I tested all pathways with this PR but clearly I missed the older data path. Added the needed fcl file and tested stage0 processing on an "old" data file before the new trigger system and it completed successfully.

rennney:
trigger build

PR #410, opened 2022-06-27
Adding additional tools for TPC-CRT matching

biswaranjan91:


tboone42:
This update has a number of changes to extant code for doing TPC-CRT matching. There is also new code that does other kinds of TPC-CRT matching, this code is still under construction but is usable in its current state. The edits to extant code all add functionality to access trigger timestamps as well as allow for separating data from MC analysis.  New additions include code to create "event display" TTrees for the TPC-CRT matching, as well as my first draft of an efficiency measurement tool.

PR #412, opened 2022-06-27
Creating "run1" fhicl files for stage0 processing of data

SFBayLaser:
The daq system is now sending up the data separated by stream type (e.g. "BNB") and it is not longer necessary, or even possible, to filter the data at stage0. These fhicl files are meant to be used for running stage0 on data without filtering to streams. I have also taken this opportunity to try to rationalize the fhicl naming scheme a bit, removing some old terminology (e.g. "multiTPC") that really no longer has context. Hopefully "run1" is descriptive to the use here! 

I have tested in limited capacity on a sample BNB file provided to me.

### DIFF
--- a/icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
+++ b/icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
@@ -17,8 +17,6 @@ physics.outana:        [ purityinfoana0, purityinfoana1 ]
 physics.trigger_paths: [ path ]
 physics.end_paths:     [ outana, streamCommon ]
 
-outputs.outCommon.fileName: "BNB_%ifb_%tc-%p.root"
-
 # Drop the artdaq format files on output
 outputs.outCommon.outputCommands:         ["keep *_*_*_*", "drop *_*_*_DAQ*", "drop *_daqTPCROI_*_*", "drop *_decon1droi_*_*", "drop *_decon1DroiTPC*_*_*" ]
 
--- a/icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
+++ b/icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
@@ -1,0 +1,27 @@
+###
+## This fhicl file is used to run "stage0" processing specifically for the case where all
+## TPC data is included in an artdaq data product from a single instance
+## 
+## It drops TPC raw digits but not the PMT ones.
+##
+#include "stage0_icarus_driver_common.fcl"
+
+process_name: stage0
+
+## Define the paths we'll execute depending on data
+#physics.pathOptical: [ @sequence::icarus_stage0_pmt ]
+physics.path:         [ @sequence::icarus_stage0_data ]
+
+## boiler plate...
+physics.outana:        [ purityinfoana0, purityinfoana1 ]
+physics.trigger_paths: [ path ]
+physics.end_paths:     [ outana, streamCommon ]
+
+outputs.outCommon.fileName: "BNB_%ifb_%tc-%p.root"
+
+# Drop the artdaq format files on output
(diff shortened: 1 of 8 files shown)


rennney:
trigger build

PetrilloAtWork (commented):
Thank you for the reorganisation (although I find ironic that the configuration with less light information is called "lite").
* it is not clear to me the difference between `stage0_run1_icarus.fcl` and `stage0_run1_rawpmt_icarus.fcl`
* I would consider to enforce that one output file be produced for each input file; that minimises the risk of mixing streams at Stage0 (which might be catastrophic for downstream processes). The configuration to do that looks more or less like this:
```
outCommon:
   {
     module_type: RootOutput
     dataTier: "reconstructed"
     compressionLevel: 1
     saveMemoryObjectThreshold: 0
     fileName: "%ifb_%tc-%p.root"
     fileProperties: { maxInputFiles: 1 }
     checkFileName: false
   }
```

SFBayLaser:
I have included Gianluca's suggestion and updated the files. As to "rawpmt"... I would agree that it does not seem there is a difference and I admit freely and without reservation that I simply copied what I saw in the repository for the old style... I would presume a pmt expert would update this as needed? Or we can delete going forward?

icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
```diff
@@ -0,0 +1,25 @@
+###
+## This fhicl file is used to run "stage0" processing specifically for the case where all
+## TPC data is included in an artdaq data product from a single instance
```
PetrilloAtWork, line 3: I don't understand this comment. Is this outdated by now?
```suggestion
##
## Standard Stage0 processing, but it explicitly preserves optical waveforms.
##
```

icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
```diff
@@ -0,0 +1,25 @@
...
+physics.outana:        [ purityinfoana0, purityinfoana1 ]
+physics.trigger_paths: [ path ]
+physics.end_paths:     [ outana, streamCommon ]
+
+# Drop the artdaq format files on output
+outputs.outCommon.outputCommands:         ["keep *_*_*_*", "drop *_*_*_DAQ*", "drop *_daqTPCROI_*_*", "drop *_decon1droi_*_*", "drop *_decon1DroiTPC*_*_*" ]
+
+## Modify the event selection for the purity analyzers
+physics.analyzers.purityinfoana0.SelectEvents:    [ path ]
+physics.analyzers.purityinfoana1.SelectEvents:    [ path ]
```
PetrilloAtWork, line 25: I would probably go a different approach, a bit more fragile but potentially requiring less maintenance:
```suggestion

#include "stage0_run1_icarus.fcl"

outputs.outCommon.outputCommands: [ @sequence::outputs.outCommon.outputCommands, "keep *_daqPMT_*_stage0" ]

```

PetrilloAtWork (commented):
I suggest a way to deal with the `rawpmt` configuration in the previous comments.
@SFBayLaser can decide whether to adopt it or not (note that I did GitHub-coding, i.e. no test), and whichever decision he states, this PR will be good to go for me.

PetrilloAtWork:
> a pmt expert would update this as needed? Or we can delete going forward?

That depends on what that configuration is meant to be.

PetrilloAtWork:
What the... I meanr "close the comment", not "close the PR with a comment"!!

icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
SFBayLaser: yes, I believe so. I do not know who the original author of this fhicl file is/was... I imagine they copied and modified. So, yes, should be modernized.

icaruscode/Decode/fcl/stage0_run1_rawpmt_icarus.fcl
SFBayLaser: I am ok if you are ok. I will try resolve this evening.

PR #413, opened 2022-06-27
Adding additional tools for TPC-CRT matching

tboone0042:
I've made edits to the TPC-CRT matching code which Bishu originally got from SBND code. The T0 matching code should now use the trigger timestamp correctly, and even has the ability to work on both data and MC depending on the fcl.

I've added some other analysis code to allow for making the "event display" TTrees which contain geometrical and timing data from attempts at matching CRT Hits with TPC tracks. This should work for both MC and data as well.

Finally, it's still under construction but I also have an efficiency measurement code suite, which I envision as the way to test TPC-CRT matching algorithms with truth data to evaluate how effective a given matching algorithm is.


rennney:
trigger build

PR #416, opened 2022-07-07
Fixes to scripts to get run file lists from SAM

PetrilloAtWork:
Minor changes to my scripts:
* `runFilesFromSAM.sh`:
    * fixed a bug when no limit on file number is set
    * added options to use SAMweb `locate-file` instead of `get-file-access-url`
    * added option to specify the SAM station name ("experiment")
* `sortDataLoggerFiles.py`: in the discovery of the first data logger, stop when a file is found which contains the first event of the run

### DIFF
--- a/scripts/runFilesFromSAM.sh
+++ b/scripts/runFilesFromSAM.sh
@@ -8,6 +8,8 @@
 # Date:   March 2021
 #
 # Changes:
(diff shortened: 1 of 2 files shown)


rennney:
trigger build

rennney:
trigger build

PR #418, opened 2022-07-13
Reimplemented channel mapping query for side CRT.

PetrilloAtWork:
I have tested the new implementation by an intermediate version running the same query in the two ways and throwing on mismatch, and successfully running on 2 events. More testing may be desired.

The pattern that I replaced is also present in other parts of the code. This commit also includes a proposal for one of them, but my test did not call that part of the code so the proposed reimplementation was not really tested, which is the reason it stays as a comment.

In case this new implementation turns out being still too slow, further gain may be possibly achieved turning the map into a hash table (`std::unordered_map`).


biswaranjan91 (approved):
Thank you. Looks good to me.

tboone0042 (approved):
Looks good to me!

gputnam (approved):

PR #421, opened 2022-07-28
Top CRT delays in crt_decoderdefs_icarus corrected; fatal error intro…

francescopoppi:
…duced in CRTHitRecoAlg.h corrected (CRT Timestamps were missing last significative bits due to a type conflict); CRTDataAnalysis module now included Flags in DAQTree; decodercrtjob now works with the new Trigger


rennney:
trigger build

PR #422, opened 2022-08-01
Updates to enable using 1D ROIs on 2D deconvolution, also noise filter updates

SFBayLaser:
The proposed plan for 2D deconvolution is to use the ROIs found in the 1D deconvolution and apply them to the 2D deconvolved waveforms. In principle this was already allowed but it seems was not totally functional so these updates make this chain work. As well there are modifications to the basic fhicl include file to allow this to work. 

In addition some small updates have been made to the noise filtering which are included in this pull request. These do not affect processing of waveforms, but should make event displays a bit cleaner (as the pedestal subtraction is slightly better).


SFBayLaser:
trigger build

SFBayLaser:
It would be ideal if this PR also updated icaruscode dependence on icarus_signal_processing to v09_56_00. I inadvertently pushed to develop for that package so went ahead and made a full tag (on the master branch) which includes some small udpates there too. This means updating the icaruscode ups product_deps file, I leave this for the release managers. Thanks.

SFBayLaser:
trigger build

rennney:
trigger build

PR #425, opened 2022-08-05
Wire Cell Field Response Adjustment

rennney:
Introduce a way to configure Field Response input for WireCell simulation through fcl files

PS. Needs to be merged ASAP + icarus_data needs an update with Filed Response files for tests


mmrosenberg:
trigger build

PR #427, opened 2022-08-09
Trigger decoder final updates

PetrilloAtWork:
This request adds a couple of missing features to the trigger decoder and polishes the code.
I consider this the final release of the decoder for the current version of the trigger, barred fixes.

The added features:
1. the beam gate time is corrected for the artificial veto (currently, 4 µs) used to effectively extend the PMT readout ahead of the beam
2. the width of the gate is correctly saved in the `sim::BeamGateInfo` data product
3. support is added for calibration type of gates

The first two features leverage on the new module by @ascarpel which saves the configuration of the trigger in a data product, where both veto and beam gate duration are stored. This comes with the addition in all the relevant workflows of that module.

In addition, the documentation of the trigger decoder has been completed describing in detail the content of the trigger data products.

Finally, the code has been polished.
The offset option has been deprecated and will trigger an exception if specified with a non-neutral value. This is a feature that I believe was already added, but must have been lost in the game of the merges.
Obsolete and temporary code has been removed, as some debugging messages.
In a couple of locations, the code has been restructured (e.g. the creation of `sim::BeamGateInfo` has been refactored).

The branch has been tested on top of `v09_56_00` running on a file of run 8650 (special minimum bias run) the configurations `decodeTrigger_icarus.fcl`, `decodePMT_icarus.fcl` and `stage0_run1_icarus.fcl`.

I would like @jzettle  to take a look at the code changes for approval, and @SFBayLaser  to check the changes in the Stage0 job configuration to check that the workflow is fine.

I consider this pull request a *high priority*, because it fixes the beam gate time which is otherwise broken.

List of commits:
- Indentation fixes
- Trigger decoder uses trigger configuration from input
- Added BeamGateInfo object in trigger output for calibration runs
- Trigger decoding tool code cleanup.
- Removed unsupported `TimeOffset` parameter from trigger decoding.

### DIFF
--- a/icaruscode/Decode/DaqDecoderICARUSTrigger_module.cc
+++ b/icaruscode/Decode/DaqDecoderICARUSTrigger_module.cc
@@ -34,6 +34,7 @@ namespace daq
     DaqDecoderICARUSTrigger & operator = (DaqDecoderICARUSTrigger const &) = delete;
     DaqDecoderICARUSTrigger & operator = (DaqDecoderICARUSTrigger &&) = delete;
 
+    void beginRun(art::Run& run) override;
     void produce(art::Event & e) override;
     
   private:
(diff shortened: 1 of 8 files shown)


rennney:
trigger build

jzettle (approved):

SFBayLaser (approved):
I don't see any obvious red flags in the stage0 configuration so seems fine to me.

PR #428, opened 2022-08-11
Integrate Energy-Scale Calibration Procedure with Data Processing

gputnam:
Depends on https://github.com/LArSoft/lardataalg/pull/35, https://github.com/LArSoft/larreco/pull/45, and https://github.com/SBNSoftware/sbncode/pull/279.


gputnam:
This update is intended for the fall production. It is not super urgent and also depends on a few upstream updates to larsoft. It deletes the local copies in icaruscode of the Calorimety modules since (I believe) those are no longer used. @SFBayLaser is that ok?

jzettle:
@gputnam @mrmooney What is the status of this and its dependents? We might be very interested in access to this for the trigger efficiency work. Thanks!

mmrosenberg:
trigger build

SFBayLaser (approved):
I don't see any issues in the updates

wesketchum:
@SFBayLaser so CI failures reported are ok?

gputnam:
@wesketchum @mmrosenberg I found the CI failure cause I think -- just pushed a commit to fix it.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #429, opened 2022-08-15
Add value of ion+excitation work function

ikatza:
Somehow not inherited from `simulationservices.fcl`

Needed for https://github.com/LArSoft/larsim/pull/97


mmrosenberg:
trigger build

PR #431, opened 2022-08-19
Trigger decoder prints input on parsing error

PetrilloAtWork:
I was not satisfied by the message that I was given on trigger parsing error.
This pull request makes the decoder print on console the whole string being parsed when the parser fails.

As a side, an additional identical parsing when diagnostics are required has been removed: diagnostics will show the same parsed data the trigger information is extracted from.


rennney:
trigger build

rennney:
trigger build

PR #433, opened 2022-08-20
Various small improvements to decoding implementation

PetrilloAtWork:
The most noticeable features of this pull request are:

1. Some helper for deleting from memory data products that _art_ had read from the input file (the ones produced in the current job don't abide); they are used in the TPC and PMT decoding to delete from memory the DAQ data fragments immediately after decoding; this should decrease by some hundred MB the amount of memory required by the job¹.
2. A `TriggerDumper` module (and a standard ICARUS job configuration) printing on screen the decoded information of the hardware trigger. This module requires as input an already decoded set of trigger data products.

This is part of a private branch with a lot of other commits. The branch itself was extensively used with data production-like jobs and I consider it well tested. The changed in this pull request are also involved in those jobs, but they have not been tested standalone.

--- 

¹ Unfortunately the CRT decoder uses a reading pattern that loads all data fragments anyway. I have discussed with @aaduszki and @wesketchum on how to proceed to avoid that, but I haven't implemented that change yet.

### DIFF
--- a/icaruscode/Decode/BeamBits.h
+++ b/icaruscode/Decode/BeamBits.h
@@ -47,9 +47,9 @@ namespace sbn {
     // --- BEGIN -- Generic bit functions --------------------------------------
     /// @name Generic bit functions
     /// @{
-    //template <typename EnumType>
-    //using mask_t = std::underlying_type_t<EnumType>;
     
+    /// Type for bit masks.
+    /// @note This is a glorified integral type.
     template <typename EnumType>
     struct mask_t {
       using bits_t = EnumType; ///< Enumeration type of the bits.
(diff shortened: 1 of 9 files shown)


rennney:
trigger build

rennney:
trigger build

PetrilloAtWork:
Fixed the Clang issue (it was just formal).

rennney:
trigger build

SFBayLaser (approved):
I think it is a good idea to drop the data as soon as it is no longer needed, so certainly I would approve of this change. My only beef is I wish people who modified code written by others would either respect their formatting conventions -or- assume all responsibility for future maintenance. Ok, rant off...

rennney:
trigger build

PR #435, opened 2022-08-24
Addition of trigger information in CAF files for time within beam gate information

jzettle:
Including trigger information in CAF files to compute the time within the beam gate. Requires changes in icaruscode to move the `sbn::ExtraTriggerInfo` object to sbncode to try to ensure continued running for both ICARUS and SBND. That is the bulk of the change here

Depends on equivalent PRs in sbncode, sbnanaobj, and sbnobj that accompany this. Links to follow in a comment

This is a **breaking change** in that existing code may not compile or link any more. Once building is fixed, no effect on the produced data is expected.

# Upgrade guide

Code using the `sbn::bits` namespace and objects (`BeamBits.h`) or the `sbn::ExtraTriggerInfo` object will need to point the path of the relative headers to the right direction:
```diff
-#include "icaruscode/Decode/DataProducts/ExtraTriggerInfo.h"
-#include "icaruscode/Decode/BeamBits.h"
+#include "sbnobj/Common/Trigger/ExtraTriggerInfo.h"
+#include "sbnobj/Common/Trigger/BeamBits.h"
```
and also likely add the `sbnobj_Common_Trigger` library to the link list of the proper `CMakeLists.txt`:
```diff
-  icaruscode_Decode_DataProducts
+  sbnobj_Common_Trigger
```
(`icaruscode_Decode_DataProducts` may still be needed if `icarus::TriggerConfiguration` class is used).


jzettle:
Depends on sbncode https://github.com/SBNSoftware/sbncode/pull/281, sbnanaobj https://github.com/SBNSoftware/sbnanaobj/pull/73, sbnobj https://github.com/SBNSoftware/sbnobj/pull/61

rennney:
trigger build

gputnam:
Hi Jacob! Thanks for putting this into the CAFs. Do we expect the SBND trigger information to be in this same format? And does this work for both NuMI and BNB?

PetrilloAtWork:
I am working on the resolution of the conflicts arising from competing merges.

jzettle:
Thanks @PetrilloAtWork, I did correctly remove those files from this PR as I had originally intended so it likely collided with the PR you mentioned that got merged already

PetrilloAtWork:
Conflicts are fixed. I have added an "upgrade guide" in the description for future reference.
Minor changes are also in the `sbnobj` pull request (SBNSoftware/sbnobj#61).

icaruscode/Decode/DecoderTools/TriggerDecoderV2_tool.cc
```diff
@@ -685,7 +683,7 @@ namespace daq
     // absolute time trigger (raw::ExternalTrigger)
     //
     fTrigger->emplace_back
-      (fTriggerExtra->triggerID, fTriggerExtra->triggerTimestamp);
+      (fTriggerExtra->triggerID, artdaq_ts);//fTriggerExtra->triggerTimestamp);
```
PetrilloAtWork, line 686: What is the intention in this change?

PetrilloAtWork (approved):
1. I suggest that the old paths be completely removed as opposed to commented out.
2. I have added also a secondary question about the trigger decoder change.
3. There are some libraries that are not required in the link list of the decoder directory... rather than guessing, I checked and pushed an amended list. That is the commit that is going to appear after this review.

The choice on point 1 is for @jzettle, the question 2 is mostly for my education, and point 3 is going to be there when I push it.
So overall as soon as @jzettle decides how to handle the suggestion at point 1, we are good to go one way or the other.

mmrosenberg:
trigger build

icaruscode/Decode/DecoderTools/TriggerDecoderV2_tool.cc
jzettle: I was having problems with this variable fTriggerExtra->triggerTimestamp producing an empty value here. Can revert this if I have to

jzettle:
Okay I think I got it to work. Perhaps it is safer to look at this again to ensure when I pulled my local branch with what is here I have the right setup after the changes

PetrilloAtWork:
I confirm that everything is good to go.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #436, opened 2022-08-25
Trigger emulation and efficiency analysis

PetrilloAtWork:
A monstrous commit (two, actually) collecting several months of work by @jzettle, @animeshgit20 and myself.
It includes trigger emulation updates (most noticeably the ones allowing emulation on an arbitrary number of time intervals within the event) and trigger efficiency code (analysis tree creation, extraction of track time for emulation, configurations to process data and simulation).
Pushing in `icaruscode` because it will be used (at least in part) also for CAF and production in general.

It should not change any test result.
Review: I have no idea how this code could be reviewed. @jzettle may try, but he's also author of a good chunk of it. Also, I don't think it's worth his time...
Incidentally, this code has been used for the most recent trigger efficiency analysis production (run `8650`).


rennney:
trigger build

PetrilloAtWork:
I have pushed two fixes (the same problem) for Clang, even if I, in my ancient wisdom, don't agree with it.
The other remaining issue is solved by SBNSoftware/icarusalg#60.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

jzettle (approved):
Took a look as best as I could. This code is a general commit of the trigger efficiency work. I have used this as a base for updating the trigger efficiency measurement and was able to process data from raw all the way to the tuples produced for the efficiency analysis mentioned as part of this pull request. It should be mergeable and we can use this to process the data with no issues

PR #434, opened 2022-08-24
Infrastructure for data-driven Single Photoelectron Response

PetrilloAtWork:
This pull request merges the infrastructure needed for using the "new" Single Photoelectron Response derived from data by @ascarpel in February 2022.
It does not change the configuration to _use_ it, but it makes it available for people willing to use it.

This pull request depends on #432 and includes it. The review will be easier if rebased on the branch in there (`PetrilloAtWork:feature/gp_CSVparser`).

Note to the release manager: please pay particular attention to the outcome of the `DetSim` and `Reco` CI test: the results are not expected to be changed, but I am not 100% sure that is achieved, given the dance of configurations I needed to do.

PR #437, opened 2022-08-26
Optical reconstruction algorithm and configuration updates

PetrilloAtWork:
This pull request introduces configurations for the optical reconstruction that were developed earlier this year. It also adds a new baseline algorithm estimating the pedestal of the optical detector waveforms from the beginning of all the waveforms recorded in the event.
The new configurations are not selected by default, but are available.
It is expected that the results of the reconstruction are the same as in the previous release.

The new baseline "algorithm" is technically quite complex, because I had to apply several workarounds to Kazu's optical reconstruction mini-framework, which has a specific protocol for baseline algorithms which assumes that the baseline can be extracted knowing only the values of the samples of the waveform at the current time.
Instead, the actual algorithm is a separate module (in the _art_ sense) that extracts the pedestal and RMS value from each channel and stores it into a data product, and what used to be the algorithm now is just a courier delivering to each waveform its pre-computed pedestal value. So it needs to know both which channel the waveform belongs to, and the pre-computed baseline value for all channels, both of which are normally not accessible at that point.

Note that this pull request accidentally depends on #434, since I made a mistake in the splitting of the different components and the algorithm in `icaruscode/PMT/Algorithms/SharedWaveformBaseline.h` which belongs to this pull request was actually pulled by that one. If this is a major problem, I can change that pull request and this one, but it's an effort that I would rather avoid. In effect, only the last commit (6160441108fe69f291ed11eafae1dee17e2e3fe9) is specific to this pull request.

Note for the release manager: please double-check that the integration tests give the expected results.


mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

ascarpel (approved):

PR #438, opened 2022-08-31
Memory footprint improvements related to CRT decoding

PetrilloAtWork:
These two commits are intended to reduce the memory footprint of Stage0 job.
1. The first commit uses a "selector" to decide which data products to look for CRT data, instead of trying all of them. The previous pattern loaded all the artDAQ fragment in memory instead. The net effect of this alone is a dramatic speed up in jobs decoding only CRT data, but little to no gain is expected by full Stage0 jobs, where most data will be loaded by the other decoders anyway.
2. The second commit enables dropping the data products (CRT data fragments) from the memory after use. This is a feature _art_ provides on demand (and currently only for data products from the _input file_), and since nothing else is ever expected to use this CRT data, no reload will follow. This is also not particularly relevant in terms of memory gain, since CRT data is not that large. But it's easy to add.

So, the real deal is actually somewhere else: with the first commit, the CRT decoder avoids reloading the TPC and PMT data fragments after their respective decoders have dropped them (in the same way as the second commit here, since pull request #433). So the interplay of this first commit with the rest of the Stage0 workflow is what is supposed to buy us some memory footprint reduction.
A test with `stage0_run1_icarus.fcl` on 5 events of run `8650` showed a decrease in memory of 1 GB compared to disabling all the data product dropping. No way I am going to trust this figure. But I am selling it anyway.

This commit is supposed not to change anything in the output.

Reviewers:
* @aaduszki and I discussed about the possible approaches to be taken here, and he is the maintainer of the decoder.
* @SFBayLaser as workflow manager, for backup.

### DIFF
--- a/icaruscode/CRT/CRTDecoder/CMakeLists.txt
+++ b/icaruscode/CRT/CRTDecoder/CMakeLists.txt
@@ -170,6 +170,7 @@ simple_plugin( CRTEventProducer "module"
 )
 
 simple_plugin( DecoderICARUSCRT "module"
+  icaruscode_Utilities
   art_Framework_Services_Registry
   art_root_io_tfile_support
   art_root_io_TFileService_service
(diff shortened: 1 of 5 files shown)


mmrosenberg:
trigger build

SFBayLaser (approved):
How can one criticize less memory? If I understand the changed files this happens entirely within the CRT decoding step and does not appear to modify the actual workflow as defined in fcl files... but I might be missing a point here. In any case this would be good to have running for full evaluation by the CRT folks.

aaduszki (approved):
I am not a maintainer of the decoder, I only helped with making sure the data from DAQ are being read correctly.
We discussed the changes with Gianluca, and his solution seems fine to me.

PR #440, opened 2022-09-07
Feature/ascarpel timing pmt crt

ascarpel:
This PR introduces several changes related to the PMT system and the way timing is corrected and calculated:

1. In Decode/ChannelMapping:  modified the  `PMTChannelMapping`  to include the laser optical channel in preparation of a future PR including the code for the laser calibration. It should be harmless
2. In `PMT/ICARUSFlashAssAna_module.cc`  new variable are added to the TTree holding the `OpFlash`  information. The new variables group together OpHit-level information related to the same PMT, in order to have a more intuitive overview of the components of each single OpFlash .  It should be harmless although it would be good to verify the final data size of the calibration ntuples after this modification and verify if compatible with the space allocated
3. Files `Timing/PMTTimingCorrectionService_service.cc` , `Timing/PMTTimingCorrectionsProvider.*`  are added to include a new LArSoft service that interface with the PostgreSQL calibration database and load the database entries given a run number. It should be verified the robustness of the code proposed and if there is room to improve its speed and performance. 
4. A new data product `Timing/DataProducts/PMTWaveformTimeCorrection.*` with the corresponding extractor `PMTWaveformTimeCorrectionExtractor.*` is designed to calculate and hold the time correction coming from the sampled reference signal on the spare channel of the CAEN1730V digitizers. This would allow dropping the corresponding waveform (the ones that now are saved with the instances “trigprm”, “BNB”, “NuMI”). The data product is saved following the same convention followed for the waveforms
5. A module called `Timing/OpHitTimingCorrection_module.cc`  which takes the OpHit data products vector and generates a new OpHit data product list with `PeakTimeAbs()`  and `PeakTime()`corrected for the delays calculated using the laser and muon calibration. The  choice to apply that correction to the OpHit was made to preserve the representation of the `OpWaveform::TimeStamp()`
6. Changes to `DaqDecoderICARUSPMT_module.cc`  to implement the correction of the first two bullet point above and produces the `PMTWaveformTimeCorrection`  data-products. It also recalculates the `OpWaveform::TimeStamp()` recalculating the offset starting from the sample of the "trgprim" labelled waveform. It should be verified if this approach is valid and consistent with the remaining part of the decoder logic. 

There should be backward compatibility with data collected previously, however, since most of the changes apply to the stage0 step of the production. Older samples must be reprocessed to include the changes here introduced


mmrosenberg:
trigger build

PetrilloAtWork:
I have updated to the current `develop`.

PetrilloAtWork:
It is now clear that I will have no time for further checks on this branch.

PetrilloAtWork:
The last commit is a fix triggered by a report by @francescopoppi and @aheggest who did not see the time correction on the hits in some output trees. The branch was creating the `ophitcorr` data product from `ophit`, and using the former for reconstructing the flashes, which is all well and good as long as you use the flashes. But some trees (`simpleLightAna` and probably any CAF tree) used to pick `ophit` which still did not have the correction.
Now `ophit` is corrected, and the uncorrected... `ophituncorrected` is just an accident of the workflow. We can even drop it if it bothers. The corrections are saved in their own data product, in case one wanted to attempt to change or undo them.

PetrilloAtWork (approved):
The main requests were addressed. I have become part of the solution, so my review is not fair any more.
Thus, **I hereby approve**.

SFBayLaser (approved):
I am not entirely sure why I was selected as a reviewer here since I am not familiar with the details of this code. I see that Gianluca has done a comprehensive review and given his long track record of careful work if he has approved then it is fine for me.

rennney:
trigger build

PetrilloAtWork:
> I am not entirely sure why I was selected as a reviewer here since I am not familiar with the details of this code. I see that Gianluca has done a comprehensive review and given his long track record of careful work if he has approved then it is fine for me.

AKA: "if the branch messes up anything, it's _his_ fault."

ascarpel:
After a quick investigation, the CI test failure is related to the change of the Optical Hit reconstruction process name.  the former `ophit` is now `ophituncorrected` while since this PR the `ophit` relates to the OpHit corrected including timing. The new names should be appropriately homogenized across the framework, in particular for Monte Carlo configuration files.

ascarpel:
commit [60121d3](https://github.com/SBNSoftware/icaruscode/pull/440/commits/60121d33d6e142f02d49c80d477ef10d507b72a3) should fix the above

rennney:
trigger build

mmrosenberg:
trigger build

rennney:
@PetrilloAtWork Hi, we got a merge conflict with recent restructuring of icarus config files. Please have a look

Issue #197, opened 2021-07-08
Trigger decoder update with sbndaq-artdaq-core


PetrilloAtWork:
When the release including SBNSoftware/sbndaq-artdaq-core#31 is adopted in `icaruscode`, the trigger decoder needs to be updated accordingly.

PetrilloAtWork:
Assigning to the interim release manager (@SFBayLaser) in order to ring the bell when the time is ripe.
Then it will be up to @PetrilloAtWork or @jzettle to act.

wesketchum:
I think this is now adapted? @SFBayLaser @PetrilloAtWork @jzettle , can I resurrect discussion so we can resolve?

PetrilloAtWork:
I rely on you (@wesketchum) to know if it has been adopted.
I can take care of removing the workaround code I added, although I don't see myself doing that this week.

SFBayLaser:
The link above indicates the PR in sbndaq_artdaq_core was merged successfully. I presume this means the workaround in @PetrilloAtWork 's code can now be removed? Is there already an icaruscode PR for this?

wesketchum:
@PetrilloAtWork has this been addressed already?

PetrilloAtWork:
I think it's [lines 340-345](https://github.com/SBNSoftware/icaruscode/blob/9677f3aedea2eb3ae29b9787f7dc21536e75c636/icaruscode/Decode/DecoderTools/TriggerDecoder_tool.cc#L340-L345), and this hasn't been fixed.
In fact, I hadn't even noticed your answer with the all clear.
Assigning this to myself, but there is a competing pull request that I don't want to mess up (besides not being sure of when I will have the time to study the code and remember where the workaround actually was).

PetrilloAtWork:
Ok, it was already assigned to me :-P

PR #441, opened 2022-09-10
Use corrected trigger time in trigger decoding

PetrilloAtWork:
This is an explicit way to address `icaruscode` issue #197.

The background as I understand it: the trigger decoder in `icaruscode` needs its timestamps in UTC; the hardware stamps in [TAI](https://en.wikipedia.org/wiki/International_Atomic_Time); their difference is some number of whole seconds.
The trigger decoder needs to correct all the timestamps from the hardware into UTC. Rather than doing the correction by itself, it acquires the corrected value from the online code (`icarus::ICARUSTriggerInfo::getNanoseconds_since_UTC_epoch()`) and it gets the correction offset by comparing it to the uncorrected timestamp; then it applies that correction offset to all timestamps it deals with.
Unfortunately the method `icarus::ICARUSTriggerInfo::getNanoseconds_since_UTC_epoch()` was affected by a bug (rounding error), so the trigger decoder needed to find a different way to do the correction. First, it did it locally, and at the same time issue #197 was opened.
Then, it moved to use the timestamp of the fragment, that the online code reset to match with the trigger time in UTC.

We may either choose to keep that way, relying on the convention by the online code, or to use now amended `icarus::ICARUSTriggerInfo::getNanoseconds_since_UTC_epoch()` as it originally was done.
Both are valid solutions, both do not charge the decoder with concocting the correction. 
This pull request implements the second option. The code currently in `develop` is an implementation of the first one.
The disadvantage of the fragment timestamp is that it relies on a convention that might in principle change. The disadvantage of using `icarus::ICARUSTriggerInfo::getNanoseconds_since_UTC_epoch()` is that the correction is computed a second time and it might differ from the one used in the online if different versions of the code are used; this may also be an advantage in case it is found that the online code applied the correction in a wrong way.

@jzettle is the author of both the board reader code and the trigger decoder; I am asking him to:
* review this change for correctness
* express his preference on which way to stick to

In case he chooses the way in this pull request, issue #197 should be closed as soon as this one is merged. If he prefers the way it is now, this pull request should be closed without merging, and issue #197 should be closed immediately.

### DIFF
--- a/icaruscode/Decode/DecoderTools/TriggerDecoderV2_tool.cc
+++ b/icaruscode/Decode/DecoderTools/TriggerDecoderV2_tool.cc
@@ -688,7 +688,7 @@ namespace daq
     // absolute time trigger (raw::ExternalTrigger)
     //
     fTrigger->emplace_back
-      (fTriggerExtra->triggerID, artdaq_ts);//fTriggerExtra->triggerTimestamp);
+      (fTriggerExtra->triggerID, fTriggerExtra->triggerTimestamp);
     
     //
     // previous absolute time trigger (raw::ExternalTrigger)
--- a/icaruscode/Decode/DecoderTools/TriggerDecoderV2_tool.cc
+++ b/icaruscode/Decode/DecoderTools/TriggerDecoderV2_tool.cc
@@ -438,25 +438,29 @@ namespace daq
   
   void TriggerDecoder::process_fragment(const artdaq::Fragment &fragment)
   {
-    // artdaq_ts is reworked by the trigger board reader to match the corrected
-    // trigger time; to avoid multiple (potentially inconsistent) corrections,
-    // the decoder trusts it and references all the times with respect to it.
-    uint64_t const artdaq_ts = fragment.timestamp();
     icarus::ICARUSTriggerV2Fragment frag { makeTriggerFragment(fragment) };
     std::string data = frag.GetDataString();
     char *buffer = const_cast<char*>(data.c_str());
-
+    
     icarus::ICARUSTriggerInfo datastream_info = parseTriggerString(buffer);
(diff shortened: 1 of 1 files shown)


jzettle:
I am not seeing the idea behind this change. Is there a reason not to just use the timestamp that comes with the data fragment which does this conversion already and instead do this conversion again within the decoder with the raw White Rabbit TAI timestamp? I am not against doing this if we want to there should not be any difference if that TAI->UTC conversion were to change.

jzettle (approved):

jzettle:
If we want to go this route in applying the timestamp, I do not have an issue with doing this.

rennney:
trigger build

PetrilloAtWork:
Fixed what was likely a merge issue. The humongous amount of commits is from merging `develop` in.

It should be ready to go now.

mmrosenberg:
trigger build

rennney:
@PetrilloAtWork Hi, we got a merge conflict with recent restructuring of icarus config files. Please have a look

wesketchum:
@PetrilloAtWork it looks like there are merge conflicts on the triggerdecoderv2 tool?

jzettle:
Do I need to do anything else here? I already approved these changes via review because it sounds like this is the way we want to handle this going forward.

PetrilloAtWork:
It seems clear now that the other approach, of relying on the board reader to set the proper trigger time as fragment time, is the safer one.
It is less robust in case mistakes are discovered, but it works better when processing old data, whether one does not need to make sure the correct offset is configured. Also, offset of 1 second may manifest in DAQ errors, meaning that if the correction to the time is wrong DAQ may provide a hint of that.

Closing this request.

Issue #243, opened 2021-08-21
Restore (modified) per-event random policy as default


PetrilloAtWork:
LArSoft `v09_28_02` adopts a newer version of `nugen` which introduces some level of control for seeds outside `art::Event` in the `perEvent` random seed policy.

As a reminder, the original issue, reported as [LArSoft issue #26116](https://cdcvs.fnal.gov/redmine/issues/26116), was that `perEvent` policy would set the random seeds only just before processing each event, so that the engine seeds remain uninitialised (or better, initialised to `0`) from the beginning of the job until the first event.
This caused major issues in `CORSIKAGen`, whose `beginJob()` would use a random generator to pick one of the available database files, and possibly to `GENIEGen` and others.

The new `perEvent` policy allows to specify a fallback policy (`initSeedPolicy`) being used when not in processing an event. This is not a perfect solution, but provides some control before the first event.
Given the known behaviour of `CORSIKAGen`, I advocate the restoration of `perEvent` as main policy, using `random` as fallback. This makes it much harder to exactly reproduce _some_ jobs (still possible via a master seed though), but it "guarantees" no fixed seed.

This issue should be addressed when LArSoft `v09_28_02` or newer is adopted in `icaruscode`.

wesketchum:
@PetrilloAtWork has this been done, now that we're beyond this LArSoft version?

PR #442, opened 2022-09-10
Random seed policy moved back to per-event

PetrilloAtWork:
The policy now uses `random` as fallback policy for per-event seeds.

This addresses issue #243.

The background: LArSoft provides a policy for providing seeds to the random number generators which ensures that the same event will always receive the same seeds, guaranteeing job reproducibility. Unfortunately that policy does not define what happens when seeds are required outside the event boundaries; a typical case is the LArSoft implementation of CORSIKA interface, `CORSIKAGen`, which _at beginning of the job_ decides randomly which pool of pre-generated events to use in the job. This choice happens at a time when no event is being processed (no input file, in fact, has been opened yet). The undefined behaviour of the per-event policy was in fact to always use `0` as a seed, which would cause the same pool of cosmic ray events to be used for every job.

While LArSoft offers no real solution to this issue, it now provide a workaround in allowing to control which seeds are provided at that time. My choice here is to use a `random` policy, which does not guarantee reproducibility (e.g. rerunning a job on a certain input will pick a different CORSIKA event pool) but at least behaves as a random generator.

This pull request defines this new configuration, defines also an alias `icarus_default_NuRandomService` supposed to define the "default" random seed service configuration for ICARUS, and then it uses the latter wherever there is the need to configure that service.

This may be a breaking change for the simulation, since different random streams may be generated for some of the modules, but it will likely not be noticed by the integration tests, which fix the random seeds overriding the standard configuration settings.

This pull request has been tested running the unit tests, a `prodcorsika_genie_standard_icarus.fcl` job, a `standard_g4_icarus.fcl` job and a `standard_detsim_icarus.fcl` job.

### DIFF
--- a/fcl/caf/cafmakerjob_icarus.fcl
+++ b/fcl/caf/cafmakerjob_icarus.fcl
@@ -19,7 +19,6 @@ services:
   @table::icarus_wirecalibration_minimum_services
 
   SpaceChargeService: @local::icarus_spacecharge
-  NuRandomService: @local::per_event_NuRandomService
 }
 
 
--- a/fcl/caf/cafmakerjob_icarus_data.fcl
+++ b/fcl/caf/cafmakerjob_icarus_data.fcl
@@ -2,7 +2,6 @@
 #include "channelmapping_icarus.fcl"
 
 #include "correctionservices_icarus.fcl"
-#include "seedservice.fcl"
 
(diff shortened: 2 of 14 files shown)


SFBayLaser (approved):
I believe this is a good change to make. Thank you!

rennney:
trigger build

PR #444, opened 2022-09-13
Add in SQLite-based tools for normalization procedure

gputnam:

PR #448, opened 2022-09-15
Fix Value of Mod-Box Beta in Calorimetry

gputnam:
Fixes the value, which was mistakenly put to 0.215 in https://github.com/SBNSoftware/icaruscode/pull/428.

### DIFF
--- a/icaruscode/TPC/Calorimetry/CMakeLists.txt
+++ b/icaruscode/TPC/Calorimetry/CMakeLists.txt
@@ -22,6 +22,7 @@ art_make( MODULE_LIBRARIES
            ROOT::Physics
            ${MF_MESSAGELOGGER}
 	TOOL_LIBRARIES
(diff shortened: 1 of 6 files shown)


gputnam:
Note -- this has a few extra commits in it because it also (mistakenly) includes https://github.com/SBNSoftware/icaruscode/pull/444. Once that is merged in this will just be a 1-commit PR.

mmrosenberg:
trigger build

mrmooney (approved):
This looks fine, but note that we will need to revisit this once we get the 2D signal processing in place, as the value of beta extracted from data would likely change in that case.

Also note that, if we really want to do this correctly, we need to fold in low-energy delta rays into our simulation, or account for them somehow.  When we come back to this problem (requiring changes in LArG4), we will once again need to make changes here.

PR #455, opened 2022-09-25
Decoder now updates trigger type (minimum bias, majority)

PetrilloAtWork:
The field of `sbn::ExtraTriggerInfo` in charge of recording which type of trigger logic is applied (minimum bias, majority) is now correctly set from the decoded information.
Plain and simple, this bit had escaped us.

The PMT decoder tree now also stores that information.

We might also consider to set a bit in `raw::Trigger` with whether the trigger is minimum bias or not... but that is not for this time.


SFBayLaser (approved):
Thank you for fixing this!

jzettle (approved):

mmrosenberg:
trigger build

wesketchum (approved):

PR #457, opened 2022-09-25
Enable by default new PMT simulation and reconstruction

PetrilloAtWork:
These two commits enable by default:
* data-driven single photoelectron response with gain 9.7·10⁶ (`202202` by @ascarpel) in digitisation simulation
* tuned baseline algorithm (also by @ascarpel) in optical hit reconstruction
* charge-to-photoelectrons conversion for optical hits updated according to the gain and SPR above

**This pull request must be merged in `develop` only at the time of cutting the next production release (2022B)**. This is to preserve the physics until then.


SFBayLaser (commented):
The only real issue I can comment on here is that "stage0_multiTPC_icarus_MC.fcl" has already been deprecated and is slated to removal in the new production release. This will be replaced with a "run 2" MC fhicl file that still has to land on a feature branch (coming today!). So... what do we do here?

rennney:
I believe that https://github.com/SBNSoftware/icaruscode/pull/458 will be (must be?) merged before this PR. I expect a bunch of merge conflicts out of it

PetrilloAtWork:
One of the two PR‌ need to go first, and the other needs to be adapted to the result.
I won't be able to work on this on Wednesday.

PetrilloAtWork:
Given the amount of conflicts that #458 can generate, if that PR‌is going to be merged first I will wait until a `develop` branch with that one merged in is available, so that I can concentrate on the conflicts of _this_ PR.

PetrilloAtWork:
Strangely enough, I did not have any GIT merge conflict. But I did have to propagate the changes in new files.
I hope there is no other configuration that I missed...

mmrosenberg:
trigger build

SFBayLaser (approved):
Making it official

PR #458, opened 2022-09-26
Stage0/Stage1 Fhicl files for ICARUS run 2

SFBayLaser:
NOTE THAT THIS IS A MAJOR REORGANIZATION OF ICARUS STAGE0/STAGE1 FHICL FILES!

We have created the set of fhicl files for run 2 which includes (notably) files for running the 2D drift simulation AND the 2D reconstruction on those files. The general naming convention is `stage0_run2_icarus.fcl` and 'stage1_run2_icarus.fcl' with variants for several special cases. At this time I do not claim to have anticipated every variant that people may want but it is important to get the base set out and included in the next release as a starting point. 

Among the major changes, all the fhicl files previously found in icaruscode/Decode/fcl have been depracated. Any fhicl fiiles remaining in this folder (and subfolders) are no longer installed as part of the build process. Instead, the set of fhicl files needed to run on run 1 and run 2 data have been placed in the fcl/reco folder to be consistent with all other production fhicl files. 

I anticipate it make take some time to completely sort out the details but, as mentioned above, it is important to get this into the latest release.

PR #459, opened 2022-09-29
Feature/aheggest crt t1

aheggest:
Use CRT Hit time now saved in T1 (T0 wrt GT) as the CRT PMT Matching module, tiny fix in top CRT Hit reco loop


rennney:
I recommend looking at PR https://github.com/SBNSoftware/icaruscode/pull/458. It will bring changes to fcl structure for stage0  stage. It seems like it is very relevant to this PR and will most  likely break a thing or two (@aheggest @mmrosenberg )

wesketchum:
@SFBayLaser is this critical for the upcoming 2022B release? We're marking 'high' priority right now, but if we should make sure this gets in, update status please.

rennney:
Hi @wesketchum , I will add a note. This should be important as this PR with combination with https://github.com/SBNSoftware/icaruscode/pull/440 should be a first step in reducing cosmic background and  potentially reducing the amount of data that we would need to read from tape for reprocessing 

PS. My concern with this particular PR is that it has all of commits from  https://github.com/SBNSoftware/icaruscode/pull/440 in it... which is bad, because https://github.com/SBNSoftware/icaruscode/pull/440 is not finished yet. So I would advise @aheggest to remake it only with proper changes

aheggest:
hello @rennney, I will make a new PR with only the relavent changes included. Thanks!

PR #460, opened 2022-09-30
Save Trigger Fragment Through Stage 0 Processing

gputnam:
This PR explicitly keeps the ICARUSTriggerV2 fragment through stage0 processing. This is an important piece of information for understanding / processing events that is also very small (~2KB per event). Some of the information in the trigger is already saved in higher level data products, but keeping the fragment ensures that we will not throw away information that would be very painful to re-process from the raw data.

This PR is made to Tracy's branch on https://github.com/SBNSoftware/icaruscode/pull/458. Either this PR should be merged first then that one, or https://github.com/SBNSoftware/icaruscode/pull/458 should be merged and then this PR should be re-based to develop.


SFBayLaser:
I don't want to sound unsympathetic to the problem or saying we should absolutely not do this... however I think we should at least discuss what the "right way" to solve the problem would be. In theory the trigger information is meant to be saved in LArSoft format data products and we should not need to keep DAQ level information. If this is not the case then it should be a very high priority to fix the problem (whatever that solution might end up being). I worry that saving DAQ information in parallel to the existing trigger information is a workaround that can lead to confusion at a higher level when people become confused about where to find the "correct" information. We have a software workshop scheduled after next week's collaboration meeting and I think we should try to discuss this in detail there. There will be at least 2 more releases before data so we have time to understand what we should do here.

gputnam:
I don't think this change is mutually exclusive with also saving the information in higher level products. I think we should do both -- and there already is something that saves some of the fragment in a data product. This PR will ensure our data is resilient against mistakes that might be made. Happy to discuss more at the workshop.

rennney:
@gputnam Hi, we got a merge conflict with recent restructuring of icarus config files. Please have a look

cerati:
Hi @gputnam, can you please comment whether this PR is still active and if this is needed for the imminent production release? If it is, please go ahead and fix the conflicts

SFBayLaser:
Please note that due to fallout from the current effort to deal with new trigger fragments this PR --should-- be included. Will try to update accordingly short.

SFBayLaser:
trigger build

SFBayLaser:
trigger build

mmrosenberg:
trigger build

PR #465, opened 2022-10-10
Updating the run 2 fhicl files previously introduced

SFBayLaser:
This adds fhicl files to keep noise corrected waveforms as well as making sure the correct data products are kept for WC processing.


SFBayLaser:
trigger build

SFBayLaser:
I assume CI tests will crash because of issue with the simple flash matching needed a PR to sort out.

PR #466, opened 2022-10-12
Fixed CAF making jobs with generator systematics

PetrilloAtWork:
Recently I reworked the random number service configuration and at the same time removed the configuration of the random services from the places where it is not supposed to belong.
CAF making is one of them, since it's about translation of data from one format to another.
Unfortunately CAF making currently does more than just translation, and in particular for simulation input it runs a generation-class process, the stochastic extraction of generation parameters and the computation of the weights bridging the extracted parameters to the ones chosen by the experiment as the best ones. These jobs were broken by my changes.

This pull request simply adds the configuration of the required services to the two CAF making jobs including weight generation that I could find.
It is a non-breaking bug fix.

### DIFF
--- a/fcl/caf/cafmakerjob_icarus.fcl
+++ b/fcl/caf/cafmakerjob_icarus.fcl
@@ -2,7 +2,6 @@
 #include "channelmapping_icarus.fcl"
 
 #include "correctionservices_icarus.fcl"
-#include "seedservice.fcl"
 #include "particleinventoryservice.fcl"
 #include "backtrackerservice.fcl"
 #include "photonbacktrackerservice.fcl"
@@ -17,6 +16,7 @@ services:
   ParticleInventoryService:  @local::standard_particleinventoryservice
   BackTrackerService:        @local::standard_backtrackerservice
   @table::icarus_wirecalibration_minimum_services
+  @table::icarus_random_services  # from services_common_icarus.fcl
 
   SpaceChargeService: @local::icarus_spacecharge
 }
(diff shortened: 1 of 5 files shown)


jedori0228:
I checked `fcl/caf/cafmakerjob_icarus_sce_genie.fcl` and `fcl/caf/cafmakerjob_icarus_sce_genie_and_fluxwgt.fcl`, which are the ones that are updated, run fine, but the other MC CAFMaking fcls (e.g., `cafmakerjob_icarus_sce.fcl`) still have issues, should we also add those lines to them?

Ooohu:
It looks like this is targeting the random services only; other issues occurred within `afmakerjob_icarus_sce.fcl` (I am not aware of what are they) may be fixed in a new PR?

jedori0228:
Also `cafmakerjob_icarus_evtw.fcl`

PetrilloAtWork:
I did not realise that
1. the simplified ("fake") reconstruction happens in the module and uses random numbers,
2. the blinding decision is also stochastic (subject of issue SBNSoftware/sbncode#306).

With this, the random engines are needed in both MC and data.

PetrilloAtWork:
Now this branch restores the random number services in all `CAFMaker` jobs. Thank you for spotting the fact that the issue was more spread than I thought.

> It looks like this is targeting the random services only; other issues occurred within `cafmakerjob_icarus_sce.fcl` (I am not aware of what are they) may be fixed in a new PR?

I am not sure which other issues you are referring to. If they are about random streams, I should fix them here.
Otherwise... it depends. Please point me to the report for such issues if there are any.

jedori0228 (approved):
I tested cafmakers and runs fine.

mmrosenberg:
trigger build

PR #469, opened 2022-10-17
Remotes/origin/feature/fpoppi crt

francescopoppi:
This pull request fixes the T0 delays applied to two Top CRT modules (it was underestimated by 10-ish ns).
This PR also corrects the timing of the CRT hits by correcting for the distance of the furthest Channel from the hit coordinates. This is needed since the timestamp is recorded at the fullfiment of the channel-ANDs trigger requirement.


francescopoppi:
Standby please: I found a minor bug, for which I am testing a fix and I need to run it on a small sample of events to be sure it works

rennney:
trigger build

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
```diff
@@ -393,27 +392,41 @@ sbn::crt::CRTHit CRTHitRecoAlg::MakeTopHit(art::Ptr<CRTData> data, ULong64_t Glo
...
-
-    if(adsid_max<8){
-        thit -= (uint64_t)round(abs((92+hitpos.X())*fPropDelay));
-        thit1 -= (uint64_t)round(abs((92+hitpos.X())*fPropDelay));
+    thit -= fSiPMtoFEBdelay;
+    thit1 -= fSiPMtoFEBdelay;
+    //thit1 = (Long64_t)(thit-GlobalTrigger[(int)mac+73]);
+    //thit -= (uint64_t)round(abs((92+hitpos.Z())*fPropDelay));
+    //std::cout<<"Hit position local: X "<<hitpos.X()<<" Z "<<hitpos.Z();
+    if((92-hitpos.X())<=(92+hitpos.Z())) {
```
PetrilloAtWork, line 400: It would be nice to have a comment explaining this magic.

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
```diff
@@ -393,27 +392,41 @@ sbn::crt::CRTHit CRTHitRecoAlg::MakeTopHit(art::Ptr<CRTData> data, ULong64_t Glo
...
+	//std::cout<<"  Thit raw: "<<thit<<" Correction (-hitposX) "<<thit - (uint64_t)round(abs((92+hitpos.X())*fPropDelay))<<"  Bar: "<<maxx*2<<"  "<<maxz*2<<std::endl;
+	thit -= (uint64_t)round(abs((92+hitpos.Z())*fPropDelay));
+	thit1 -= (uint64_t)round(abs((92+hitpos.Z())*fPropDelay));
+    }
+    else {
+	//std::cout<<"  Thit raw: "<<thit<<" Correction (-hitposZ) "<<thit - (uint64_t)round(abs((92+hitpos.Z())*fPropDelay))<<"  Bar: "<<maxx*2<<"  "<<maxz*2<<std::endl; 
+	thit -= (uint64_t)round(abs((92-hitpos.X())*fPropDelay));
+	thit1 -= (uint64_t)round(abs((92-hitpos.X())*fPropDelay));
+    }
```
PetrilloAtWork, line 409: Is my reading right that this code translates into:
```suggestion
    double const corrPos = std::max(-hitpos.X(), hitpos.Z());
    uint64_t const corr = (uint64_t)round(abs((92.0+corrPos)*fPropDelay));
    thit -= corr;
    thit1 -= corr;
```

PetrilloAtWork (commented):
In terms of coding, I don't see anything wrong. I have left a question to make sure my understanding of what the code does is right — and that's what the code is supposed to do.
I would appreciate a comment explaining the magic value `92.0` cm.
Also, you may dare and remove all the old code commented out.

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
francescopoppi: Sure! I will include a comment in the code:
basically the magic number 92 comes from the fact that the bars are 184 cm long, 92 cm is the center.
center of the module is located at coordinates (0,0), this means that when you compute the distance of the hit from the SiPM, you have to translate it by 92 cm.
Concerning why for X is (92-hitpos.X) while for Z is (92+hitpos.Z) this is due to a bug/feature of the geometry, the local position of the modules are defined before rotation operations, due to some weird effect, it is like the readout of the bar happens to be on -92cm in X in local coordinates and then is located correctly after module placement. Anyway, light corrections are applied to local values, not to reconstructed CRT hit positions, therefore this fixes it. Z coordinates was fine.

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
francescopoppi: You are reading right and your proposal looks more elegant (as always), thanks

mmrosenberg:
trigger build

SFBayLaser (approved):
I wonder if at some point there can be a way to move hardcoded numbers from the code into some form of "database"... Somehow it seems more straightforward to update a database as our knowledge increases rather than to modify code... particularly when it might not be apparent what the numbers mean. But this is for the future, I think we should adopt these now.

rennney:
trigger build

rennney:
trigger build

rennney:
trigger build

rennney:
trigger build

PR #470, opened 2022-10-17
Turn off blinding in for MC cafmaker

jacoblarkin:
Turn off newly added blinding code in cafmaker for MC since it's only needed for data. Added to cafmakerjob_icarus.fcl which should propagate to other MC cafmaker fcls.


mmrosenberg:
trigger build

PR #477, opened 2022-11-01
Get CaloSkimmer Working for MC Digits

gputnam:
Switch default caloskim behavior to include tracks w/out a T0. Specify name of MC RawDigit product. Silence missing data products (to automatically handle case where RawDigits are not present).

Depends on https://github.com/SBNSoftware/sbncode/pull/315.

### DIFF
--- a/fcl/reco/Definitions/stage1_icarus_defs.fcl
+++ b/fcl/reco/Definitions/stage1_icarus_defs.fcl
@@ -81,8 +81,8 @@ icarus_stage1_filters:
 
 icarus_stage1_analyzers:
 {
-  caloskimE: @local::caloskim_cryoe_nodigits_goldentracks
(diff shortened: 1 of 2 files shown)


mrmooney (approved):
This is fine to me, just we need to note that this should be changed back if we start running out of disk space.  I also think it is less useful when we are not saving waveform information, which we can not do during nominal production (given raw waveforms are not saved beyond stage0).

mmrosenberg:
trigger build SBNSoftware/icarusalg#61 SBNSoftware/icaruscode#474

mmrosenberg:
trigger build SBNSoftware/icaruscode#474 SBNSoftware/icarusalg#61

SFBayLaser (approved):
This looks like straightforward update to me.

mmrosenberg:

PR #478, opened 2022-11-02
Infrastructure to keep PMT waveforms containing the trigger

PetrilloAtWork:
The core of this pull request is the introduction of the _art_ module `CopyBeamTimePMTwaveforms`.
This module copies the PMT waveforms containing a specified time (e.g. the beam gate opening time) into a new data product, and replicates the supported associations (so far, only the ones with `sbn::OpDetWaveformMeta` — I am not aware of any other produced at Stage0).

## Workflow suggestion

In the larger picture, this module may be part of a workflow that
1. decodes all PMT channels;
2. reconstructs hits from them;
3. duplicates the ones to be preserved as they are (e.g. the ones covering the beam spill);
4. drops from the _art_/ROOT output the full list decoded at point 1, keeping only the one at point 3.

To implement this workflow, this pull request _does_ provide:
1. the module needed for step 3 above, called `CopyBeamTimePMTwaveforms`;
2. a module configuration called `copyPMTonBeam` to copy only the waveforms containing the beam gate opening time

It _does not_ provide:
1. a `Stage0` workflow change including that module in the processing list (e.g.
   ```
   physics.producers.daqPMTonbeam: @local::copyPMTonBeam
   physics.producers.daqPMTonbeam.Waveforms: "daqPMT"
   physics.path: [ @sequence::physics.path, daqPMTonbeam ]
   ```
   ). 
2. a command output dropping 
   ```
   outputs.rootoutput.outputCommands: [
     @sequence::outputs.streamROOT.outputCommands
     , "drop raw::OpDetWaveforms_daqPMT__*"
     , "keep raw::OpDetWaveforms_daqPMTonbeam_*_*"
     ]
   ```
   (note that the example drops only the "regular" PMT channels from `daqPMT` but it keeps the "special" ones)

## Savings

Unfortunately the saving is less than I had hoped: the surviving PMT waveform size is roughly 30% of the full collection (160 MB vs. 540 MB for 50 events from my test BNB majority stream of run `8552`).

The underlying figures are a compression factor of a bit less than `3`, beam gate PMT data for 28 µs and 360 PMT (sampling 2 ns ⇒ 10 MB) and additional 10 µs on 180 channels (1.8 MB) for any additional trigger primitive, of which we have a rate of 5.4 kHz × 2 ms/event = 11/event (the overburden reduced this rate considerably). So the outcome is about a reduction to ⅓.

As a side note, the data collected with this reduction should still be good as overlay to Monte Carlo signal simulation, since the latter will also comfortably fall in the preserved 28 µs and the reconstruction of the other optical hits won't be affected.


SFBayLaser:
Oh! I like this PR! Let me look more closely...

SFBayLaser (approved):
This looks great to me. I agree that now is not the time to modify the production fhicl files (stage 0 and stage 1) just yet since they are being modified for other workflow issues, but we should aim to bring this into them once the release is made.

rennney:
trigger build

PetrilloAtWork:
After quite some attempts, I could finally complete the update to the new build system by linking my module to a library that should have nothing to do with it.
Good to go?

mmrosenberg:

PR #481, opened 2022-11-04
Add CRT ntuple creation module to MC

jzettle:
Add CRT ntuple creator module to MC. This simply runs the same module that is run on the data to fill the ntuples. The reconstructed HitTree is filled and the module runs. Tested on what I understand is recent MC produced for testing WireCell, which is unrelated except for it was produced with a recent icaruscode version so this runs on the right MC for merging here.

### DIFF
--- a/fcl/reco/Stage1/Run2/stage1_run2_icarus_MC.fcl
+++ b/fcl/reco/Stage1/Run2/stage1_run2_icarus_MC.fcl
@@ -43,7 +43,7 @@ physics.analyzers.caloskimE.SimChannelproducer: "largeant"
 physics.analyzers.caloskimW.G4producer: "largeant"
 physics.analyzers.caloskimW.SimChannelproducer: "largeant"
 
-physics.outana:            [ caloskimE, caloskimW, simpleLightAna]
+physics.outana:            [ caloskimE, caloskimW, simpleLightAna, CRTDataAnalysis]
 physics.trigger_paths:     [ reco ]
 physics.end_paths:         [ outana, stream1 ]
 outputs.out1.fileName:     "%ifb_%tc-%p.root"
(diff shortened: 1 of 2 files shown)


SFBayLaser:
I am not sure I understand the intention of this module... We will be outputting CAF files and Calibration tuples as part of the production workflow. As well, the purity monitoring runs and those files are recovered. But I am not sure there is any plan to recover other tuple output that is not part of any of the above? Perhaps it is better to include this output as part of the calibration tuple output? Or perhaps I am not following something correctly?

jzettle:
@SFBayLaser perhaps I did not describe this well. All this PR does is run the module that outputs the CRTHit information into the calibration ntuples when Stage1 is run on MC files. This was being done on data but not on MC. @mrmooney requested this so he could explain more the exact motivation that he had. I believe that I have only added this to the calibration tuple output and anyone is free to test this further if they like. If I have done something else I would appreciate having an offline discussion so I can understand the correct thing to do.

jzettle:
@SFBayLaser did that explain the idea behind the PR and what it does better? Does this make sense to you?

cerati:
Looks like this PR is adding one more module to the outana path. Based on the discussion the goal is add information to the calibration ntuples. Does that module have access to the calibration ntuples? Seems unlikely... but happy to be educated

jzettle:
@cerati I'm not sure I fully understand the details of the fhicl routing so I can't explain well why it works that way but I can confirm it goes into the calibration ntuples. Anyone interested can look at `/icarus/data/users/jzettle/CRTMCTuples/` on a gpvm.

jzettle:
@mrmooney Can you confirm this is meeting the initial need? @SFBayLaser Have I sufficiently answered your questions on what this is doing?

mmrosenberg:
trigger build

mmrosenberg:

mrmooney (approved):
Looks good to me.  Responding to the discussion: it is a very simple addition.  Just add the CRTHit information to calibration ntuples, so we can study CRT-based calibrations identically in MC as we do for data.  I have confirmed that the calibration ntuples indeed have the correct information added.  Thanks!

SFBayLaser (approved):

PR #483, opened 2022-11-05
Trigger emulation for MC simulation

PetrilloAtWork:
A configuration for running the trigger emulation on simulated samples is introduced.
The configuration runs all the steps required (LVDS pairing, trigger window composition, identification of the beam gate and evaluation of the trigger logic) and produces a `std::vector<raw::Trigger>` data product. All the other data products that usually come with the trigger decoding are not provided at the moment.

The configuration `simulate_triggers_icarus.fcl` runs the emulation on a `DetSim` input file with all the thresholds configured in `icarus_triggergate_basic.ChannelThresholds` (from `trigger_icarus.fcl`), and with a few trigger majority settings (`3`, `5`, `10`).

In addition, `Stage0` definitions have been modified so that a minimal emulation is also run with Stage0-MC. There, a single discrimination threshold is applied with a single pattern (defined in `icarus_triggergate_basic.SelectedThreshold` and `icarus_triggergate_basic.SelectedPattern` respectively, both also from `trigger_icarus.fcl`; currently, `400` ADC and majority `5`, reflecting the data configuration for Run1 and likely for Run2).

A relevant note about the beam gates: the modules in the configuration recognises whether the event is from BNB or from NuMI using the `generator` label, and then set a fixed beam gate duration, which is the nominal one plus some margin, similarly to the settings of data runs. If there is no `generator` or if that `generator` does not emit a beam gate, the job will fail.
The settings of the nominal beam gates are, unsurprisingly, in `trigger_icarus.fcl`, as the amount of the extension (`BNB_settings`, `NuMI_settings`, `icarus_triggersimgates_Run1`).

The branch being merged is up to date with the current `develop`.
I have tested `Stage0` on a single GENIE+CORSIKA simulated file after fixing `stage0_run1_icarus_mc.fcl`, which is currently broken in `develop` (`stage0_run2_icarus_mc.fcl` is _also_ broken).


PetrilloAtWork:
Adding @gputnam to the list of reviewers.
In particular from him I would learn if the strategy of having a trigger collection for each event is compatible with the trigger information extraction in CAF. By this pull request, events which do not trigger will still have a single trigger object, with no bit set and a outrageous time (`numeric_limits<double>::max()`, I think).
I can in alternative code something that in some cases emits no trigger object at all leaving the collection empty.

mmrosenberg:
trigger build

mmrosenberg:

gputnam (approved):
Hi Gianluca! Looks good. Sooo with integration for CAF stuff. It looks like this produces a raw::Trigger object. I think CAF mostly relies on the sbn::ExtraTriggerInfo object. Are you planning to also add code to produce an ExtraTriggerInfo, or does the raw::Trigger have everything you need?

PetrilloAtWork:
> Sooo with integration for CAF stuff. It looks like this produces a raw::Trigger object. I think CAF mostly relies on the sbn::ExtraTriggerInfo object. Are you planning to also add code to produce an ExtraTriggerInfo, or does the raw::Trigger have everything you need?

As far as I can tell, `raw::Trigger` does _not_ have everything that is needed. As you have written the extraction code, you may confirm that. `raw::Trigger` only contains relative times, the one of the trigger (the one that in ICARUS data is `1500`) and the beam gate opening time (which will be `149x`), plus some bits telling whether it's NuMI or BNB or neither.
I think I have seen some absolute timestamps around in CAF.
I would benefit from a summary of the trigger information that is needed; my intention was to produce a reduced version of `sbn::ExtraTriggerInfo` (that is, same object, but most fields not filled) with as much information as possible. For the absolute timestamp, I may go either a random timestamp or use the timestamp of the generated event (that's the UTC time of when the _art_ `EmptyEvent` source module created the event). But all this is only vaguely in my mind and nowhere else.

PetrilloAtWork:
I will try to fix the Clang issue. Surely, with Clang 15 out, I would love to use something more modern than Clang 7... but that's not for us to choose.
About the CI issues: that is indeed a workflow issue. I summon @SFBayLaser for a suggestion.

The issue is that the trigger simulation needs to know when and how long the beam gate is, in order to run the simulation.
Given that the events may come either from NuMI or from BNB, my addition to the current workflow goes to read the `sim::BeamGateInfo` produced by `GENIEGen` module to know which one. Unfortunately all the other generator modules (or maybe just `SingleGen` and `CORSIKAGen`) do not emit any beam gate information. Currently there is no way I can see to preserve a single workflow for these different options.
Please advise.

PetrilloAtWork:
I have opened issue #487 on this topic.
Until a decision is taken, I am going to disable the trigger simulation from the standard workflow.

PetrilloAtWork:
Try again...

mmrosenberg:

SFBayLaser (approved):
The expert reviewers all seem happy so good for me.

mmrosenberg:

PR #461, opened 2022-10-03
fhicl files updated for SystTools-based reweight calculator

jedori0228:
This PR adds/updates fhicl files that runs the reweight calculation for  interaction systematic, through modules that are added by  https://github.com/SBNSoftware/sbncode/pull/299.

PR #491, opened 2022-11-23
Trigger simulation in production workflow

PetrilloAtWork:
This pull request attempts to integrate trigger simulation in the standard production workflow, following PR #483 .

The general issue is that trigger emulation needs to know the beam gate, and it turns out that LArSoft generators do not fix a beam gate, except `GENIEGen` which fixes it wrong.

The strategy I propose adds to all generator jobs the creation of a beam gate with the module name `beamgate`. Downstream jobs will be able to learn the nominal beam gate by referring to that data product; trigger simulation does that (potentially it could also extend the gate to be closer to the hardware, but that's for a different pull request).

This pull request is made of two parts:

* reintroduction of trigger simulation in standard Stage0 MC workflow
* addition of a beam gate in pretty much all non-legacy generation configurations, with the following pattern:
    * neutrino generation: see below
    * neutrino interaction with cosmic overlay: follows the same pattern as the previous;
    * cosmic rays: using NuMI beam gate
    * in-time cosmic rays: using a beam gate matching the in-time definition (matching is manual, it can't easily be coded in the configuration)
    * radiogenic background: using NuMI beam gate
    * single particles: using NuMI beam gate
    * external particle list via text file: using NuMI beam gate
    * other physics (e.g. dissonant Higgs): using NuMI beam gate
    * other non-physics: using NuMI beam gate

The rationale of NuMI as the fallback gate is that it makes it easy to remove events whose trigger have fired after 1.6 µs to simulate a BNB beam instead.

Beam gates from `GENIEGen`
---------------------------

The module `GENIEGen` is pretty much the only one that can natively produce a beam gate data product (`std::vector<sim::BeamGateInfo>`). It does it by setting the beam type after decoding the name of the beam flux, the beam opening time from the global offset time and the beam width from the random offset interval (`RandomTimeOffset`). The latter in particular is used by `GENIEGen` to randomise the time of the event.
Unfortunately, that setting also interferes with the interaction time plugin that we do or could use to mirror the actual beam time structure; so when using the latter feature, `RandomTimeOffset` is set to `0`.

Therefore this pull request uses the "copy" option (i.e. create a copy of the GENIE beam gate `generator` which will be called `beamgate`) only if no time plugin is configured.
Otherwise, it sets a fixed beam gate reflecting the configuration job (BNB or NuMI).

Testing
========

I have tested a couple of these configurations to verify they do run and produce the new beam gate data product. I haven't tested the change in Stage0 because both the Monte Carlo configurations (Run1 and Run2) are currently broken for reasons apparently unrelated with this pull request.
I did run the unit tests, which include a grammar test for all the FHiCL configurations.

Reviewer notes
==============

I am asking the review of:
* @SFBayLaser for the general workflow and the modifications to `Stage0`
* @gputnam as author of the in-time and Higgs-like configurations
* @betan009 as a maintainer of the NuMI job configurations

### DIFF
--- a/fcl/configurations/beamgates_icarus.fcl
+++ b/fcl/configurations/beamgates_icarus.fcl
@@ -1,0 +1,147 @@
+#
+# File:    beamgates_icarus.fcl
+# Purpose: Module configurations to ensure the presence of `sim::BeamGateInfo`.
+# Author:  Gianluca Petrillo (petrillo@slac.stanford.edu)
+# Date:    November 22, 2022
+# Version: 1.0
+#
+# This is a collection of simple module configurations to help simulation
+# workflows (generation stage) to ensure the presence of a beam gate information
+# data product with a standard name.
+# 
+# Three main scenarios may be present:
+# 1. the generator module(s) does generate the proper beam gate data product,
+#    but with a non-standard name;
+# 2. the generator module does generate a beam gate data product, but it needs;
+#    adjustment;
+# 3. no generator module generates a beam data data product.
+# 
+# The workflow should rely on a standard beam gate data product name which is
+# not a standard generator name, to support the case (3) where one of those
+# generators is run with the standard generator name.
+# 
+# Thus, I propose here also a standard beam gate definition name, "beamgate",
+# which modules looking for a beam gate data product can confidently refer to:
+#     
+#     BeamGateTag: @local::icarus_standardbeamgate_tag
+#     
+# For the rest, this configuration file supports two of the cases above:
+# 
+# 1. `icarus_standardbeamgate_copy` can "rename" the data product; the label of
+#    the generator module creating the original beam gate can be overridden in
+#    the `BeamGateTag` parameter, and its default value is the standard
+#    "generator" (for example ICARUS GENIE configurations usually use that one).
+#    Example:
+#        
+#      physics.producers: {
+#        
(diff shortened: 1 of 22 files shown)


PetrilloAtWork:
Release managers: is it possible to start an integration test before the approval?
The last attempt to add the trigger simulation resulted in a C.I. test failure that highlighted the need for further work.

wesketchum:
@PetrilloAtWork does this, by chance, also fix #461?

rennney:
trigger build

PetrilloAtWork:
Part of this problem is the fact that new reference files will be needed, the reason being that the new workflow introduces an additional data product in `Gen` stage, and that needs to be present in the input to `Stage0`.
I have flagged the PR with the appropriate label.
I still need to get more confidence that this is the only issue.

PetrilloAtWork:
From the last CI test it seems to me that the PR works as it should. So I just confirm than new reference files are needed, so that the input of `Reco` test contains the new data products from `Gen`.

SFBayLaser (approved):
I was not aware that stage 0 was failing for the MC for both versions and will need to investigate that as a separate problem unrelated to this PR. The change to pick up in stage 0 looks fine to me.

betan009 (approved):

mmrosenberg:

PR #494, opened 2022-12-02
Making a single main PandoraSettings XML

brucehoward-physics:
There used to be a NuMI-specific one with a wider tolerance in x (i.e. a sort of out-of-timeness) before calling something a clear cosmic. This PR moves that setting to a single main XML (so then one would use the same setting for both BNB and NuMI reconstruction), and also adds a line meant to adopt the `cautious` mode of cosmic ray tagging, as studied by Jaesung Kim.

**SOME NOTES (please read)**
1. If I am properly thinking about the situation, this wider window for NuMI should only really needed for simulation even. Actually if we modify the t=0 situation for MC, which is being thought about, then we might not need the wider window at all any more.
2. This PR removes the previous NuMI XML and the NuMI stage1 fhicls that I found -- if we instead want to keep these and deprecate in some fashion, let me know.

*For any requested validation purproses:* Do we currently have the ability to run standard validation tests? If someone can confirm and/or help run the validation test that is appreciated!

Per the discussion this morning, I believe something like this is what we want. For more information see sbn-doc-28976 or the recording of the discussion sbn-doc-28985


rennney:
trigger build

jedori0228 (approved):

PR #495, opened 2022-12-05
A "clean" version of PR 484

SFBayLaser:
As the title indicates, have tried to eliminate the extraneous commits inadvertently included with a rogue merge. 

Reference to the original PR #484.

### DIFF
--- a/icaruscode/Filters/FilterCRTPMTMatching_module.cc
+++ b/icaruscode/Filters/FilterCRTPMTMatching_module.cc
@@ -1,0 +1,555 @@
+////////////////////////////////////////////////////////////////////////
+// Class:       FilterCRTPMTMatching
+// Plugin Type: analyzer (Unknown Unknown)
+// File:        FilterCRTPMTMatching_module.cc
+//
+// Generated at Thu Feb 17 13:31:00 2022 by Biswaranjan Behera using cetskelgen
+// from  version .
+////////////////////////////////////////////////////////////////////////
+
+#include "art/Framework/Core/EDFilter.h"
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Principal/Event.h"
+#include "art/Framework/Principal/Handle.h"
+#include "art/Framework/Principal/Run.h"
+#include "art/Framework/Principal/SubRun.h"
+#include "canvas/Utilities/InputTag.h"
+#include "fhiclcpp/ParameterSet.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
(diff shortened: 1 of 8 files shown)


SFBayLaser:
trigger build

SFBayLaser:
trigger build

mmrosenberg:
trigger build

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -311,9 +313,11 @@ bool icarus::crt::FilterCRTPMTMatching::filter(art::Event &e)
 
     auto const& crtHitList = e.getProduct<std::vector<CRTHit>>(fCrtHitModuleLabel);
 
-    bool Filter = false;
+    bool Cosmic = false;
     int  Type   = 9;
 
+    if (fFilterLevel != "loose" || fFilterLevel != "tight" || fFilterLevel != "medium") fFilterLevel="loose";
```
PetrilloAtWork, line 319: This part should be in the constructor, since it will happen only once.
And I recommend that the default value of the parameter be changed from `"FilterOption"` to `"loose"`.
I also recommend that if an unknown option is passed, an exception is thrown, and otherwise that a message is printed saying that the filter option was not recognised and therefore we are using `"loose"`. Otherwise a user using by mistake `"Medium"` or `"tight "` will end up with the wrong level without ever noticing it.

And actually my real suggestion is, surprise surprise, to use another enumerator:
```cpp
enum class FilterLevel {
  loose,
  medium,
  right,
  defLevel = loose
};
```
which can be parsed like:
```cpp
  std::string const& filterOptStr = p.get<std::string>("FilterLevel", "loose");
  if     (filterOptStr == "loose") fFilterLevel = FilterLevel::loose;
  else if (filterOptStr == "medium") fFilterLevel = FilterLevel::medium;
  else if (filterOptStr == "tight") fFilterLevel = FilterLevel::tight;
  else throw art::Exception{ art::errors::Configuration } << "Filter level not supported: '" << filterOptStr << "'\n";
```

PetrilloAtWork (approved):
@francescopoppi  I have annotated the code with an additional technical comment for your consideration.

----

Overall, I think this is good enough to go in right now.
I would consider though to make it more robust on the case where there are two (or more) different matched CRT hits in the same event. Currently, if I did not get it wrong the latest matched CRT hit decides the fate of the event, and I don't know which order they are in and if this produces a bias.
An alternative would be to keep the count of how many hits are matched per type, and take a decision based on that.

PR #499, opened 2022-12-15
Fixed trigger configuration module configuration

PetrilloAtWork:
Unfortunately I am not in the conditions to even test these changes.
There is an issue with the name configured in `triggerconfig` module label for Run2, and this solution sets the "default" to the latest version and overrides the legacy Run1 configuration.

**The same fix must be applied to the production branch.**

### DIFF
--- a/fcl/reco/Definitions/decoderdefs_icarus.fcl
+++ b/fcl/reco/Definitions/decoderdefs_icarus.fcl
@@ -4,6 +4,7 @@ BEGIN_PROLOG
 
 extractTriggerConfig: { 
                     module_type:        TriggerConfigurationExtraction
+                    DataFragmentType:   ICARUSTriggerV3
 }
 
 extractPMTconfig: {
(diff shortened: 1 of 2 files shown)


mmrosenberg:
trigger build

jzettle (approved):
This is fine. I handled this differently by just changing the Run2 fhicl files to handle the V3 fragment and leaving the default that was in the module to V2. I see no problem changing this but every test I ran in advance of the production release on both new V3 and old V2 data was successful and processed successfully. If I missed some important other change than I apologize as I am not an expert in this module and what in detail it is doing but the data itself that I checked was finishing processing before.

mmrosenberg:

PetrilloAtWork:
I believe that the changes in #502 makes this obsolete, since they change the default value into something that can detect both V2 and V3.
Moving this PR back to draft, waiting for #502/#503 to be merged and see if this is in any way still needed (actually my guess is that at most we will need to _remove_ configuration).

jzettle:
I agree with Gianluca above although I must admit I have not encountered the issue mentioned in this PR in any release I have tried since the first production release where this V3 version got implemented. I was able to successfully run the stage0 processing. I handled this in the current releases by adding `icarus_stage0_producers.triggerconfig.TriggerFragmentType: ICARUSTriggerV3` to the `stage0_run2_icarus.fcl` and letting the `stage0_run1_icarus.fcl ` leave the default in the TriggerConfiguration module as the V2 version. This perhaps is better anyway but I do see now that @PetrilloAtWork used a different fcl parameter `DataFragmentType` that I do not see listed in the TriggerConfiguration module

PetrilloAtWork:
This request has been superseded by other changes.

PR #500, opened 2022-12-20
add kdif-only fcls for higgs portal generator

jamiecd:
These fcl files were added to a production branch of icaruscode (v09_37) in March 2022, and I would like them in develop also so that they continue to be accessible in later code releases.


mmrosenberg:

mmrosenberg (approved):

PR #502, opened 2022-12-22
Trigger decoder tool supporting both V2 and V3 data fragment versions

PetrilloAtWork:
This branch introduces and enables in the standard workflow a trigger decoding tool which is able to process trigger data fragments both from V2 (Run 1) and from V3 (Run 2).
Mixing the two is untested and may cryptically fail.

The new tool is actually just a dispatcher which will read from the run which version is being saved in the input file and will instantiate and utilise the proper single-version existing trigger decoder tool for processing.
In addition, the trigger configuration data product and extractor have been extended to discover and store which DAQ "generator" was used in the run, which then allows the new decoder tool to take the proper action.

Because ultimately the existing decoders are used for the actual processing, it is expected that no breaking change happens. The old tools still work stand-alone, as the old configurations should.

This is a request for merge into `develop`. The commits in this branch should also work with the production branch, which is matter of pull request #503.

### Testing

The configurations `decodeTrigger_icarus.fcl` and `stage0_run2_icarus.fcl` were tested with a Run 1 and a Run 2 file, while `stage0_run1_icarus.fcl` was tested with a Run1 file.

### Reviewers

I suggest @jzettle as the author of all the actual decoders this one delegates to, and @SFBayLaser to check the changes in the workflow.

### DIFF
--- a/fcl/reco/Definitions/decoderdefs_icarus.fcl
+++ b/fcl/reco/Definitions/decoderdefs_icarus.fcl
@@ -71,6 +71,13 @@ decodeTriggerV3: {
                     FragmentsLabel:     "daq:ICARUSTriggerV3"
                     DecoderTool:        @local::TriggerDecoderV3Tool
 }
+
+decodeTriggerAutodetect: {
+                    module_type:        DaqDecoderICARUSTrigger
+                    FragmentsLabel:     ""
+                    DecoderTool:        @local::TriggerDecoderAutoTool
+}
+
 copyPMTonBeam: {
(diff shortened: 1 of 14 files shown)


jzettle (approved):
Same comment as on #503. This makes sense to me and is something I had wanted to try to do at some point to make things easier and not require different fhicl paths for the different trigger types. I did not test anything on my own yet but Gianluca did so I am confident it is working as intended and the data comes out of the decoder.

mmrosenberg:

PR #504, opened 2022-12-23
Stage0 configuration only keeps waveforms at beam gate time

PetrilloAtWork:
Following PR #478, this pull request enables retaining only the waveforms including the beam gate.

Technically, all PMT decoding and reconstruction proceeds through the data product `daqPMT` as before, but a module is appended that copies the regular waveforms containing the regular beam gate opening time from `daqPMT` into `daqPMTonbeam`. Then the regular PMT waveforms from `daqPMT` are all dropped.
Note that this affects only the "regular" waveforms, i.e. the ones associated with a PMT. The waveforms on the other readout channels (the "16th channel") are in different data products and are retained.
This behaviour has been encoded in the default `Stage0` Run 2 job configuration `stage0_run2_icarus.fcl`.

The job configuration `stage0_run2_raw_icarus.fcl` has been modified not to run the waveform duplication and not to drop any PMT waveform.

Implementation-wise:
1. the _default_ workflow definitions have changed to include also the copy of the gates;
2. the standard `Stage0` configuration has been modified to drop the optical waveforms from the decoder;
3. the "raw" configuration has been modified to run a dummy producer instead of the one copying the gates.

Because of this, any other configuration using the default workflow will have the duplicated waveforms, and possibly also the original ones if they were not dropped. If this is not acceptable, a different implementation is needed.

**Note**: because the waveforms on the beam gate are duplicated, the job memory usage is going to increase accordingly (the data product is 360×26 µs = 9.36 MB large).

This is a request for the `develop` branch. A twin request will be opened for the production branch.

### Reviewers

* @SFBayLaser to check the workflow changes

### DIFF
--- a/fcl/reco/Definitions/decoderdefs_icarus.fcl
+++ b/fcl/reco/Definitions/decoderdefs_icarus.fcl
@@ -71,6 +71,7 @@ decodeTriggerV3: {
                     FragmentsLabel:     "daq:ICARUSTriggerV3"
                     DecoderTool:        @local::TriggerDecoderV3Tool
 }
+
 copyPMTonBeam: {
                     module_type:        CopyBeamTimePMTwaveforms
                     Waveforms:          @nil # must override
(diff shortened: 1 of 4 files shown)


mmrosenberg:

PetrilloAtWork:
Deleted the branch from the wrong remote repository — should be fixed now.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_64_01 SBNSoftware/sbncode@v09_64_01 SBNSoftware/sbnanaobj@v09_20_01 SBNSoftware/sbndaq_artdaq_core@v1_04_02of0 SBNSoftware/sbnobj@v09_15_02

PR #508, opened 2023-01-05
Fixing CRTT0Matching_module

tboone0042:
Replaces a now-faulty if statement to correctly search for the default value of DCA for the closest CRT Hit and skip if that default is found, which is a sign no match was found and thus will return null pointers that have caused seg faults.


tboone0042:
this one should be based on _develop_

PetrilloAtWork (changes requested):
The comment is the same as for PR #509, except that here there is no issue with the package versions.

I recommend that Tyler completes the work on that PR before this one be merged.
The reason is just that if he decides to change anything in there because of the review, those changes should be replicated here too.

If this PR stays identical to #509 then it's good to go at the same time the other one goes too.

mmrosenberg:

PR #513, opened 2023-01-12
GENIE event generator list set back to Default

PetrilloAtWork:
Tune (chosen with the UPS qualifier of `genie`, currently `G1810a0211a:e1000:k250`) in GENIE v3 includes MEC, and the old event list should not be used any more.

Many thanks to @mmrosenberg, @sjgardiner and Stephen Dolan for the support.


mmrosenberg (approved):
Thanks again Gianluca!

mmrosenberg:

PR #515, opened 2023-01-12
Update WC parameters to match ICARUS simulation configuration

gputnam:
Updates a few parameters:

-G4RefTime is set to the correct value
-A delta to the "time_offsets"'s is added to account for electronics shaping time
-The gain is set to match data (see https://github.com/SBNSoftware/icaruscode/pull/356)
-The drift velocity is set to the more precise value found in data

### DIFF
--- a/icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
+++ b/icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
@@ -10,9 +10,9 @@ base {
     // Transverse diffusion constant
     DT :  8.8 * wc.cm2/wc.s,
     // Electron lifetime
-    lifetime : 9.6*wc.ms,
+    lifetime : 3.5*wc.ms,
     // Electron drift speed, assumes a certain applied E-field
-    drift_speed : 1.6*wc.mm/wc.us, // at 500 V/cm
+    drift_speed : 0.15756*wc.mm/wc.us, // at 500 V/cm
   },
   daq: super.daq {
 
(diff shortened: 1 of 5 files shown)


icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
```diff
@@ -10,9 +10,9 @@ base {
     // Transverse diffusion constant
     DT :  8.8 * wc.cm2/wc.s,
     // Electron lifetime
-    lifetime : 9.6*wc.ms,
+    lifetime : 3.5*wc.ms,
```
rennney, line 13: This electron liftime should not impact simulation, as it is set with https://github.com/SBNSoftware/icaruscode/blob/c988c8617a4c104e5c446517f9719c8d3bb6fd06/icaruscode/TPC/ICARUSWireCell/wcls-multitpc-sim-drift-simchannel.fcl#L60

rennney (changes requested):
Request one very simple fix

icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
```diff
@@ -10,9 +10,9 @@ base {
...
     // Electron drift speed, assumes a certain applied E-field
-    drift_speed : 1.6*wc.mm/wc.us, // at 500 V/cm
+    drift_speed : 0.15756*wc.mm/wc.us, // at 500 V/cm
```
rennney, line 15: Need change to 1.5756 to match units

rennney (changes requested):

icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
gputnam: Good catch! Thanks -- fixed it.

rennney (approved):

mmrosenberg:

mrmooney (approved):
Looks fine, though it would be good to know for sure which parameters are, and aren't, used.  Lifetime isn't used, but what else?

PR #523, opened 2023-02-01
Trigger decoder made robust to inconsistent trigger beam gate configurations

PetrilloAtWork:
This pull request has a main and an ancillary part.

# Trigger decoding workaround

There have been some (inconsistent) trigger configurations with a veto duration longer than the beam gate. This has no consequence in the trigger configurations not using the beam gate ("calibration"/zero bias gates, and any minimum bias gate). One method of `icarus::TriggerConfiguration` though _requires_ the veto to be shorter than the beam gate (via assertion). The trigger decoder uses that method, and therefore fails to decode the runs where that inconsistent configuration was used.
While there is being a campaign to correct the new configurations, in order to decode the old ones I am proposing a trigger decoder workaround. This is of technical nature, tailored to avoid specifically that failure.

The change is simply that the veto and raw gate durations are checked first and the touchy method is called only if the former is smaller than the other, otherwise a gate width value of `0` is used.

### Testing

Decoding has been tested on events of runs `9260` and `9304`, which show the inconsistent configuration with trigger versions 2 and 3 respectively.

# PMT decoding standalone job configuration update

The configuration `decodePMT_icarus.fcl` and a couple of derivative have been updated to use trigger version autodetection, as the Stage0 decoding for Run 2 is already doing. The two derivatives have been "promoted" from the `deprecated` directory.

In addition, some configuration files have been reorganised.
* `dump` configurations for PMT raw data fragments and for triggers have been moved to `fcl/utilities`, which already hosts other data product dumping job configurations;
* the `decodePMT_icarus` class of configurations join the trigger-only decoding in `fcl/reco/Stage0/Run2/partial` (some may work also with Run 1 though);
* duplicate configurations under the `icaruscode/Decode/fcl` are removed.

### Testing

The five "new" or updated configurations have been successfully tested on a single Run 2 data file.

In addition, unit tests have been successfully run (including the test of the formal correctness of all FHiCL configuration files).

## Reviewers

* @jzettle as the author of the trigger decoder, to cross-check the changes to the trigger decoder
* @SFBayLaser as the maintainer of the configurations and their storage tree


jzettle:
I'm not sure I follow the need for this. Have there been other discussions within a group that I am not privy to? What did we do to the configurations that caused a change in the veto gate or for it to be longer than the beam gate? Is there a necessary change in the data taking configuration (i.e. at the online level) or the data processing configuration (i.e. at the stage0 fcl level)? I think the code changes themselves look fine on a technical level but I'm not sure I understand the impetus for them or what we have noticed that caused this change to be necessary.

PetrilloAtWork:
The starting point is that there are runs already acquired with a "calibration" trigger that were configured with a veto window larger than the whole beam gate. The decoding of these run fails with an assertion failure error.

Donatella is going to correct the known wrong configuration _for the future acquisitions_, and we need a way to deal with the runs already acquired. The workaround implemented here is only one of the possible workarounds.
Other that come to my mind are to remove the assertion in the data product, have a workaround like the one here but enabled only by configuration (if we know the run we are decoding needs it), or changing the configuration extractor module that produces the data product where the configuration is stored so that when this condition happens it sets the gate duration to `0` (this requires some explanation: the decoder gets the beam gate duration from a data product, that data product code contains the assertion based on its data content, i.e. raw beam gate duration and veto time, and there is a module which creates that data product; here I am hypothesising to change the latter).

jzettle (commented):
I have no issues with the idea behind this change.

mmrosenberg:
trigger build

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_66_00

PR #529, opened 2023-02-08
fhicl files for MC processing chain

SFBayLaser:
Assuming we move to the 2D drift simulation as the default, this updates the fhicl files needed to run the processing chain. 
NOTE: this also includes a reversion of the decoder access to the channel mapping database which uses the database value for a board's slot vs the value included in the data.


mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_66_00

PR #530, opened 2023-02-13
Store peak time and start time in optical hits

PetrilloAtWork:
A change was introduced some time ago to decide whether to set the start time or the peak time in `recob::OpHit::PeakTime()`. Start time was considered to be very important but storing it instead of the peak time was, of course, a hack.
Some time later that hack was rectified by LArSoft/larana#19 by adding a variable for the start time, and letting each quantity have their own.
The default `OpHitFinder` configuration was set in a legacy mode so that the behaviour would not change, i.e. the start time would still overwrite the peak time (and incidentally the new start time variable would also store the start time).

This configuration change sets the optical hit finder configuration so that [`PeakTime()`](https://nusoft.fnal.gov/larsoft/doxsvn/html/classrecob_1_1OpHit.html#af11bf47d83b22b92149081ff30eccf63) again returns the peak time; the start time is still available in its own [`recob::OpHit::StartTime()](https://nusoft.fnal.gov/larsoft/doxsvn/html/classrecob_1_1OpHit.html#a4190af5f7a0fd1d287f21bb5d71c055e).

To be explicitly: until now ICARUS code was effectively using the optical hit start time, but it was packaged under cover (`PeakTime()`).

## Impact on the current code

I am changing most of ICARUS code to use `StartTime()` instead of `PeakTime()`, so that the change is effectively transparent. In some code instead of replacing the peak time value in a ROOT tree I have added another branch specifically for the start time.
The optical hit correction module has been updated to propagate and correct also the start time.

Note that the start time assigned by the `SlidingWindow` algorithm is not that great either, since it includes a fixed, small amount of baseline (presumably) before the hit.

Standard CAF does not have optical hit information, so I am not expecting an impact specifically there (while the simple flash algorithm that is used in CAF does use optical hits; the configuration was updated accordingly).

As an exception, the following modules were _not_ updated:
* `icaruscode/PMT/OpReco/arxiv/OpDetWFDump_module.cc`: it's archived
* `icaruscode/PMT/OpReco/ICARUSOpHitTuple_module.cc`
* `icaruscode/PMT/TrigInfo_module.cc` (I think it's obsolete)

## Other changes in this PR

There are two loosely related addition in this pull request.

### Optical flash time

I am introducing a small utility that allows to choose at run-time which optical hit time to use. I have changed our `ICARUSOpFlashFinder` to use that utility, and set its default to `"Start"` (for `StartTime()`), since that is what was effectively used before and is what makes the most sense to use so far.
The utility should be moved to `sbncode` in the future.

### Rise time

In that same LArSoft PR (LArSoft/larana#19) a way to estimate the rise time of the optical hit was introduced, and a standard algorithm was provided.
While the standard algorithm is not perfect, it's fast enough and I am activating it (otherwise the rise time will be unassigned and have weird values — the standard algorithm can't be worse than that).

## Reviewer notes

If I did all correctly, all changed so that nothing could change (apart from the three exceptions noted above).

Recommended reviewers:
* @amenegol 
* @francescopoppi
* @JackSmedley (whom I can't persuade GitHub to add as an official reviewer)
* @fjnicolas (SBND, author of the original PR)


JackSmedley:
Reading through the [SimpleFlashMatch scheme](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/FlashMatch/FlashPredict_module.cc), it looks like one time values is considered for each OpHit, and this can be optionally set to `RiseTime`, `StartTime`, or `PeakTime`. If my understanding of lines 44-47 is correct, then it looks like this is by default using `RiseTime`, so I would expect no change here from your PR. 

I can't think of any other variables stored in the CAFs that might be impacted by this, other than the OpFlash. Gianluca I am sure you have your eyes on this already, but this uses `PeakTime` and may be one of those cases that requires some "disentangling."

PetrilloAtWork:
> Gianluca I am sure you have your eyes on this already, but this uses PeakTime and may be one of those cases that requires some "disentangling."

For the future record: I changed the plan, biting it and changing all `PeakTime()` to `StartTime()` as explained in the _new_ description above. Jack's comment referred mostly to when I was moving from `StartTime()` to `PeakTime()` wondering what would happen.

So yes, I actually _extended_ `ICARUSFlashFinder` so that we can choose which time to use (`"StartTime"` by default in this PR). I also (belatedly) saw the flash finding for CAF, and what Jack described. That is also set to `StartTime` now (I hope you are wrong that rise time is what is used by default, since it's bugged).

PetrilloAtWork:
Ready for review!
I'll mention this at the next optical reconstruction meeting.

francescopoppi (approved):
This PR includes major improvements. It does not break exiting analysis tools (to my knowledge) and will open the possibility to several interesting and important analysis.
I see that at CAF stage opHits are not stored, but I would expect, in the future to include for a given OpFlash all the three times associated with the reference OpHit (should be the one with the highest pulse heigth, right?). e.g. OpFlash.StartTime OpFlash.RiseTime OpFlash.PeakTime

gputnam (approved):

fjnicolas (approved):

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #531, opened 2023-02-16
Renamed trigger decoder tool classes

PetrilloAtWork:
Three ICARUS trigger decoder tools have the same class name.
While this is not a problem as long as only one of their libraries is loaded at a time (which is currently the case although it does not have to be in the future), this causes confusioni in the [automatic documentation rendering](https://icarus-exp.fnal.gov/at_work/software/doc/icaruscode/latest/classdaq_1_1TriggerDecoder.html) by Doxygen.

These changes modify the class names of two of the tools, and should have no other observable consequence.

A sibling request #532 covers the production branch.


jzettle (approved):

mmrosenberg:
trigger build

PR #533, opened 2023-02-20
Correct time shift in Wire-Cell truth info

gputnam:


mmrosenberg:
trigger build

brucehoward-physics (approved):
From a practical perspective this value _works_ in that I was able to arrive at something with seemingly fairly reasonable truth matching.

Others (@varanini @rennney etc.) may have considerations as to why 1 us is used over 1.3 us. It's also worth noting that 1 us to my understanding is not really the result of an optimization, just that it seemed like it was a bit less than 1.3 and this worked (@gputnam please correct if wrong).

So one can possibly arrive at a better number, but from my perspective I am happy to accept this number. But, I'm NOT an expert in WireCell (hence a reason for tagging @rennney)

varanini:
> From a practical perspective this value _works_ in that I was able to arrive at something with seemingly fairly reasonable truth matching.
> 
> Others (@varanini @rennney etc.) may have considerations as to why 1 us is used over 1.3 us. It's also worth noting that 1 us to my understanding is not really the result of an optimization, just that it seemed like it was a bit less than 1.3 and this worked (@gputnam please correct if wrong).
> 
> So one can possibly arrive at a better number, but from my perspective I am happy to accept this number. But, I'm NOT an expert in WireCell (hence a reason for tagging @rennney)

What relation we expect between the shaping time and the shift ? In @gputnam 's commit i see that also in the slower response case the shift was somehow smaller than the field response time (2.0 vs. 2.2, now ~1.0 vs. 1.3).
I think we should clarify this point before making a decision

PR #536, opened 2023-02-22
Mtsp develop

cerati:
This PR includes the code changes that mainly correspond to what presented in this presentation from last Fall:
https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=28239

The changes make two modules (`MCDecoderICARUSTPCwROI` and `ROIFinder`) thread safe. It also includes a bug fix in `TPCNoiseFilter1DMC` which would cause crashes when running multithreaded.

Since we still run the 1D signal processing (as well as the 2D one), these changes can give some ~20% speedup when running `stage0_run2_wc_icarus_mc.fcl`.  The speedups for the actual modules that are multithreaded are larger, but the overall process time is dominated by 2D signal processing, which is not multithreaded yet.

Note that in order to get these speedups, jemalloc needs to be enabled:
```
setup jemalloc v5_3_0b -q e20:prof
export LD_PRELOAD=$JEMALLOC_LIB/libjemalloc.so
```
and `lar` commands should include `--nschedules 1 --nthreads $t`, where `$t` refers to the number of threads we want to use. This PR does not include any change to production job scripts, which would need to be updated in a separate PR if we need to enable these features.


mmrosenberg:
trigger build

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
```diff
@@ -498,110 +499,54 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&
...
-
-        // Commence looping over raw digits
-        for(const auto& rawDigit : rawDigitVec)
-        {
-            raw::ChannelID_t channel = rawDigit->Channel();
-
-            ChannelToBoardWirePlaneMap::const_iterator channelToBoardItr = fChannelToBoardWirePlaneMap.find(channel);
-
-            if (channelToBoardItr == fChannelToBoardWirePlaneMap.end())
+        const unsigned int dataSize = art::Ptr<raw::RawDigit>(digitVecHandle,0)->Samples(); //size of raw data vectors
```
PetrilloAtWork, line 502: How is this different from
```suggestion
        const unsigned int dataSize = digitVecHandle->front().Samples(); //size of raw data vectors
```
?

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
```diff
@@ -498,110 +499,54 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&
...
+
+	std::map<unsigned int, std::vector<const raw::RawDigit*> > boardToRawDigitMap;
+	for(size_t idx = 0; idx < digitVecHandle->size(); idx++) {
+	  const raw::RawDigit* rawDigit = &digitVecHandle->at(idx);
```
PetrilloAtWork, line 506: I **suggest** that this loop be made simpler:
```suggestion
	for(const raw::RawDigit& rawDigit: *digitVecHandle) {
```
having `rawDigit` a reference. The two places where it is used below will need to be adapted.

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
```diff
@@ -498,110 +499,54 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&
...
-            {
-                processSingleImage(clockData, boardInfo.second, boardWireCountMap[boardInfo.first], concurrentRawDigits, concurrentRawRawDigits, coherentRawDigits, concurrentROIs);
-            }
-        }
+	  unsigned int board = channelToBoardItr->second.first;
+	  auto mapIter = boardToRawDigitMap.find(board);
+	  if (mapIter != boardToRawDigitMap.end()) {
+	    mapIter->second.push_back(rawDigit);
+	  } else {
+	    boardToRawDigitMap.insert({board,std::vector<const raw::RawDigit*>{rawDigit}});
```
PetrilloAtWork, line 521: This should be equivalent to:
```suggestion
          boardToRawDigitMap.try_emplace(board).first->push_back(&rawDigit);
```
(also added address operator `&` for my previous suggestion).
You can argue whether it is an improvement or not (`first` is quite cryptic...).

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
```diff
@@ -498,110 +499,54 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&
...
-        for(const auto& rawDigit : rawDigitVec)
-        {
-            raw::ChannelID_t channel = rawDigit->Channel();
-
-            ChannelToBoardWirePlaneMap::const_iterator channelToBoardItr = fChannelToBoardWirePlaneMap.find(channel);
-
-            if (channelToBoardItr == fChannelToBoardWirePlaneMap.end())
+        const unsigned int dataSize = art::Ptr<raw::RawDigit>(digitVecHandle,0)->Samples(); //size of raw data vectors
+
+	std::map<unsigned int, std::vector<const raw::RawDigit*> > boardToRawDigitMap;
```
PetrilloAtWork, line 504: Why do you use a map rather than a vector here?

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
```diff
@@ -498,110 +499,54 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&
...
+          ChannelArrayPair chanArr;
+          for (const auto rawDigit : rawDigitVec) {
+	    // Declare a temporary digit holder and resize it if downsizing the waveform
+	    raw::RawDigit::ADCvector_t rawDataVec(dataSize);
+	    // Decompress data into local holder
+	    raw::Uncompress(rawDigit->ADCs(), rawDataVec, rawDigit->Compression());
+	    // Fill into the data structure
+            raw::ChannelID_t channel = rawDigit->Channel();
+            unsigned int planeIdx       = fChannelToBoardWirePlaneMap.find(channel)->second.second.second;
+	    icarus_signal_processing::VectorFloat boardDataVec(dataSize);
```
PetrilloAtWork, line 543: It looks like this initialization + conversion + copy can be compacted into:
```suggestion
	    icarus_signal_processing::VectorFloat boardDataVec(rawDataVec.cbegin(), rawDataVec.cend());
```

PetrilloAtWork (commented):
I think I should yield to Tracy for the details.
I have dropped a few comments on coding.

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: I guess they are the same. Testing your version and if it works I will commit it

cerati:
General comment wrt @PetrilloAtWork 's suggestions. I have not tried to change the style or conventions from the previous version, just restructured things for multi-threading over the boards.

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: I think this will require to change more that two places, as the vector in the map will have to change type. Since this is how the code was written before, I think this is not in the scope of the current changes

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: yes it could be a vector of vectors, but a map allows to keep track of which board we are processing (in case of debugging)

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: yeah, why first? I would have thought second as in line 519

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: thanks

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: actually it should be `boardToRawDigitMap.try_emplace(board).first->second.push_back(&rawDigit);` and I am also not sure it is an improvement

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: ah no, this is actually doable without changing the map type (which would be a nightmare)

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
cerati: so I am not going to make this change, but it can be done if you have a strong opinion

SFBayLaser (approved):
These changes look to be doing what we expect! Thanks!

mmrosenberg:
trigger build

PetrilloAtWork (approved):
Annuntio vobis gaudium magnum.

PR #542, opened 2023-03-03
Fixing fhicl file reported issues

SFBayLaser:
Issues that have been reported include the flash matching having been commented out for stage 1 running and that the default MC stage1 processing was requesting timing offsets. 
If others will add them here...


mmrosenberg:
trigger build

brucehoward-physics:
Looks good to me. A few comments below but not holding things up for the answers.

@SFBayLaser are there other places where FM was removed and we want it to be re-enabled? In any case that can be a separate PR if it needs to be.

Also, I noted in a few other places but will also say here -- I think we _probably_ will want the metrics remade for the FMatch.

brucehoward-physics (approved):
See previous comments. But approving.

cfarnese (approved):

PR #543, opened 2023-03-04
Speed optimisation for `GnocchiCalorimetry` tools

PetrilloAtWork:
I propose here three optimisations to speed up the normalising tools used in `GnocchiCalorimetry` module (copy setup data only once per event, avoid data structure copies, use faster lookup algorithm).
One of the three optimisations requires an interface change, which is the subject of pull request LArSoft/larreco#54 to `larana`.
The bottlenecks have been identified with ForgeTools Map profiler.

Overall in my test I observe a conspicuous speedup (larger than ×20), with an event being processed in 0.15" (compared to 5" before the changes). This makes a big difference in a workflow I am running.

I haven't provided a branch for production. If this is needed, a patch release of `larreco` will also be needed, unless we give up the optimisation that requires it or a different approach is used (possible, but less elegant).

Testing
------------

I have tested the two more complicate optimisations by running a version of both codes at the same time and comparing the result (exception thrown if discrepant). I tested that on 50 data events from run `8650`.

Reviewers
-----------------

* @gputnam: original author of the code

### DIFF
--- a/icaruscode/TPC/Calorimetry/NormalizeDriftSQLite_tool.cc
+++ b/icaruscode/TPC/Calorimetry/NormalizeDriftSQLite_tool.cc
@@ -22,6 +22,8 @@
 
 // C++
 #include <string>
+#include <optional>
+#include <cassert>
 
 namespace icarus {
   namespace calo {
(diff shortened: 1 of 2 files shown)


gputnam (approved):
Looks good! Thanks for taking the time to improve this @PetrilloAtWork.

mmrosenberg:
trigger build

mmrosenberg:
trigger build LArSoft/larreco#54

PR #548, opened 2023-03-09
fhicl file updates as agree at Analysis meeting of 3/9/23

SFBayLaser:
At the ICARUS Analysis Meeting of 3/9/23 it was agree that we would run the stage 0 processing without the 2D deconvolutuion and so this has been removed from the stage0 fhicl files. Note that for MC production we should use "stage0_run2_icarus_mc.fcl". 
In addition, in order to reduce the memory footprint, the stage 1 files have been modified to remove the "SelectEvents" qualifier for root output. For stage 1 the file for MC should be "stage1_run2_icarus_MC.fcl". 
This PR also introduces a fhicl definition file for using the database to recover channel status information. Currently the default usage is unchanged from previous, but this allows upgrading in the near future. 
Finally, there is also the introduction of a tool for using wavelets to do ROI finding. This code is unused but will facilitate future development...

### DIFF
--- a/fcl/configurations/database_icarus.fcl
+++ b/fcl/configurations/database_icarus.fcl
@@ -1,0 +1,28 @@
+#
+# File:    database_icarus.fcl
+# Purpose: Configuration for services delivering database related information.
+# Date:    March 7, 2023
+# Author:  Gianluca Petrillo (petrillo@fslac.stanford.edu) <== we'll blame him
+#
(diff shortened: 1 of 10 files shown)


SFBayLaser:
trigger build

cerati (approved):
Looks good to me, and tested on top of v09_68_00 it works as intended.

PR #549, opened 2023-03-14
Feature/howard duplicate hit filter tool v09 67

brucehoward-physics:
Depends on https://github.com/PandoraPFA/larpandora/pull/28

This adds to icaruscode an alternative HitCollection algorithm, as an art tool, that we can use with pandora to try to avoid it getting passed two copies of some of the hits from Cluster3D.

In addition to just generally liking that people have eyes on the code to make sure they don't spot issues, I have a few questions:

1. I'm not super familiar with, let's say, the "right" way to put together the cet stuff in CMakeLists to get the code to build. This code at least builds but may well be bringing in libraries it doesn't need or doing something else impractical or even unwanted. Is there guidance on this or is what's here seemingly ok?

2. I realized that the tool uses a few things which aren't expressly `#include`d like `fhicl::ParameterSet` or `lar_pandora::HitVector`, for examples. The latter should be included via `LArPandoraHitCollectionTool.h` bringing in `LArPandoraHelper.h`. Not sure off-hand where `ParameterSet` is brought in from the way it is now. Is it preferred that I explicitly add includes for `LArPandoraHelper.h` and `ParameterSet.h`?

### DIFF
--- a/icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
+++ b/icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
@@ -2,37 +2,12 @@
 # where should the scripts/..xml file be installed?  Perhaps in bin?
 
 set( TOOL_LIBRARIES
-	larpandora::LArPandora
-	larpandora::GeometryComponents
-	larpandoracontent::LArPandoraContent
-	lardata::AssociationUtil
-	lardata::Utilities
-	lardataobj::AnalysisBase
+	larpandora::LArPandoraInterface
 	lardataobj::RecoBase
-	lardataobj::Simulation
-	larcoreobj::SimpleTypesAndConstants
-	nusimdata::SimulationBase
-	art::Persistency_Common
 	canvas::canvas
-	PandoraPFA::PandoraSDK
-	larpandora::GetDetectorType
-	larreco::ClusterFinder
-	larreco::ClusterParamsImportWrapper
-	larreco::RecoAlg_ClusterRecoUtil
-	larevt::ChannelStatusProvider
-	larevt::ChannelStatusService
-	larsim::MCCheater_ParticleInventoryService_service
-	lardata::DetectorClocksService
-	lardata::DetectorPropertiesService
-	larcore::Geometry_Geometry_service
-	larcorealg::Geometry
-	art_root_io::TFileService_service
 	art::Framework_Principal
-	art::Framework_Services_Registry
 	messagefacility::MF_MessageLogger
 	cetlib::cetlib
-	cetlib_except::cetlib_except
-	PandoraPFA::PandoraSDK
 )
 
 cet_build_plugin(LArPandoraHitCollectionToolICARUS art::tool LIBRARIES ${TOOL_LIBRARIES})
--- a/icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
+++ b/icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
@@ -1,7 +1,44 @@
 
 # where should the scripts/..xml file be installed?  Perhaps in bin?
 
+set( TOOL_LIBRARIES
+	larpandora::LArPandora
+	larpandora::GeometryComponents
+	larpandoracontent::LArPandoraContent
+	lardata::AssociationUtil
+	lardata::Utilities
+	lardataobj::AnalysisBase
+	lardataobj::RecoBase
+	lardataobj::Simulation
+	larcoreobj::SimpleTypesAndConstants
+	nusimdata::SimulationBase
+	art::Persistency_Common
+	canvas::canvas
+	PandoraPFA::PandoraSDK
+	larpandora::GetDetectorType
+	larreco::ClusterFinder
+	larreco::ClusterParamsImportWrapper
+	larreco::RecoAlg_ClusterRecoUtil
+	larevt::ChannelStatusProvider
+	larevt::ChannelStatusService
+	larsim::MCCheater_ParticleInventoryService_service
+	lardata::DetectorClocksService
+	lardata::DetectorPropertiesService
+	larcore::Geometry_Geometry_service
+	larcorealg::Geometry
+	art_root_io::TFileService_service
+	art::Framework_Principal
+	art::Framework_Services_Registry
+	messagefacility::MF_MessageLogger
+	cetlib::cetlib
+	cetlib_except::cetlib_except
+	PandoraPFA::PandoraSDK
+)
+
+cet_build_plugin(LArPandoraHitCollectionToolICARUS art::tool LIBRARIES ${TOOL_LIBRARIES})
+
 install_fhicl()
+install_source()
 
 add_subdirectory(scripts)
 
--- a/icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
+++ b/icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
@@ -43,8 +43,7 @@ namespace lar_pandora {
 
     for (unsigned int i = 0; i < theHits->size(); ++i) {
       const art::Ptr<recob::Hit> hit(theHits, i);
-      if ( channelTimeHits.find(std::make_pair(hit->Channel(), int(hit->PeakTime()))) == channelTimeHits.end() ) channelTimeHits[std::make_pair(hit->Channel(), int(hit->PeakTime()))] = 1;
-      else continue;
+      if ( std::exchange( channelTimeHits[std::make_pair(hit->Channel(),int(hit->PeakTime()))], 1 ) ) continue;
       hitVector.push_back(hit);
     }
 
--- a/icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
+++ b/icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
@@ -1,0 +1,56 @@
+/**
+ *  @file  larpandora/LArPandoraInterface/Tools/LArPandoraHitCollectionToolICARUS_tool.cc
+ * 
+ *  @brief Define class for hit collection tools and implements ICARUS tool, with a duplication filter
+ * 
+ */
+
+#include "art/Utilities/ToolMacros.h"
+
+#include "larpandora/LArPandoraInterface/LArPandoraHitCollectionTool.h"
+
+#include "art/Framework/Principal/Event.h"
+#include "lardataobj/RecoBase/Hit.h"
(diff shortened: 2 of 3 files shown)


SFBayLaser (commented):
Don't forget to modify the CMakeLists.txt file in the folder above this one so it includes this folder. In addition, we'll need to include the fcl file into the ICARUS pandora definitions file and make sure this tool is put in place by default.

brucehoward-physics:
Hi @SFBayLaser this work is in a folder that already exists so I don't think anything else is required to be added above it? Maybe I'm missing something though, can you clarify.

What is meant by the fcl in this case? This line https://github.com/SBNSoftware/icaruscode/blob/feature/howard_duplicateHitFilterTool_v09_67/icaruscode/TPC/Tracking/ICARUSPandora/pandoramodules_icarus.fcl#L22 should add it to `icarus_basicpandora` which I think should bring it into the pandora definitions we use. Maybe I misunderstand the comment? Sorry if so!

SFBayLaser:
Yes, sorry, I was looking in the wrong place and wondering why I didn't see a folder for Pandora there! Sorry, retract my comment.

brucehoward-physics:
Ok thanks for the clarification!

brucehoward-physics:
One of the suggestions on the LArPandora PR was that I should break out the definition and implementation into a header and cc file -- is this the preferred style on the icaruscode side too?

Also - what is the preferred behavior on the icaruscode side with respect to things like rebasing? I have a version of this branch on my local area with a few updates that is rebased against a newer release as well. Is it preferred that I force push this update to this remote, or should I follow some other guidelines? Thanks!

brucehoward-physics:
Also should comment that the newer PR version is here: https://github.com/PandoraPFA/larpandora/pull/29

though the one for LArSoft will come from Pandora.

mmrosenberg:
trigger build

cerati (approved):
Looks good to me. Eventually it may be useful to have the info about the duplicated hits in the event, but this can come as a future development

SFBayLaser:
With respect to breaking into header and cc file... Generally, art producer modules, services and tools are "top of the food chain" meaning that you are making a self contained module that satisfies a particular interface. If I understand correctly, "IHitCollectionTool" is the interface so the way you have written the tool is what I understand to be correct.

icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
```diff
@@ -1,7 +1,44 @@
...
+	art_root_io::TFileService_service
+	art::Framework_Principal
+	art::Framework_Services_Registry
+	messagefacility::MF_MessageLogger
+	cetlib::cetlib
+	cetlib_except::cetlib_except
+	PandoraPFA::PandoraSDK
+)
+
+cet_build_plugin(LArPandoraHitCollectionToolICARUS art::tool LIBRARIES ${TOOL_LIBRARIES})
```
SFBayLaser, line 38: It used to be the case that the most straightforward thing was to simply use "art_make" which had the ability to figure out what you were doing. I think because of the move to supporting the new build system this was replaced with the new cet build tools and I think what you have done is correct. I suppose you could go through the list of libraries you have included to see which you actually really need vs just there because of cut and paste...

SFBayLaser (approved):
This should work fine

icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+    else {
+      mf::LogDebug("LArPandoraHitCollectionToolICARUS") << "  Found: " << theHits->size() << " Hits " << std::endl;
+    }
+
+    for (unsigned int i = 0; i < theHits->size(); ++i) {
+      const art::Ptr<recob::Hit> hit(theHits, i);
+      if ( channelTimeHits.find(std::make_pair(hit->Channel(), int(hit->PeakTime()))) == channelTimeHits.end() ) channelTimeHits[std::make_pair(hit->Channel(), int(hit->PeakTime()))] = 1;
+      else continue;
```
PetrilloAtWork, line 47: A couple of tricks to avoid double map lookup (`find()` then `operator[]`), both relying on the fact that map `operator[]` creates an element if it does not exist, and it assigns the "default" value (`0` for integers):
* one liner:
    ```suggestion
          if ( std::exchange(channelTimeHits[std::make_pair(hit->Channel(), int(hit->PeakTime()))], 1) ) continue;
    ```
    This is a decently common pattern in "atomic" operations (operations which must not be interrupted e.g. by other threads), but it is not particularly readable.
* if that's to cryptic, alternative:
    ```suggestion
          int& count = channelTimeHits[std::make_pair(hit->Channel(), int(hit->PeakTime()))];
          if (count) continue;
          else count = 1;
    ```

icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
```diff
@@ -1,7 +1,44 @@
...
+	lardata::DetectorPropertiesService
+	larcore::Geometry_Geometry_service
+	larcorealg::Geometry
+	art_root_io::TFileService_service
+	art::Framework_Principal
+	art::Framework_Services_Registry
+	messagefacility::MF_MessageLogger
+	cetlib::cetlib
+	cetlib_except::cetlib_except
+	PandoraPFA::PandoraSDK
```
PetrilloAtWork, line 35: I think these are the ones needed for this tool:
```suggestion
	larpandora::LArPandoraInterface
	lardataobj::RecoBase
	canvas::canvas
	art::Framework_Principal
	messagefacility::MF_MessageLogger
	cetlib::cetlib
```

PetrilloAtWork (commented):
This code selects one hit per channel per microsecond — I suppose this is both intended and reasonable, considering that two hits in the same microsecond are quite unlikely.
An explicit documentation comment about this choice would be good...

For the rest, the usual collection of comments from me, but everything looks ok (I still encourage you to reduce the list of libraries to undermine the random proliferation — somebody, some day, might copy this list).

brucehoward-physics:
> 

Since I'm using PeakTime it should be one per _tick_ right? I had put together a more complicated algorithm at one point in case having multiple per tick would be a concern, but I think it's not?

icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  LArPandoraHitCollectionToolICARUS::LArPandoraHitCollectionToolICARUS(const fhicl::ParameterSet& pset){}
+
+  void LArPandoraHitCollectionToolICARUS::CollectHits(const art::Event& evt, const std::string& label, HitVector& hitVector)
+  {
+    std::map< std::pair<raw::ChannelID_t, int>, unsigned int > channelTimeHits; // converting the float PeakTime to int for the second item in pair
+
+    art::Handle<std::vector<recob::Hit>> theHits;
+    evt.getByLabel(label, theHits);
```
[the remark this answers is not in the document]
brucehoward-physics: Copied from the algorithm I was making into a tool:
https://github.com/PandoraPFA/larpandora/blob/develop/larpandora/LArPandoraInterface/LArPandoraHelper.cxx#L75

What is the preferred alternative?

icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
```diff
@@ -0,0 +1,56 @@
...
+
+    if (!theHits.isValid()) {
+      mf::LogDebug("LArPandoraHitCollectionToolICARUS") << "  Failed to find hits... " << std::endl;
+      return;
+    }
+    else {
+      mf::LogDebug("LArPandoraHitCollectionToolICARUS") << "  Found: " << theHits->size() << " Hits " << std::endl;
+    }
```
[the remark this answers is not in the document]
brucehoward-physics: Similarly https://github.com/PandoraPFA/larpandora/blob/develop/larpandora/LArPandoraInterface/LArPandoraHelper.cxx#L77

icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
brucehoward-physics: Thanks for the suggestions -- I will try to add one of these and test the behavior on a test event.

icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
brucehoward-physics: Cool, I'll try to build with this reduced set! 🤞

icaruscode/TPC/Tracking/ICARUSPandora/LArPandoraHitCollectionToolICARUS_tool.cc
brucehoward-physics: Addressed.

icaruscode/TPC/Tracking/ICARUSPandora/CMakeLists.txt
brucehoward-physics: Addressed. Thanks for the suggestion!

PetrilloAtWork:
> Since I'm using PeakTime it should be one per _tick_ right? I had put together a more complicated algorithm at one point in case having multiple per tick would be a concern, but I think it's not?

My bad, I was confused with the optical hits.

PetrilloAtWork (approved):
Full score :wink:

rennney:
trigger build

PetrilloAtWork:
Answer beyond this PR: 

> 2. I realized that the tool uses a few things which aren't expressly `#include`d like `fhicl::ParameterSet` or `lar_pandora::HitVector`, for examples. The latter should be included via `LArPandoraHitCollectionTool.h` bringing in `LArPandoraHelper.h`. Not sure off-hand where `ParameterSet` is brought in from the way it is now. Is it preferred that I explicitly add includes for `LArPandoraHelper.h` and `ParameterSet.h`?

Yes.
The general guideline is to `#include` all the headers declaring the objects _explicitly_ used, and in the link list of `CMaleLists.txt` all the libraries containing definitions _directly_ used.

An exception is when the classes belong to an interface that is inherited. For example, if you have a _art_ producer module, you will have `void produce(art::Event& event)`. Since this is a override of a member function of `art::EDProducer` (or similar), this must already be known and there is no need to declare it; _but_ if you are actually using the class (as you most likely are), then you'd better include its header anyway, because the header of `EDProducer` might have only _declared_ that it is a class (`namespace art { class Event; }`) without saying anything of what it contains, and that would still work as long as the compiler sees only pointers and references to `art::Event` without actual access to the pointed/referenced object.

PR #550, opened 2023-03-14
Processing for trigger efficiency and related changes to trigger efficiency analysis

jzettle:
This PR encompasses the updates discussed in the trigger meeting on the trigger efficiency. Most of the changes are restricted to a separate directory used for the trigger efficiency analysis directly but the main change relevant for the collaboration as a whole is a stage1 fcl file that runs the CRTT0Matching module separately for both the west and east cryostats and creates a separate data product based on those. At the moment this is a standalone fcl file that can be run separately. 

One thing I recall is that some of these changes are tied to a larreco change, https://github.com/LArSoft/larreco/pull/54 (@PetrilloAtWork am I correct?) and https://github.com/SBNSoftware/icaruscode/pull/543 so we need to know whether that is compatible I have it working in my own area thanks to Gianluca but I assume this does depend on that as well. It could be adventurous for this to converge in a week+ but it would help assuming if not we are locked out of doing this via the production group for the foreseeable future. 

Please add any other reviewers that are useful, I at least want to start the discussion. I have not done extensive stage1 processing tests using this branch myself but I do have it compiling also including the above dependent PRs.

### DIFF
--- a/fcl/reco/Definitions/stage1_icarus_defs.fcl
+++ b/fcl/reco/Definitions/stage1_icarus_defs.fcl
@@ -61,6 +61,8 @@ icarus_stage1_producers:
  ## crt producer
   crttrack:          @local::standard_crttrackproducer
   CRTT0Matching:     @local::standard_crtt0producer
+  CRTT0MatchingW:    @local::standard_crtt0producerW
+  CRTT0MatchingE:    @local::standard_crtt0producerE
 }
 
 icarus_stage1_filters:
@@ -165,6 +167,8 @@ icarus_reco_fm:                    [ fmatchCryoE,
 icarus_crttrack:                   [crttrack]
 
 icarus_crtt0match:                 [CRTT0Matching]
+
+icarus_crtt0match_eff:		   [CRTT0MatchingE, CRTT0MatchingW]
 
 ### Below we include overrides for the modules above
 
--- a/fcl/reco/Definitions/stage1_icarus_defs.fcl
+++ b/fcl/reco/Definitions/stage1_icarus_defs.fcl
@@ -168,7 +168,7 @@ icarus_crttrack:                   [crttrack]
 
 icarus_crtt0match:                 [CRTT0Matching]
 
-icarus_crtt0match_eff:		   [CRTT0MatchingE, CRTT0MatchingW]
+icarus_crtt0match_eff:             [CRTT0MatchingE, CRTT0MatchingW]
 
 ### Below we include overrides for the modules above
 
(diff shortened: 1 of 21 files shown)


jzettle:
Adding Tyler as well in case he wants to comment on anything CRT-related in here (mainly just fcl changes to enable what I said at the beginning)

PetrilloAtWork:
> One thing I recall is that some of these changes are tied to a `larreco` change, [LArSoft/larreco#54](https://github.com/LArSoft/larreco/pull/54) (@PetrilloAtWork am I correct?) and #543 so we need to know whether that is compatible I have it working in my own area thanks to Gianluca but I assume this does depend on that as well.

The changes in that pull request are self-contained and should not affect the physics results.
They did speed up the processing time of the calorimetry from 6" to 0.5" in our workflow, but that's all the difference they should make.
More explicitly, those changes don't require any change in existing user code nor configuration.

jzettle:
Okay good, I remembered I had issues compiling initially due to those dependencies (at least the larreco one) but I may have missed if subsequent updates here made it not directly depend on that other PR (or if that is not a concern anymore).

PetrilloAtWork:
No, you just missed the fact that I am messy.

mmrosenberg:
trigger build LArSoft/larreco#54

SFBayLaser:
Likely I am not following the comments well... does this PR rely on an as yet included PR to larreco so is not ready to go right now?

jzettle:
@SFBayLaser I think Gianluca has removed the necessity for the PR to larreco on this but @PetrilloAtWork should comment to if that is the correct way to interpret the most recent updates on this.

PetrilloAtWork:
That is correct (the C.I. test should be able to confirm).

rennney:
trigger build

PetrilloAtWork:
Please repeat the compilation test... I think I fixed the issue with Clang (not sure if I moved it back to GCC though).

mmrosenberg:
trigger build

PetrilloAtWork (approved):
"Oh, yeah, this PR is great, exactly what we need. I commend all the contributors!", said one of the reviewers, who also happens to be a contributor.
Anyway, it should not break anything, which is a low bar, but still an appreciable one.

mmrosenberg:
trigger build

SFBayLaser:
Question for @jzettle and/or @PetrilloAtWork... do we want this in the next release (work starting now)? Or still waiting on other things?

jzettle:
@SFBayLaser I would like this in the next release so it can help support the efficiency work. Was there something else we needed to do here?

jzettle:
It should not affect anything in the standard processing that I am aware the changes are either adding a special fcl or code changes in a directory only used for this analysis

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build LArSoft/lar*@v09_71_00

PR #551, opened 2023-03-20
Include the channel status service plus more fixes for fhicl files

SFBayLaser:
Introduce an ICARUS specific version of the Channel Status Service so we can form a "correct" timestamp to give to the standard code. This now correctly accesses the channel status database so now make using this in the 1D deconvolution (both for data and MC) the default. 
In addition, continue updating the data/MC fhicl files to take into account last minute updates/fixes.


SFBayLaser:
Just to note this somewhere... The timestamp for an event is recovered from the art Event object which has two 32 bit fields. The upper 32 bits ("timeHigh()") is the number of seconds for the event referenced to the epoch, the lower 32 bits ("timeLow()") are the number of ns for the event referenced to the second of the event. To form a "proper" time stamp that the standard LArSoft code can use to access the database one forms "timeHigh() * 1000000000 + timeLow()" (but be aware that these need to be UInt64_t objects).

SFBayLaser:
This needs icarus_data v09_69_01 to work

SFBayLaser:
trigger build

mmrosenberg:
trigger build

cerati (approved):
Looks good to me. We'll need to test the effect on production workflows after a new release is made

SFBayLaser:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
trigger build

PR #552, opened 2023-03-20
Feature/cerati fix pnfs paths to cvmfs

cerati:
This PR updates the paths in production files that point to /pnfs (not visible from grid nodes) and updates them to the area under /cvmfs (visible)


mmrosenberg:
trigger build

PR #554, opened 2023-03-25
PMT digitization does not truncate large signals any more.

PetrilloAtWork:
@Temigo  (and possibly another person who right now I don't remember, for which I apologise) has reported an issue which was tracked down to the following.
The PMT digitisation creates PMT waveforms out of simulated photoelectrons. It utilises a form of zero suppression driven by a discrimination algorithm (which has not much to do with the one used in the CAEN V1730B readout boards). Whenever the algorithm determines a "trigger primitive",  a buffer is filled around it and stored.
The discrimination algorithm was written to issue a trigger primitive every time the signal moves from below a threshold to above a threshold. Unfortunately this algorithm was paired with a 4 µs buffer length. In this situation, a strong signal which stays above threshold longer than 4 µs will have its tail cut, since to save it another trigger primitive is needed, that may be issued only when the signal has first gone under threshold and then back above it.
While the algorithm results are correct (no bug), this choice causes faulty results.

A solution to this problem may be to extend the readout window; for example, extending it to 10 µs like in the actual detector readout would have very little of this problem, and at a level similar to data anyway. This also means that for each noise sample, single photoelectron or fluke we store 10 µs of waveform, which is wasteful.
The solution in this pull request is a different algorithm that preserves all samples above the discrimination threshold, equivalent to produce a trigger primitive on every such sample. There is a small increase in data size (~1.5% in a sample of 10 cosmic ray events), which is in part due to the information being recovered, and in part due to an excess storage of the tail.
Note that **this pull request adopts the new algorithm as default** (it is thus a breaking change).

As a final note, this module in my opinion should abandon any attempt to reproduce the hardware readout (a downstream module, [`icarus::opdet::SimPMTreadout`](https://icarus-exp.fnal.gov/at_work/software/doc/icaruscode/latest/classicarus_1_1opdet_1_1SimPMTreadout.html), has been prototyped for that), and because of that I think moving to an algorithm farther from the readout is acceptable. Likewise, I think that as soon as the readout simulation module is validated, the readout window of _this_ stage can be shortened to just accommodate the single photoelectron duration, saving some memory, as the other module can pad waveforms with electronic noise as needed.

**Note**: this modification hasn't been approved nor discussed by the optical detector simulation and reconstruction group. A convener is added in the reviewer list to make sure the approval follows the right path.

Suggested reviewers:
* @Temigo (reporter)
* @amenegol (optical detector simulation and reconstruction coordinator)
* @wesketchum (involved, AFAIK, in the design and implementation of the first version of the module)
* @mvicenzi (as a punishment for being useful)

This pull request is for `develop`. There is no matching request for the production branch (@SFBayLaser may chime in on that).

### DIFF
--- a/icaruscode/PMT/Algorithms/PMTsimulationAlg.cxx
+++ b/icaruscode/PMT/Algorithms/PMTsimulationAlg.cxx
@@ -57,6 +57,15 @@ namespace util {
 // -----------------------------------------------------------------------------
 // ---  icarus::opdet::PMTsimulationAlg
 // -----------------------------------------------------------------------------
+util::MultipleChoiceSelection<icarus::opdet::PMTsimulationAlg::DiscriminationAlgo>
+const icarus::opdet::PMTsimulationAlg::DiscrimAlgoSelector {
+  //  { DiscriminationAlgo::Unset,              "Unset" }  // NOT an acceptable value!
+    { DiscriminationAlgo::AboveThreshold,     "AboveThreshold"    }
+  , { DiscriminationAlgo::CrossingThreshold,  "CrossingThreshold" }
+};
+
+
+// -----------------------------------------------------------------------------
 double
 icarus::opdet::PMTsimulationAlg::ConfigurationParameters_t::PMTspecs_t::
   multiplicationStageGain(unsigned int i /* = 1 */) const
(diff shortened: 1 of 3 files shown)


mmrosenberg:
trigger build

mmrosenberg:
trigger build

mvicenzi (approved):

mmrosenberg:
trigger build

wesketchum (approved):
This sounds and looks good to me.

PR #557, opened 2023-04-03
Adding trigger emulation to CAF [SBN2023A]

PetrilloAtWork:
This pull request aims to have a simulated trigger response in the CAF data for production.

The trigger simulation is already ran in the standard simulation workflow, with the input tag `emuTrigger`.
The changes operate at two levels:
1. Enable the production of `sbn::ExtraTriggerInfo` by the trigger simulation module; this is a feature that was recently introduced but had not been enabled for this workflow.
2. Direct CAF making module to the trigger response data product.

## Enable the production of `sbn::ExtraTriggerInfo`

The module responsible of the final step of the simulation, including the emission of `raw::Trigger` and `sbn::ExtraTriggerInfo` data products, is [`TriggerSimulationOnGates`](https://icarus-exp.fnal.gov/at_work/software/doc/icaruscode/latest/classicarus_1_1trigger_1_1TriggerSimulationOnGates.html).
As documented in that module, the latter data product is fairly incomplete, as many of the aspects of the trigger system are at a level lower than the one we simulate.
Also, this module always produces a `raw::Trigger` data product and a `sbn::ExtraTriggerInfo` one; the documentation also describes the values assigned when no trigger is fired.

## Direct CAF making module to the trigger response data product

The base configuration for CAF making is written for simulation. However, it is using as trigger data product tag the standard _data_ trigger `daqTrigger`. In simulation this ended up having no effect, since such data product deliberately does not exist — trigger simulation module instance is called `emuTrigger`.
Several modules refer to `daqTrigger` as trigger data product for whatever they need it for. The approach of this pull request is to just add the response of the trigger simulation, without affecting the reconstruction.
Therefore, the CAF making module in the simulation workflow now uses `emuTrigger` instead of `daqTrigger`, but all the other modules still look at `daqTrigger`, most noticeably `DetectorClocksService`.
At the same time, the trigger tag in the data configuration is now explicitly set (back) to `daqTrigger`.
The topic whether it's feasible and desirable to use `daqTrigger` for all the modules is tracked in issue #556.

## Other considerations

[...]

PR #558, opened 2023-04-03
Adding trigger emulation to CAF

PetrilloAtWork:
This is the twin PR of #557.
Discussion goes there.
As for review, I think as soon as that is approved, this should as well.
Anyway, I set as reviewers @SFBayLaser and @rennney in case they know about differences between production and development branch that warrant a specific change for this PR.


mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #559, opened 2023-04-06
Various fhicl file fixes to production release

SFBayLaser:
A few fhicl file fixes:
1) Remove the leading/trailing bins from the ROI finding to address apparent issues with pandora track/shower id
2) Add explicit fhicl files running the CRT PMT filter at stage 0 for both run 1 and run 2
3) Raise the message level in the stage 1 fhicl files from "DEBUG" to "INFO"
4) Kill the warning level messages at the g4 stage from SimDriftElectrons and McReco when it can't find a wire corresponding to a position, not sure if this is a real problem that we need to track down but so far it seems more an annoyance than a bug
5) Fix the intime generation file to insure the MCParticleLite collection is stored


SFBayLaser:
trigger build

mmrosenberg:
trigger build

SFBayLaser:
trigger build

PR #560, opened 2023-04-11
Remotes/origin/feature/fpoppi crt filter

francescopoppi:


francescopoppi:
This pull request is meant to address two items:
1) The CRTPMT filtering on stage0 is improved and tested both in loose and in medium mode. Default mode is loose and does not filter anything, although, in the output you can find the what could have happened for each event. In particular, for each event it is saved the list of optical flashes during the beam gate, if they were matched with CRT hits (and store the corresponding CRT information) and in general if the event would have been filtered out or not. The current filter is a logic AND of all the optical flashes classification in the gate window, e.g. if in the event there are two optical flashes in the 2.2 us window for BNB and the first one is matched with an entering Top CRT Hit, while the second is not, the filtering logic is applied to the and of both the flashes.
2) The CRTPMT analysis module is added to the definitions of the Stage1 for Run 1 and a Stage1 fcl with CRTPMT analysis module is included. Using this new fcl included the CRTPMT analysis tree in the calibration ntuple. The overall increase in size is 5 MB for 50 events (currently a Supplemental file for Run1 is 180MB-ish).

This PR is intended for the foreseen reprocessing of Run1, in particular the point2 of this list. There is an open issue, which I will address very soon, to create a CRTPMT matched data product. I don't think it will be ready by the time of the reprocessing, but it will be concluded by the beginning of summer.

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+          int CRTSys = 0;
+          if (CRTRegion >= 36)
+            CRTSys =
+                1;  // Very lazy way to determine if the Hit is a Top or a Side.
+                    // Will update it when bottom CRT will be availble.
+          double CRTTof_opflash = CRTtime - tflash;
+          std::vector<int> HitFebs;
+          for (auto& crts : entering.CRTHit->feb_id) {
+            HitFebs.emplace_back((int)crts);
+          }
```
PetrilloAtWork, line 485: Prefer copy for small data types:
```suggestion
          for (auto crts : entering.CRTHit->feb_id) {
            HitFebs.emplace_back((int)crts);
          }
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+                                     exiting.CRTHit->z_pos};
+          double CRTtime = exiting.CRTHit->ts1_ns / 1e3;
+          int CRTRegion = exiting.CRTHit->plane;
+          int CRTSys = 0;
+          if (CRTRegion >= 36) CRTSys = 1;
+          double CRTTof_opflash = CRTtime - tflash;
+          std::vector<int> HitsFebs;
+          for (auto& crts : exiting.CRTHit->feb_id) {
+            HitsFebs.emplace_back((int)crts);
+          }
```
PetrilloAtWork, line 510: As above:
```suggestion
          for (auto crts : exiting.CRTHit->feb_id) {
            HitsFebs.emplace_back((int)crts);
          }
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+          if (CRTSys == 0) TopEn++;
+          if (CRTSys == 1) SideEn++;
+          MatchedCRT thisCRTMatch = {.CRTHitPos = thisCRTpos,
+                                     .CRTPMTTimeDiff_ns = CRTTof_opflash,
+                                     .CRTTime_us = CRTtime,
+                                     .CRTSys = CRTSys,
+                                     .CRTRegion = CRTRegion};
+          thisFlashCRTmatches.push_back(thisCRTMatch);
+          HitFebs.clear();
```
PetrilloAtWork, line 494: This is superfluous since `HitFebs` is a local variable that is going out of scope on the next line:
```suggestion
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+          }
+          if (CRTSys == 0) TopEx++;
+          if (CRTSys == 1) SideEx++;
+          MatchedCRT thisCRTMatch = {.CRTHitPos = thisCRTpos,
+                                     .CRTPMTTimeDiff_ns = CRTTof_opflash,
+                                     .CRTTime_us = CRTtime,
+                                     .CRTSys = CRTSys,
+                                     .CRTRegion = CRTRegion};
+          thisFlashCRTmatches.push_back(thisCRTMatch);
+          HitsFebs.clear();
```
PetrilloAtWork, line 519: As above:
```suggestion
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+        }
+      }
+      FlashType thisFlashType = {.FlashPos = flash_pos,
+                                 .FlashTime_us = tflash,
+                                 .FlashGateTime_ns = fThisRelGateTime,
+                                 .inBeam = fThisInTime_beam,
+                                 .inGate = inTime,
+                                 .Classification = eventType,
+                                 .CRTmatches = thisFlashCRTmatches};
```
PetrilloAtWork, line 528: As above:
```suggestion
      FlashType thisFlashType = { /* .FlashPos = */ flash_pos, // C++20: restore initializers
                                 /* .FlashTime_us = */ tflash,
                                 /* .FlashGateTime_ns = */ fThisRelGateTime,
                                 /* .inBeam = */ fThisInTime_beam,
                                 /* .inGate = */ inTime,
                                 /* .Classification = */ eventType,
                                 /* .CRTmatches = */ thisFlashCRTmatches};
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
-    vector<int>                fMatchedCRTregion;       ///< Region of the matched CRT Hits.
-    vector<vector<int>>        fMatchedCRTmodID;        ///< ModID of the CRT module.
-    vector<int>                fMatchedCRTsys;          ///< Subsystem of the matched CRT Hit.
-    vector<double>             fMatchedCRTamplitude;    ///< Amplitude of the matched CRT Hit.
-    vector<double>             fTofOpFlash;             ///< Time of Flight between matched CRT and Optical Flash.
-    vector<double>             fCRTGateDiff;            ///< Difference between CRTHit and BeamGate opening.
-    int                        fEventType;              ///< Was triggered the event?
-
-    geo::GeometryCore const*   fGeometryService; ///< pointer to Geometry provider.
+class icarus::crt::FilterCRTPMTMatching : public art::EDFilter {
```
PetrilloAtWork, line 189: It would be good to have a long and exhaustive description of what the module does and how to configure it, in particular for the different levels. A reference to a SBN DocDB document is also welcome.
```suggestion
/**
 * @brief Rejects events with only incoming particles in time with the beam.
 *
 * Here is a long and detailed explanation on what are the choices.
 */
class icarus::crt::FilterCRTPMTMatching : public art::EDFilter {
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+      // CRT Hit before the Flash is filtered as clear cosmic
+      else
+        ThisFlashType = false;
+      HitsInGate = HitsInGate && ThisFlashType;
     }
-    else if (fFilterLevel == "thight"){
-	if (fEventType != 0 || fEventType != 4 || fEventType != 5) Cosmic = true;
-	// With Thight filter, everything (inTime) associated with a CRT Hit before the Flash is filtered as clear cosmic
-	else Cosmic = false;
+  } else if (fFilterLevel == "thight") {
```
PetrilloAtWork, line 550: (Critical) typo:
```suggestion
  } else if (fFilterLevel == "tight") {
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
-	else Cosmic = false;
+  }
+  bool HitsInGate = false;
+  if (fFilterLevel == "loose")
+    HitsInGate = false;  // By default, with loose Filtering, everything is
+                         // "intersting", nothing tagged as clear cosmic
+  else if (fFilterLevel == "medium") {
+    HitsInGate = true;
+    for (const auto& h : thisEventFlashes) {
+      bool ThisFlashType = false;
```
PetrilloAtWork, line 539: More appropriate variable name? For example:
```suggestion
      bool isCosmic = false;
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
-    else if (fFilterLevel == "thight"){
-	if (fEventType != 0 || fEventType != 4 || fEventType != 5) Cosmic = true;
-	// With Thight filter, everything (inTime) associated with a CRT Hit before the Flash is filtered as clear cosmic
-	else Cosmic = false;
+  } else if (fFilterLevel == "thight") {
+    // With Thight filter, everything (inTime) associated with a CRT Hit before
+    // the Flash is filtered as clear cosmic
+    HitsInGate = true;
+    for (const auto& h : thisEventFlashes) {
+      bool ThisFlashType = false;
```
PetrilloAtWork, line 555: As above:
```suggestion
      bool isCosmic = false;
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
     }
-    fEventType   = Type;
-
-    if (fFilterLevel == "loose") Cosmic = false;  // By default, with loose Filtering, everything is "intersting", nothing tagged as clear cosmic
-    else if (fFilterLevel == "medium") {
-	if (fEventType == 1 || fEventType == 3 || fEventType == 6 || fEventType == 7) Cosmic = true;
-	// With Medium filter, everything (inTime) which is associated with Top CRT Hit before the Flash is filtered as clear cosmic
-	else Cosmic = false;
+  }
+  bool HitsInGate = false;
```
PetrilloAtWork, line 532: The name of this variable to me fails to capture its meaning.
```suggestion
  bool hasOnlyCosmics = false;
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+      else
+        ThisFlashType = false;
+      HitsInGate = HitsInGate && ThisFlashType;
     }
-    else if (fFilterLevel == "thight"){
-	if (fEventType != 0 || fEventType != 4 || fEventType != 5) Cosmic = true;
-	// With Thight filter, everything (inTime) associated with a CRT Hit before the Flash is filtered as clear cosmic
-	else Cosmic = false;
+  } else if (fFilterLevel == "thight") {
+    // With Thight filter, everything (inTime) associated with a CRT Hit before
```
PetrilloAtWork, line 551:
```suggestion
    // With Tight filter, everything (inTime) associated with a CRT Hit before
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
-    fMatchedCRTtime.clear();
-    fMatchedCRTregion.clear();
-    fMatchedCRTsys.clear();
-    fMatchedCRTamplitude.clear();
-    fTofOpFlash.clear();
-    fCRTGateDiff.clear();
+  }
+  fFiltered = HitsInGate;
+  EventCRTPMT thisEvent = {.Filter = HitsInGate,
+                           .inGateFlashes = thisEventFlashes};
```
PetrilloAtWork, line 566:
```suggestion
  EventCRTPMT thisEvent = {/* .Filter = */ HitsInGate, // C++20: restore initializers
                           /* .inGateFlashes = */ thisEventFlashes};
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+      fCRTHitSystem.push_back(crt.CRTSys);
+    }
+    fMatchTree->Fill();
+    fCRTHitPos_X.clear();
+    fCRTHitPos_Y.clear();
+    fCRTHitPos_Z.clear();
+    fCRTHitTime_us.clear();
+    fCRTFlashTime_ns.clear();
+    fCRTHitRegion.clear();
+    fCRTHitSystem.clear();
```
PetrilloAtWork, line 592: Isn't this:
```suggestion
    ClearVecs();
```
?

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -13,556 +13,607 @@
...
+    e.getByLabel(fTriggerLabel, trigger_handle);
+    if (trigger_handle.isValid()) {
+      sbn::triggerSource bit = trigger_handle->sourceType;
+      m_gate_type = (unsigned int)bit;
+      m_gate_name = bitName(bit);
+      m_trigger_timestamp = trigger_handle->triggerTimestamp;
+      m_gate_start_timestamp = trigger_handle->beamGateTimestamp;
+      m_trigger_gate_diff =
+          trigger_handle->triggerTimestamp - trigger_handle->beamGateTimestamp;
+      m_gate_width = fTriggerConfiguration.getGateWidth(m_gate_type);
```
PetrilloAtWork, line 376: Unless I have missed it, `fTriggerConfiguration` is not initialised anywhere (`getTriggerConf()` is not called).
The best place to call it is in the `beginRun()` method:
```cpp
void icarus::crt::FilterCRTPMTMatching::beginRun(art::Run& run) {
  getTriggerConf(run);
}
```
which need to be also declared in the class:
```cpp
    void beginRun(art::Run& run) override;
```

PetrilloAtWork (changes requested):
My usual rain of comments...
Things that need to be taken care of before any production:
 * the C++20 syntax for initializing classes, which I bet Clang 7 does not support
 * the configuration of the filter level (user typos need to be reported)
 * the debugging output level in production configurations
 * the unit of `tof`
 * move default configuration values into explicit configuration

The rest are more or less important suggestions.

francescopoppi:
Ok I think these commit addressed all the request from @PetrilloAtWork. If the reviewer accepts, it is ready to go.
I tested the modules both with Run1 and Run2 data files.

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -187,13 +185,33 @@ icarus::crt::CRTMatches CRTHitmatched(
...
+
  public:
   using CRTHit = sbn::crt::CRTHit;
 
   explicit FilterCRTPMTMatching(fhicl::ParameterSet const& p);
-
   // Required functions.
-  void getTriggerConf(art::Run&);
+  //void getTriggerConf(art::Run const&);
+  void beginRun(art::Run const& run);
```
PetrilloAtWork, line 214: Mark `beginRun()` `override`, and I think you need to remove `const` (with `override`, GCC will tell you if that's the case: `"override" overrides nothing` or something like that).
```suggestion
  void beginRun(art::Run& run) override;
```

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -347,56 +351,42 @@ icarus::crt::FilterCRTPMTMatching::FilterCRTPMTMatching(
   }
 }
 
-void icarus::crt::FilterCRTPMTMatching::getTriggerConf(art::Run& r) {
+//void icarus::crt::FilterCRTPMTMatching::getTriggerConf(art::Run const& r) {
+//  fTriggerConfiguration =
+//      r.getProduct<icarus::TriggerConfiguration>(fTriggerConfigurationLabel);
+//}
+
+void icarus::crt::FilterCRTPMTMatching::beginRun(art::Run const& r) {
```
PetrilloAtWork, line 359:
```suggestion
void icarus::crt::FilterCRTPMTMatching::beginRun(art::Run& r) {
```
In principle, in `beginRun()` of a producer or filter, you _could_ add data products to the run, and that's the reason the argument is not declared `const`. You need to replicate the exact signature in order for _art_ to properly call it.
In that sense, the advantage of `getTriggerConf(art::Run const& r)` was that you were not imposed any signature and you could declare `r` as `const`. In this respect, having `beginRun()` call `getTriggerConf()` has still a reason, but I am not endorsing either way over the other here.

icaruscode/Filters/FilterCRTPMTMatching_module.cc
```diff
@@ -464,106 +453,105 @@ bool icarus::crt::FilterCRTPMTMatching::filter(art::Event& e) {
...
           }
           if (CRTSys == 0) TopEn++;
           if (CRTSys == 1) SideEn++;
-          MatchedCRT thisCRTMatch = {.CRTHitPos = thisCRTpos,
-                                     .CRTPMTTimeDiff_ns = CRTTof_opflash,
-                                     .CRTTime_us = CRTtime,
-                                     .CRTSys = CRTSys,
-                                     .CRTRegion = CRTRegion};
+	  MatchedCRT thisCRTMatch = { /* .CRTHitPos = */ thisCRTpos, // C++20: restore initializers
+                                     /* .CRTPMTTimeDiff_ns = */ CRTTof_opflash,
```
PetrilloAtWork, line 478: I had missed the unit mismatch here. Good catch!

PetrilloAtWork (changes requested):
I believe there is a technical issue with the new `beginRun()`.
Once that is fixed, it's pre-approved.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PetrilloAtWork:
Three new commits
1. should fix the issue with MC by having the analyzer (force push to change the wrong commit description)
2. should fix (I hope) an issue with the production of the tree (kind of heisenbug, can't be really sure);
3. the standard Stage0 configuration runs the filter, a special one does not (swapped compared to before)

mmrosenberg:
trigger build

PR #561, opened 2023-04-11
CAFmaker w/o event weights for detsim2d

jacoblarkin:
New cafmaker fcl based on existing detsim2d cafmaker fcl, but without running systtools or fluxweight. The event weights fail to run on intime cosmic MC so this is needed to make those CAFs.

### DIFF
--- a/fcl/caf/cafmaker_add_detsim2d_icarus.fcl
+++ b/fcl/caf/cafmaker_add_detsim2d_icarus.fcl
@@ -1,5 +1,3 @@
-#include "cafmakerjob_icarus.fcl"
-
 # over-write labels to use Wire-Cell SimChannels
 physics.producers.cafmaker.SimChannelLabel: "daq:simpleSC" # 2D drift
 
--- a/fcl/caf/cafmakerjob_icarus_detsim2d.fcl
+++ b/fcl/caf/cafmakerjob_icarus_detsim2d.fcl
@@ -1,16 +1,2 @@
-#include "cafmakerjob_icarus_systtools_and_fluxwgt.fcl"
-
-# over-write labels to use Wire-Cell SimChannels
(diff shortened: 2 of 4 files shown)


cerati:
Thank you @jacoblarkin ! Assuming this has been tested it should go in release asap

cerati (approved):

jacoblarkin:
I've tested it by making intime cosmic CAFs. Looking at this now, I realize I made the PR for develop. Should I be using SBN2023A instead for the production release?

cerati:
Thanks. About the target branch I'll let @rennney and @mmrosenberg comment

PetrilloAtWork (changes requested):
The current configurations have:
 * `cafmakerjob_icarus.fcl` as base configuration (no weights)
 * `cafmakerjob_icarus_systtools_and_fluxwgt.fcl` as special configuration with weights

I would prefer to preserve the same pattern for the `detsim2d` variant:
 * `cafmakerjob_icarus_detsim2d.fcl`  with 2D processing (no weights)
 * `cafmakerjob_icarus_detsim2d_systtools_and_fluxwgt.fcl`  with 2D processing and weights

And yes, I am aware that this means that `cafmakerjob_icarus_detsim2d.fcl` changes meaning. I hope we are early enough in the process that we can change that with reduced pain.

Comments welcome.

PetrilloAtWork (approved):
It looks good to me.
Thank you for your patience, @jacoblarkin.

mmrosenberg:
trigger build

mmrosenberg:
In response to the previous question from @jacoblarkin: the target branch is correct. I will be merging all of these changes into the release branch myself when I merge develop into the release branch (we're doing it this way as we need to update the larsoft version used in the release branch)

PR #563, opened 2023-04-17
`prodtext_standard_icarus.fcl`: add mandatory parameter

PetrilloAtWork:
Commit LArSoft/larsim#85 added a mandatory parameter `Offset` to `TextFileGen` module, a breaking change which ICARUS codebase did not absorb.
This is the belated fix.

Thanks to Hector Carranza (whose GitHub user I can't find) for reporting this issue.

Suggested reviewers:
* @SFBayLaser, workflow overlord
* The Real Ndrej, because of retaliation

### DIFF
--- a/fcl/gen/text/prodtext_standard_icarus.fcl
+++ b/fcl/gen/text/prodtext_standard_icarus.fcl
@@ -95,6 +95,7 @@ physics: {
     generator: {
       module_type:    TextFileGen
       InputFileName: "HEPevents.txt"
+      Offset:         0
     }
     
     # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


SFBayLaser:
I did not read the entire thread in [LArSoft/larsim#85](https://github.com/LArSoft/larsim/pull/85) but it says the default value of the new parameter is 0... which is what it is explicitly being set to here as well? So its not breaking? Or am I missing something I would have seen if I read the entire thread?

mmrosenberg:
trigger build

PetrilloAtWork:
> it says the default value of the new parameter is 0

It does. It was changed by one later commit in the same PR. Afterthought?

PetrilloAtWork:
That is the reason I am not requesting the addition of a default value — I am assuming that there was a reason behind that utterly inconvenient choice.

PR #564, opened 2023-04-18
Feature/cerati fix decoder mc

cerati:
This PR restores the behavior of MCDecoderICARUSTPCwROI before my [PR536](https://github.com/SBNSoftware/icaruscode/pull/536). The problem was that the behavior of [these lines](https://github.com/SBNSoftware/icaruscode/blob/v09_68_00_01/icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc#L600-L603) was not captured in the previous version. This fix restores the behavior.

### DIFF
--- a/icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
+++ b/icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
@@ -495,6 +495,7 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&                     
     // Read in the digit List object(s).
     art::Handle< std::vector<raw::RawDigit> > digitVecHandle;
     event.getByLabel(inputLabel, digitVecHandle);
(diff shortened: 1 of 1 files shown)


SFBayLaser (approved):
Yes, this is a critical update! 
When we run the 1D detector simulation there is a loop over all channels known to the geometry service, in particular including the disconnected channels. This means when the MC decoder runs it will always see the 64 channels needed for the coherent noise subtraction. In contrast, when we run the 2D drift simulation it will only output the active channels (meaning those that might potentially have signal)... so in this case for the first induction plane there will be missing data if we don't check the special condition in the decoder. I should have remembered that in the initial PR introducing the multi threading update... my mistake!

mmrosenberg:
trigger build

mmrosenberg:
trigger build

brucehoward-physics (approved):

PR #565, opened 2023-04-19
Last of the fhicl file updates

SFBayLaser:
Updating the run 1 and run 2 reco fhicl files to better manage the messages that are output to log files as the jobs run. Two primary issues, first is the potential volume of messages which was due to "DEBUG" level settings (unfortunately, left over from testing), the other was that at stage 1 there was suppression of all messages except those explicitly asked for. Hopefully we achieve a compromise here where we can get WARNING level (and above) messages for some number of events then suppression after. As a bonus we have recovered the job summary at the end of stage 1 (which is handy when running interactively). 
There is also a formatting change for the MC decoder which is purely aesthetics.

### DIFF
--- a/fcl/reco/Stage1/Run1/stage1_multiTPC_icarus_gauss.fcl
+++ b/fcl/reco/Stage1/Run1/stage1_multiTPC_icarus_gauss.fcl
@@ -39,23 +39,13 @@ services.message.destinations :
   STDCOUT:
   {
      type:      "cout"      #tells the message service to output this destination to cout
-     threshold: "DEBUG"     #tells the message service that this destination applies to WARNING and higher level messages
+     threshold: "WARNING"     #tells the message service that this destination applies to WARNING and higher level messages
      categories:
      {
        Cluster3D:
        {
          limit: -1
          reportEvery: 1
-       }
-       PMAlgTracker:
-       {
-         limit: -1
-         reportEvery: 1
-       }
-       default:
-       {
-         limit: 0  #don't print anything at the infomsg level except the explicitly named categories
-         reportEvery: 0
        }
      }
   }
--- a/icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
+++ b/icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
@@ -501,56 +501,51 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&                     
     {
         const unsigned int dataSize = digitVecHandle->front().Samples(); //size of raw data vectors
 
-	    std::map<unsigned int, std::vector<const raw::RawDigit*> > boardToRawDigitMap;
-
-	    for(const raw::RawDigit& rawDigit: *digitVecHandle) 
-        {
-	        raw::ChannelID_t channel = rawDigit.Channel();
-	        ChannelToBoardWirePlaneMap::const_iterator channelToBoardItr = fChannelToBoardWirePlaneMap.find(channel);
-
-	        if (channelToBoardItr == fChannelToBoardWirePlaneMap.end())
+	std::map<unsigned int, std::vector<const raw::RawDigit*> > boardToRawDigitMap;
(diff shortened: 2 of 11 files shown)


gputnam:
@SFBayLaser why is the savewires configuration being deleted in this PR? That configuration is necessary for the NuMI MC request.

Is there a presentation that describes the code changes being made in this PR?

Thanks.

SFBayLaser:
I have restored the fhicl file in question. I will point out again that you can trivially recreate the recob::Wire data from recob::ChannelROI using the module, in icaruscode, ROIConverter_module.cc which can easily be used to satisfy yourself that the wire data is correctly reproduced modulo rounding to the nearest integer. It would be equally trivial to include this module in the fhicl file to make the output CAFs (or other ntuples) to not have to modify code at that level. This has been checked and discussed, it is designed to even further reduce the footprint of stage 1 output. While we might like to think that any given data object is small enough to not worry about, we do seem to continually have the problem of storage and need to be mindful of things that can be done to reduce the event data footprint... There have been studies showing that when we don't save the raw waveforms the recob::Wire data IS a significant contributor to the remaining output space (second only to the PMT waveforms which have also gone through some slimming). 

You can easily read the commit comments in this thread to see what is being changed in the fhicl files. The primary goals is to reduce the volume of various messages output to log files while also restoring some output that was lost before.

SFBayLaser:
trigger build

gputnam:
Thank you for reverting that Tracy. 

Regarding the PR updates. I see that there are also changes to the actual MC decoder, in addition to the fcl updates. I was wondering if you could say what those are for.

mmrosenberg:
trigger build

cerati (approved):
changes look good to me. I note that apart from formatting the changes to the decoder match what is in my pending "bug fix PR" #564

fcl/reco/Stage1/Run1/stage1_multiTPC_icarus_gauss.fcl
```diff
@@ -39,24 +39,14 @@ services.message.destinations :
...
-       PMAlgTracker:
-       {
-         limit: -1
-         reportEvery: 1
-       }
-       default:
-       {
-         limit: 0  #don't print anything at the infomsg level except the explicitly named categories
-         reportEvery: 0
-       }
```
PetrilloAtWork, line 59: What happens when there is no `default` section?

icaruscode/Decode/MCDecoderICARUSTPCwROI_module.cc
```diff
@@ -501,51 +501,59 @@ void MCDecoderICARUSTPCwROI::processSingleLabel(art::Event&
...
+	            // Decompress data into local holder
+	            raw::Uncompress(rawDigit->ADCs(), rawDataVec, rawDigit->Compression());
+	            // Fill into the data structure
+                raw::ChannelID_t channel = rawDigit->Channel();
+                unsigned int planeIdx       = fChannelToBoardWirePlaneMap.find(channel)->second.second.second;
+	            icarus_signal_processing::VectorFloat boardDataVec(rawDataVec.cbegin(),rawDataVec.cend());
+	            chanArr.first.push_back(daq::INoiseFilter::ChannelPlanePair(channel,planeIdx));
+	            chanArr.second.push_back(boardDataVec);
+	        }
+              processSingleImage(clockData, chanArr, coherentNoiseGrouping, concurrentRawDigits, concurrentRawRawDigits, coherentRawDigits, concurrentROIs);
```
cerati, line 552: I think this line (552) is spurious, as processSingleImage is already called correctly below

cerati (dismissed):

fcl/reco/Stage1/Run1/stage1_multiTPC_icarus_gauss.fcl
SFBayLaser: It is defined in the common fcl file...

SFBayLaser:
trigger build

PR #566, opened 2023-04-19
Update to 3D space point building to reduce overall volume of space point output

SFBayLaser:
In looking into an issue seen in the MC reconstruction and in leaving no stone unturned discovered that a fraction of space points built by the cluster 3D space point code could be considered redundant. This is because its possible to have space points made from only 2 hits, one each from 2 planes. This defines a point in space with the YZ coordinates defined by the "intersection" point of the two wires containing the hits. If you add a third hit then the YZ position is slightly altered but this small change is well within the voxel dimensions used by the ML people and doesn't matter in the pandora processing. A redundant point can be created when you try to use a 2 hit combination more than once... so include here some protection that prevents that from happening.

At this time making this PR so it can be tested... but not ready for merging to develop just yet...


SFBayLaser:
Ok, somehow this is based to https://github.com/SBNSoftware/icaruscode/pull/565 but I don't think this is a problem since that is minor fhicl file changes (and one code formatting change)

PR #567, opened 2023-04-19
Removal of PMT delay correction

PetrilloAtWork:
Two changes to the PMT decoding, retroactive for Run1 and Run2:
 * there is now consensus that the delay between PMT readout boards from the trigger primitive chain distribution is implicitly corrected by the algorithm to timestamp the waveforms, and that the explicit correction is not needed and it was in fact double-correcting (see e.g. [SBN DocDB 30911](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=30911); this request removes that correction from both Run1 and Run2 configurations;
 * the special signal on the first PMT readout board is the global trigger, not a trigger primitive; the configuration has changed so that the labels reflect that fact; this does not change the waveforms nor the values of the corrections that are extracted from there; it does change the name of the data product (`daqPMT:globtrg` instead of `daqPMT:trgprim`) and their channel numbers (increased by `0x6000`).

This is a breaking change because of the removal of the delay correction, which will subtly change PMT timings.
This is of little consequence except for people working on the PMT/CRT matching (@francescopoppi  et al.) and on the RWM tuning (@biswaranjan91, @mvicenzi).

As a complimentary act, old configurations from `icaruscode/Decode/fcl` that were not installed have been removed in this PR: the confusion they create is not worth the keeping.

Note: these changes should be also **adopted in production** for reprocessing Run1 and Run2.

### Recommended reviewers

* @mvicenzi as person informed on the corrections
* @SFBayLaser as workflow manager, since this involves changes in the production configuration


mmrosenberg:
trigger build

PR #568, opened 2023-04-25
Fix a bug in PMT digitization

PetrilloAtWork:
The algorithm creating readout buffers (future `raw::OpDetWaveform`) from the continuous readout was written expecting (and asserting) that a trigger primitive can't be at the first tick of the enable gate. This was a reasonable assumption when the trigger algorithm looked for a transition from the previous tick from below to above threshold. Now an alternative algorithm is also available which considers just the value of the sample compared to the threshold, so that expectation is violated and the assertion failed. A small change in the algorithm makes it robust to that rare occurrence.

Thanks to @gputnam for the perfect report (code version, input file, input configuration, error message). As a punishment, he gets to review the change.

### DIFF
--- a/icaruscode/PMT/Algorithms/PMTsimulationAlg.cxx
+++ b/icaruscode/PMT/Algorithms/PMTsimulationAlg.cxx
@@ -560,7 +560,7 @@ icarus::opdet::PMTsimulationAlg::CreateFixedSizeOpDetWaveforms
   std::vector<BufferRange_t> buffers;
   buffers.reserve(std::distance(iNextTrigger, tend)); // worst case
   
-  auto lastBufferEnd { firstTick };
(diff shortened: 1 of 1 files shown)


gputnam (approved):
I've tested this PR and can confirm that it works as intended.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_72_00 SBNSoftware/sbncode@v09_72_00 SBNSoftware/sbnanaobj@v09_20_06 SBNSoftware/sbndaq_artdaq_core@v1_06_00of0 SBNSoftware/sbnobj@v09_16_00 SBNSoftware/icarus_signal_processing@v09_63_00_02 SBNSoftware/icarusalg@v09_72_00 SBNSoftware/icarusutil@v09_66_02

PR #569, opened 2023-04-26
Fix a bug in PMT digitization [`develop`]

PetrilloAtWork:
This is the request twin of #568, but for `develop` instead of the production branch.
All discussion should go there.


mmrosenberg (approved):

mmrosenberg:
trigger build

PR #571, opened 2023-04-30
Tuned Simulated Field Responses

gputnam:
Extends the WireCell jsonnet configuration to make more parameters configurable. Adds in a configuration to simulate the "tuned" signal response.

I have validated that the nominal configuration produces the exact same JSON output as before this PR.

Notes:
-I am putting this into draft form for now since we still may be tuning some things like the gain
-This will also require an update to icarus_data to add in the new field response file


gputnam:
Made this with the wrong branch -- going to fix.

PR #576, opened 2023-05-04
Fix a bug in `SlidingWindowTriggerSimulation`

PetrilloAtWork:
The affected module is used to run trigger emulation and create `raw::Trigger` data products out of them, and to create plots at the same time. It's not the module used for the global trigger simulation.
It used to crash, with this fix it does not any more. The bug does not affect the simulation, just the plots.
That's pretty much it.
Not sure if it needs a review...


mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_73_00

mmrosenberg:

SFBayLaser (approved):
Someone should check off even if an "obvioius" bug fix...

PR #577, opened 2023-05-04
Removed flash filter from Stage1 standard configuration

PetrilloAtWork:
The job configuration `stage1_run2_icarus.fcl` featured a flash filtering which has been outfashioned by the presence of a trigger and will be obsoleted by filtering based on CRT/PMT matching.
The filter in that job is also misconfigured, using a very old time reference for its flash window specification which is now way out of the PMT readout range. That means that using that configuration will suppress all the input events.

This pull request:
* removes the filter from `Stage1` configurations (Run1 and Run2); effectively, `stage1_run2_icarus.fcl` is now equivalent to `stage1_run2_nofilter_icarus.fcl`;
* updates the configuration in the `Stage0` definition (which is not used anyway)
* updates the configuration of the stand-alone filter job

I took the chance to also update the list of configurations that should be skipped by the FHiCL syntax test.

I will follow with a pull request for the production branch.

Suggested reviewer:
* @SFBayLaser, workflow maintainer
* Ivan Caro, production manager, if I could have tagged him on GitHub


mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_73_00

mmrosenberg:

SFBayLaser (approved):
Is there a reason to not also delete the file "stage1_run2_nofilter_icarus.fcl" since it will be redundant? Or am I just not seeing that in the list?

Otherwise, thank you for addressing this as it was obvious this was quite out of date.

PetrilloAtWork:
No strong reason. I did not remove it because I thought it may have a role if the CRT/PMT matching-based filter is introduced, but then it will probably be at Stage0 anyway.

PetrilloAtWork:
Adopted the reviewer suggestion of removing the configurations names `nofilter`, which indeed do not apply any filter but they are not special in that regard.
The twin PR #578 is a bit more tricky though, since one of the `nofilter` configurations is used in the production chain.
Specific discussion goes in that PR.

mmrosenberg:

PetrilloAtWork:
The test is right, a fix is needed.

PetrilloAtWork:
Please run the test again: it should be fixed now.
At least it passes the test locally.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #582, opened 2023-05-16
Crtpmt filter with data product

francescopoppi:
This PR modifies the CRTPMTFiltering to use the DataProduct created by Anna instead of rerunning an analysis.


francescopoppi:
I forgot to mention it, in this PR I also included a definition for Stage0 that only runs CRT, PMT decoding and reconstruction and CRTPMT filtering, discarding all the TPC information

PR #586, opened 2023-05-18
Add flash ID to CRT/PMT matching and sorting them by flash time

PetrilloAtWork:
This pull request is meant to address the suggestions by @francescopoppi on the candidate version of the CRT/PMT matching producer:
 * flash ID should be added
 * matches should be sorted by flash time

Both require some choices, which I am explaining here.

## Flash ID

The `icarus::crt::CRTPMTMatching` object is ready to host as data member the ID of the flash that is being matched.
Unfortunately, `recob::OpFlash` is one of the data products which do not have an ID field.
On top of that, creating one is not trivial since the flashes live in ICARUS in two independent data products, one per cryostat. Given the absence of a native ID, the only reliable way to assign a flash an ID is to extract its value from the fields of the flash itself; otherwise, given a flash ID, there is no way to figure out to which of the original flashes it belongs to.
My choice here was to attempt an unique ID, which unfortunately makes it not easy to read (no such a thing like ID `1`, ID `2` etc.). The implementation is to use as an ID the time of the flash (`recob::OpFlash::Time()`) in _picoseconds_, rounded.
Technically this is _not guaranteed_ to be unique, but the chances of collision are quite negligible.

## Sorting the matches

The choice here was to sort the matches by the time (`recob::OpFlash::Time()`) of the flash being matched.
The tie breaker, for the unlikely case where two _different_ flashes have the same time, is on the largest number of photoelectrons, and finally on a technical detail (the order of their _art_ pointers).
The implementation mixes the matches between the flashes of the two cryostats. If this is not desired, the module should be run twice, each time on a single cryostat. Also, the order of the matches will not coincide with the order of the flashes any more (that rule had some exceptions already).

### DIFF
--- a/icaruscode/CRT/CRTPMTMatchingProducer_module.cc
+++ b/icaruscode/CRT/CRTPMTMatchingProducer_module.cc
@@ -503,22 +503,22 @@ namespace icarus::crt {
         
         double const tflash = flashPtr->Time();
         double const totPe  = flashPtr->TotalPE();
-	double const flashYWidth = flashPtr->YWidth();
-	double const flashZWidth = flashPtr->ZWidth();
+        double const flashYWidth = flashPtr->YWidth();
+        double const flashZWidth = flashPtr->ZWidth();
 
-	vector<recob::OpHit const*> const& hits = findManyHits.at(flashPtr.key());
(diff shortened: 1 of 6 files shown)


francescopoppi (approved):
Thank you for these changes!
I approve them

PR #589, opened 2023-05-22
Correction of a bug where a vector was not cleared.

francescopoppi:
There was a bug where a vector of time differences between flashes and CRT hits was not cleared.


mmrosenberg:
trigger build

PR #590, opened 2023-05-25
Initial import of a filter module to check data integrity

SFBayLaser:
The idea is to check the integrity of the incoming data before running the processing chain with the hope to prevent job failures due to data issues. The main issue we have at the moment are incomplete events which can cause the decoder modules to throw exceptions and the processing to stop, so this initial import really aims to address that. The caveat is that it uses code similar to that used by the daq/dqm folks and it uses the first event, which is assumed to be fine, to initialize... so if the incomplete event is the first event then this module will miss it. Work in progress... but better than nothing.

### DIFF
--- a/icaruscode/Analysis/trigger/magic_raw_to_triggeremu_icarus_Run1_data.fcl
+++ b/icaruscode/Analysis/trigger/magic_raw_to_triggeremu_icarus_Run1_data.fcl
@@ -145,7 +145,7 @@ physics.filters: {
 
 physics.pathE: [
     @sequence::physics.pathE
-  , @sequence::icarus_filter_cluster3DCryoE
+  , @sequence::icarus_reco_cluster3DCryoE
   , @sequence::icarus_reco_pandoraGausCryoE
   , fmatchCryoE
   , @sequence::icarus_crttrack
@@ -153,7 +153,7 @@ physics.pathE: [
   ]
 physics.pathW: [
     @sequence::physics.pathW
-  , @sequence::icarus_filter_cluster3DCryoW
+  , @sequence::icarus_reco_cluster3DCryoW
   , @sequence::icarus_reco_pandoraGausCryoW
   , fmatchCryoW
   , @sequence::icarus_crttrack


wesketchum (approved):
I think this looks fine, but we also know (generally) how many fragments should be expected for good physics runs, no? Could we add a fcl parameter that is the expected fragment count, and then compare that as well? Would then work well for the first event in a file too.

icaruscode/Filters/FilterDataIntegrity_module.cc
```diff
@@ -0,0 +1,96 @@
...
+            missingFragments.erase(fragmentID);
+
+            std::string instanceName = handle.provenance()->productInstanceName();
+            std::size_t found        = instanceName.find("Empty");
+
+            if (found != std::string::npos)
+            {
+                // fragment is empty
+                emptyFragments++;
+            }
```
PetrilloAtWork, line 75: What is the role and relevance of `Empty` fragments?

icaruscode/Filters/FilterDataIntegrity_module.cc
```diff
@@ -0,0 +1,96 @@
...
+#include <fstream>
+
+/// Framework includes
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Core/EDFilter.h"
+#include "art/Framework/Principal/Event.h"
+#include "canvas/Utilities/InputTag.h"
+#include "fhiclcpp/ParameterSet.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
+#include "lardataobj/RecoBase/Hit.h"
```
PetrilloAtWork, line 15: Definitely not needed:
```suggestion
```

icaruscode/Filters/FilterDataIntegrity_module.cc
```diff
@@ -0,0 +1,96 @@
+////////////////////////////////////////////////////////////////////////
+//
+// FilterDataIntegrity class
+//
+////////////////////////////////////////////////////////////////////////
+#include <fstream>
```
PetrilloAtWork, line 6: Not needed. If you want to include something:
```suggestion
#include <set>
#include <string>
```

icaruscode/Filters/FilterDataIntegrity_module.cc
```diff
@@ -0,0 +1,96 @@
...
+            {
+                // fragment is empty
+                emptyFragments++;
+            }
+        }
+    }
+
+    mf::LogDebug("FilterDataIntegrity") << "Expected fragments: " << fExpectedFragments.size() << " expected, has " << missingFragments.size() << " missing. There are " << emptyFragments << " empty fragments";
+
+    if (missingFragments.size() > 0 || emptyFragments > 0)
```
PetrilloAtWork, line 81: Recommended the use of `empty()` to check whether a collection is empty:
```suggestion
    if (!missingFragments.empty() || emptyFragments > 0)
```

icaruscode/Filters/FilterDataIntegrity_module.cc
```diff
@@ -0,0 +1,96 @@
...
+    {
+        mf::LogInfo("FilterDataIntegrity") << "Bad fragments: " << missingFragments.size() << ", " << emptyFragments;
```
PetrilloAtWork, line 83: More info maybe?
```suggestion
        mf::LogInfo log("FilterDataIntegrity");
        log << "Bad fragments (" << missingFragments.size() << "):";
        for (int fragmentID: missingFragments) log << " " << fragmentID;
        if (emptyFragments) log << "; empty: " << emptyFragments;
```

icaruscode/Filters/CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ art_make_library()
 set(  MODULE_LIBRARIES
     sbnobj::Common_Trigger
     icarusalg::Utilities
+    sbndaq_artdaq_core::sbndaq-artdaq-core_Overlays_ICARUS
```
PetrilloAtWork, line 5: I don't think this one is needed, but:
```suggestion
    artdaq_core::artdaq-core_Data
```
may.

PetrilloAtWork (changes requested):
I need a clarification: the filter is configured to interrupt a path, but it does not seem to be bound to the output module(s). If I am not wrong, it means that the events that are filtered out will be still saved (with no data products except maybe the trigger data fragment if we are now keeping it as it was proposed). In that case, the following stages may find the event, try to process it, look for a raw waveform data product that is not there and throw `ProductNotFound` (or worse, [silently ignore the problem](https://github.com/art-framework-suite/art/issues/133) — which happens to be what we need but because of a bug). I think the solution is to add `selectEvents: [ pathname ]` to the output module(s).

For the rest, I have dropped some minor suggestions.

icaruscode/Filters/FilterDataIntegrity_module.cc
SFBayLaser: Yeah, guilty as charged. Cloned from the filter module that tests on TPC hits.

icaruscode/Filters/FilterDataIntegrity_module.cc
SFBayLaser: ditto

icaruscode/Filters/FilterDataIntegrity_module.cc
```diff
@@ -0,0 +1,96 @@
...
+/// Framework includes
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Core/EDFilter.h"
+#include "art/Framework/Principal/Event.h"
+#include "canvas/Utilities/InputTag.h"
+#include "fhiclcpp/ParameterSet.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
+#include "lardataobj/RecoBase/Hit.h"
+
+#include "sbndaq-artdaq-core/Overlays/ICARUS/PhysCrateFragment.hh"
```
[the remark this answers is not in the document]
SFBayLaser: Will check, likely you are correct.

icaruscode/Filters/CMakeLists.txt
SFBayLaser: Yes, likely not

icaruscode/Filters/FilterDataIntegrity_module.cc
SFBayLaser: Ok, in my haste I had forgotten if "empty" was universal or a properly of vectors only...

icaruscode/Filters/FilterDataIntegrity_module.cc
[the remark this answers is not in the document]
SFBayLaser: I remember now that this is not possible, our interface is through sbndaq-artdaq-core. Or at least when I tried it the other way it did not work. Unless I did something else wrong...

PetrilloAtWork:
I remind also in this venue that DAQ "promised" the addition of a list of configured components and their fragments, with in conjunction with code like this one or like [feature/gp_missingFragments](https://github.com/PetrilloAtWork/icaruscode/tree/feature/gp_missingFragments) would allow a consistent and robust check.

icaruscode/Filters/FilterDataIntegrity_module.cc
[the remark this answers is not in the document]
PetrilloAtWork: That surprises me — PMT decoder does include `Fragment.hh` (together with the SBN overlays).

icaruscode/Filters/FilterDataIntegrity_module.cc
PetrilloAtWork: In C++17 it has become an universal at the point where there is a free function `std::empty()`, so that in principle you should be able to use for any container a syntax like `empty(missingFragment)` (note the intentional absence of `std::`).

icaruscode/Filters/FilterDataIntegrity_module.cc
PetrilloAtWork: Maybe @wesketchum can comment on how `Empty` fragment(s) may arise?

icaruscode/Filters/FilterDataIntegrity_module.cc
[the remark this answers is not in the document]
SFBayLaser: when I removed "sbndaq_artdaq_core::sbndaq-artdaq-core_Overlays_ICARUS" from cmakelists I ran into a problem with TRACE.h... and being lazy I punted and put the above back in.

PetrilloAtWork (approved):
All looks good to me now.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

SFBayLaser:
trigger build

SFBayLaser:
trigger build

PR #593, opened 2023-06-08
Feature/tboone update crttpc 060823

tboone0042:
I've been needing to update the basic CRT-TPC analysis for a while. This includes new options in the CRTT0MatchAlg.* code to use PCA projection for the TPC track in lieu of the older method. This is still under investigation as a method, and as such lives under a new function name so that any code currently using CRT-TPC matching will not be changed in functionality. 

I've also updated the TSMode option for the crtt0matchingalg.fcl to 1 so that the now best way of getting the CRT Hit time is used instead of the old method of doing so. This is important for anyone using the CRT-TPC matching code as there may be issues using the old timestamp acquisition method, and is also a repeat of a calculation that was already done and stored in the CRT Hit. 

Finally, I needed some splitting up of data/MC and pandora/Kalman as I was having trouble getting the fcls to override what was set in my functions. Not sure why but it needs to be done so that I have something working. Happy to make any/all needed changes, the most important is the TSMode change since we've had issues already of people getting wrong measurements of CRT Hit time using the default mode.


mmrosenberg:
trigger build

icaruscode/CRT/CRTUtils/CRTT0MatchAlg.h
```diff
@@ -59,9 +59,31 @@
...
 
+  struct tpc_cand {
+
+	bool found_match = false;
+	recob::Track thistrack;
+	TVector3 PCA_start_pos{DBL_MAX,DBL_MAX,DBL_MAX};
+	TVector3 PCA_start_dir{DBL_MAX,DBL_MAX,DBL_MAX};
+	TVector3 PCA_end_pos{DBL_MAX,DBL_MAX,DBL_MAX};
+	TVector3 PCA_end_dir{DBL_MAX,DBL_MAX,DBL_MAX};
+	int start0_end1_whichbest = -1;
```
andrewmogan, line 82: L82-83: It's not clear to me what these variables represent. I suggest using more descriptive names.

icaruscode/CRT/CRTUtils/CRTT0MatchAlg.h
```diff
@@ -59,9 +59,31 @@
...
+
+
 
 namespace icarus{
 
+  struct tpc_cand {
+
+	bool found_match = false;
+	recob::Track thistrack;
+	TVector3 PCA_start_pos{DBL_MAX,DBL_MAX,DBL_MAX};
```
andrewmogan, line 78: In general, I'd advise against initializing values using DBL_MAX, DBL_MIN, etc. These values are OS-dependent, so someone on another machine might try "if PCA_start_pos[0] == DBL_MAX" and get different behavior. Maybe it's fine in this case since we assume this is running on gpvms, but I'd still suggest picking values that are not OS-dependent. Note that if you initialize the directions to {0}, you'll need to check the denominator in the DCA calculation to make sure it's not zero!

icaruscode/CRT/CRTUtils/CRTT0MatchAlg.h
```diff
@@ -81,6 +103,37 @@ namespace icarus{
 
   };
 
+  struct  matchCand_PCA {
+    sbn::crt::CRTHit thishit;
+    double t0 = DBL_MIN;
+    double dca = DBL_MIN;
```
andrewmogan, line 109: This is a case where I think using DBL_MIN is OK since you're definitely updating the value later.

icaruscode/CRT/CRTUtils/CRTT0MatchAlg.h
```diff
@@ -81,6 +103,37 @@ namespace icarus{
...
+    double extrapLen = DBL_MIN;
+    int best_DCA_pos = -1;//-1=no match; 0=startdir is best; 1=enddir is best; 2=both start+end are equally good;
+    bool simple_cathodecrosser = false;
+    int driftdir = -5;
```
andrewmogan, line 113: What does -5 represent? I would think this value is +/- 1. If it's just a default value, maybe NULL would be better so that you can catch cases where this isn't set correctly, where -5 would just give you a large shift value (I think).

icaruscode/CRT/crttpctrutheff_mc.fcl
```diff
@@ -54,31 +54,32 @@ outputs:
...
     {
-      module_type:         "icaruscode/CRT/CRTTPCTruthEff"
+      module_type:         "icaruscode/CRT/CRTTPCTruthEffMC"
       # The input parameters	   
       CRTHitLabel:         "crthit"           # CRT producer module label
       TPCTrackLabel:       ["pandoraTrackGausCryoE", "pandoraTrackGausCryoW"]     # Track producer module label
       PFParticleLabel: 	   ["pandoraGausCryoE", "pandoraGausCryoW"]
+      PFParticleMetadataLabel: ["pandoraGausCryoE", "pandoraGausCryoW"]
       #TPCTrackLabel:       "pandoraTrackGausCryoE"     # Track producer module label  
       TriggerLabel:        "daqTrigger"
```
andrewmogan, line 71: For MC, I find that I need to change "daqTrigger" to "emuTrigger." Unless there's been some change to this since v09_72ish, I would expect the same behavior here.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+
+    TTree* tr_crttpc;
+    int fEvent;        			///< number of the event being processed
+    int fRun;          			///< number of the run being processed
+    int fSubRun;       			///< number of the sub-run being processed
+    int crt_region;    			///< CRT hit region code
+    int ttl_tpctrks, ttl_crthits; 	///< Total number of TPC tracks and CRT Hits for a given event
+    int driftdir;			///< Drift direction of the track
+    int num_crt_candidates;		///< Number of CRT Hit candidates for a given track, including the matched hti
+	    int startflag, endflag;		///< flags for TPC track endpoints
```
andrewmogan, line 151: If there isn't a good reason for the sudden indentation here, I would try to keep the indentation consistent.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+	  //(temporary) adding all valid geo::Point_t object positions to the vectors to be able to identify PCA behavior
+	  geo::Point_t thispt;
+	  size_t nTrackPoints = trackList[track_i]->NPoints();
+	  int midPt = (int)floor(nTrackPoints*0.5);
+	  geo::Point_t midP = trackList[track_i]->LocationAtPoint(midPt);
+
+	  midpt_x = midP.X(); midpt_y = midP.Y(); midpt_z = midP.Z();
+	  track_length = trackList[track_i]->Length();
+
+	  tpc_start_x = trackList[track_i]->StartDirection().X();
```
andrewmogan, line 560:
```suggestion
	  tpc_start_dir_x = trackList[track_i]->StartDirection().X();
```
Adding "dir" here avoids confusion with start_pos_x etc.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+
+
+	  }//End loop over crtHits
+
+
+//	  int const cryoNumber = hits[0]->WireID().Cryostat;
+
+	  matchCand_PCA closest = t0Alg.GetClosestCRTHit_PCA(detProp, *trackList[track_i], hits, crtHits,  m_gate_start_timestamp, fIsData);
+
+	  if(closest.dca >=0 ){
```
andrewmogan, line 677: It's not immediately clear to me if both of these if statements are necessary. When is the DCA negative?

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+
+			auto checkhit = closest.thishit;
+			crt_trueID =  bt.TrueIdFromTotalEnergy(event, checkhit);
+			GetAncestorID(crt_trueID, crt_mother, crt_ancestor, crt_motherlayers, particles);
+			crt_pdg = particles[crt_trueID].PdgCode();
+			crt_ancestor_pdg = particles[crt_ancestor].PdgCode();
+
+		has_crtmatch=false;
+		num_crt_candidates = (int)all_crt_candidates.size();
+		if(num_crt_candidates>0){
```
andrewmogan, line 807:
```suggestion
		if (num_crt_candidates == 0) continue;
```
This will help "denest" the following block since there are a lot of nested indentations below.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+				crt_start_dca.push_back(thiscand.simpleDCA_startDir);	
+				crt_end_dca.push_back(thiscand.simpleDCA_endDir);	
+
+				all_crt_hitcode_map.push_back(i);//Gives ID to each hit for the FEB+channel+pes breakdown below to allow cross-referencing between vectors
+
+				all_crt_hitflags.push_back(thiscand.thishit.ts0_s_corr);
+
+			    	std::map< uint8_t, std::vector<std::pair<int,float> > > thishit_pes = crtcandidate.pesmap; ///< Saves signal hit information (FEB, local-channel and PE) .
+				for(uint8_t k=1; k<232; k++){
+					if(thishit_pes[k].size()>0){
```
andrewmogan, line 841:
```suggestion
					if (thishit_pes[k].size() == 0) continue;
```
Similar to the comment above about reducing nesting.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+	
+      } // validity check
+
+    } // all track labels in the product
+	std::cout << "Found " << num_tracks_badID << " tracks with bad ID, " << num_tracks_badID_greaterthan_20cm << " of which are >20cm in reconstructed track length\n";
+	std::cout << "End event #" << fEvent << "\n000000000000000000000000000000000000000000000000000000000000000000\n";
+  } // CRTTPCTruthEffMCKalman::produce()
+
+
+void icarus::CRTTPCTruthEffMCKalman::GetAncestorID(int trueid, int &motherid, int &ancestorid, int &layers, std::map<int, simb::MCParticle> all_particles){
```
andrewmogan, line 876: This function is named GetAncestorID, but it doesn't actually return the ancestor ID. It's not immediately clear to me how this gives you the ancestor ID.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+		last_ancestorid = ancestorid;
+		if(ancestorid<size_part) ancestorid = all_particles[ancestorid].Mother();
+		hitbottom = last_ancestorid == ancestorid;
+		counter++;
+	}//end while loop 
+	layers = counter;
+
+
+}//CRTTPCTruthEffMCKalman::GetAncestorID
+void icarus::CRTTPCTruthEffMCKalman::getTickRange(std::vector<art::Ptr<recob::Hit>> hits, int &mintick, int &maxtick){
```
andrewmogan, line 895: Similar comment as above. If a function has "Get" in its name, I expect a return value. Without a deeper dive, I'm not sure whether it makes more sense to add a return value here or to just rename the function. I'll leave that up to you.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+	}
+
+}//end definition of std::pair<int,int> getCatCrossYZ(std::vector<art::Ptr<recob::Hit>> trk_hits)
+bool icarus::CRTTPCTruthEffMCKalman::willEnterCRTVolume(TVector3 vecpos, TVector3 vecdir, int crtregion){
+
+	bool returnbool;
+
+	struct boundary_plane{
+		double xmin, xmax, ymin, ymax, zmin, zmax;
+		int thin_range;//<range of wall that is thinnest, x=0, y=1, z=2
```
andrewmogan, line 971: There's a variable called "thinrange" (no underscore) declared and used elsewhere. Is "thin_range" different? If not, does it make sense to pass thinrange as a function input here?

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+	all_side_planes[11].ymin = -160;       all_side_planes[11].ymax = -80;
+	all_side_planes[11].zmin = 1169.72;    all_side_planes[11].zmax = 1177.99;
+	all_side_planes[11].reg_num = 47;      all_side_planes[11].reg_name = "Side North Bottom-West";
+	all_side_planes[11].thin_range = 2;
+*/
+
+
+}//end definition of void icarus::CRTTPCTruthEffMCKalman::findIntersectingCRTVolume(recob::Track trk, TVector3 &retvec, TVector3 &endpoint, int &likely_CRT_region)
+
+void icarus::CRTTPCTruthEffMCKalman::endpoint_PCA_ana(recob::Track trk, bool usestartpt, TVector3 &bestfit_dir, TVector3 &bestfit_pos){
```
andrewmogan, line 1437: It seems like the purpose of this function is to update the PCA best-fit position and direction. If so, and if it's not too much trouble, I'd suggest naming the function something like "GetPCAParams" or something.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+	x_start_check = std::abs(startpoint.X())>=xmin && std::abs(startpoint.X())<=xmax;
+	x_end_check = std::abs(endpoint.X())>=xmin && std::abs(endpoint.X())<=xmax;
+	y_start_check = startpoint.Y()>=ymin && startpoint.Y()<=ymax;
+	y_end_check = endpoint.Y()>=ymin && endpoint.Y()<=ymax;
+	z_start_check = startpoint.Z()>=zmin && startpoint.Z()<=zmax;
+	z_end_check = endpoint.Z()>=zmin && endpoint.Z()<=zmax;
+
+	startisinTPC = x_start_check && y_start_check && z_start_check;
+	endisinTPC = x_end_check && y_end_check && z_end_check;
+
```
andrewmogan, line 1607: It would be good to note somewhere what the different values of startflag and endflag mean.

icaruscode/CRT/CRTTPCTruthEffMCKalman_module.cc
```diff
@@ -0,0 +1,1721 @@
...
+	else if(x_end_check && y_end_check && !z_end_check) endflag=3;
+	else endflag=4;
+
+}//end definition of void icarus::CRTTPCTruthEffMCKalman::pointChecker(recob::Track trk, bool &startisinTPC, bool &endisinTPC)
+
+double icarus::CRTTPCTruthEffMCKalman::getnewDCA(TVector3 CRTHitpos, TVector3 TPC_posvec, TVector3 TPC_dirvec){
+
+    double denominator = TPC_dirvec.Mag();
+    double numerator = (CRTHitpos - TPC_posvec).Cross(CRTHitpos - TPC_posvec - TPC_dirvec).Mag();
+    return numerator/denominator;
```
andrewmogan, line 1627: I suggest checking that the denominator isn't zero. If it is, you can set it to some large number.

icaruscode/CRT/CRTTPCTruthEffMC_module.cc
andrewmogan, line 1: There appear to be a lot of redundant functions between this file and the Kalman version. Consider if it makes sense to put the redundant functions into a separate file (`CRTTPCTruthEff_Functions.cc/hh` or something) and calling them when necessary in each module.

icaruscode/CRT/CRTTPCTruthEff_module.cc
andrewmogan, line 1: Again, it looks like a lot of functions here were copied from the other modules. Consider putting re-used functions in a separate file and calling them when necessary.

andrewmogan (commented):
When I tried to build this branch locally, I had to update the icarusalg and sbncode versions in icaruscode/ups/product_deps. I'm not exactly sure how this works when merging into a develop branch, as I would think the dependencies update when develop updates. This is something to check.

icaruscode/CRT/CRTUtils/CRTT0MatchAlg.h
tboone0042: would NULL get translated to 0 if the int is used in multiplication, though, or would it create a NaN?

tboone0042:
hi all, I'm closing this as I've made a few changes as well as updated the version, I'm going to hold off on a full PR for now while I figure a few more things out including upgrading the Kalman analysis

PR #597, opened 2023-06-14
Fix to PMT decoder bug

PetrilloAtWork:
When the configuration requests not to save any waveform (or maybe just the standard PMT waveforms), the decoder module would not declare them to _art_, but then it would attempt to put them into the _art_ event anyway, which _art_ did not like at all.

Fixed now removing the attempt to put the waveforms in the event in that condition.
Not a big deal, no hurry.

Tested with `stage0_run2_icarus.fcl` (verified that the waveforms are still saved) and with a special version of `decodePMT_icarus.fcl` with no waveform saving (verified that the commit fixes the issue).

No idea whom could review this... @mvicenzi? @SFBayLaser :black_joker:?

A twin PR #598 takes care of the production branch.


PetrilloAtWork:
Added @mvicenzi as reviewer for lack of empathy.

mvicenzi (approved):
Looks good to me!

PR #610, opened 2023-07-18
Provide now-required QAlpha parameter

knoepfel:
This PR simply copies the `QAlpha` parameter as listed here:

https://github.com/LArSoft/larsim/blob/3e844a9349adfc27dd7ce8ac5141ff3eae764c67/larsim/Simulation/simulationservices.fcl#L59C14-L59C14

It would be better for ICARUS to refer to the parameters from the LArSoft `simulationservices.fcl` file through the `@local::` command than to just explicitly copy each parameter.


mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_75_03

PR #611, opened 2023-07-24
YZ 2D Simulation and Separate Electron Lifetime Simulation

rennney:
Add configurations to support new WireCell module to configure eLifetime per TPC

Currently split is configured based on anode dimensions. However, it can be possible to extend the functionality to add custom input / or add separate module depending on required input in the future. 

**This PR should be merge only with new version of WireCell**

This PR suppose to break several upstream features, since it now outputs 8 simchannel versions (2 for each TPC). Numbering should be as following 0-1 : TPCEE , 2-3 : TPCEW , 4-5 : TPCWE , 6-7 : TPCWW. Contact me if this is not working correctly.
For this test I add @gputnam and @SFBayLaser as reviewers. [I will let you know as soon as new WireCell version is available]

Although I did several tests to make sure WC module and configurations are working correctly I suggest more validation to ensure eLifetime is correctly set for appropriate TPC.


gputnam (changes requested):
Hi Sergey, thank you for this.

I'm going to request what I brought up in the meeting on Tuesday. If we merge this into our mainline detector simulation, it will break downstream code right now. 

Either this needs to be put into a special configuration, or it needs to make a single SimChannel object, or we need to modify that downstream code.

I believe both CAFMaker and the Calibration NTuple-r would need modifications.

rennney:
Hi, @gputnam as was decided during the meeting it is straightforward to sum all simchannels upstream at the moment . The summation from WC will come later. Since I am not an expert in the upstream code I hope you or @SFBayLaser could add required corrections to this PR or create an extra one.

jzennamo:
this has been replaced.

PR #614, opened 2023-07-28
Read PMT timing corrections from local db 

mvicenzi:
This PR contains two upgrades:

1. PMT timing corrections are now read from local databases in `icarus_data`.
2. An additional set of timing corrections from cosmic muons is now applied to data.

The previous implementation of `PMTTimingCorrectionsProvider` read the timing corrections for the PMTs (cable delays and laser) from the online postgres databases, querying via HTTPS. This has long been discouraged since it can lead to performance issues, makes versioning the corrections difficult (once a table is online, it gets picked up immediately) and it is sentitive to expiring SSL certificates. This PR fixes the issues by reading the corrections from a local version of the databases in `icarus_data` using `DBFolder` from `larevt`.

Moreover, cosmics timing corrections are now read and applied to data using the same infrastructure of the laser corrections. These corrections are applied by default to data, but can be disabled by setting the fcl parameter `CorrectCosmics` (from icaruscode/Timing/timing_icarus.fcl) to `false` (feature requested by @francescopoppi).

Asking a review from @jedori0228, @SFBayLaser and @PetrilloAtWork.

**Note to release managers**: This PR _requires_ a file named `pmt_cosmics_timing_data.db` to be present in `icarus_data` or an exception is thrown. It is therefore important to simultaneously release a new version of `icarus_data` that contains that file. Several updates to all PMT db files are expected in the near future, but a first working version of `pmt_cosmics_timing_data.db` is in the hands of @jedori0228.


jedori0228 (approved):
Timing correction module looks good to me. I am working on updating db files in `icarus_data`.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,92 @@
...
-            double phase_correction = getDoubleValue( tuple, 13, &error );
-            if( error ) throw std::runtime_error( "Encountered error while trying to access 'phase_correction' on table " + name );
+    const std::string dbname("pmt_cables_delays_data");
+    lariov::DBFolder db(dbname, "", "", fTag, true, false);
+
+    bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
+    if( !ret ) throw std::runtime_error( "Unable to find or open " + dbname + ".db");
+
+    std::vector<unsigned int> channelList;
+    ret = db.GetChannelList(channelList);
```
PetrilloAtWork, line 65: I don't know if there is a "protocol", but the current implementation of `DBFolder::GetChannelList()` always returns `0`. I would either set a check, or ignore the return value altogether:
```suggestion
    if (int res = db.GetChannelList(channelList); res != 0) {
      throw std::runtime_error
        ( "PMTTimingCorrectionsProvider: GetChannelList() returned " + std::to_string(res) + " on run " + std::to_string(run) + " query.");
    }
```
In addition: can we print a warning/error if the list of channels is empty? or throw an exception?
```suggestion
    if (channelList.empty()) {
      throw std::runtime_error("PMTTimingCorrectionsProvider: got an empty channel list for run " + std::to_string(run));
    }
```
BTW, we should have used `cet::exception` (the only reason not to do that is to reduce dependencies, but since we are using messagefacility the dependency on `cetlib_except` is already there).

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,92 @@
...
-            if( error ) throw std::runtime_error( "Encountered error while trying to access 'trigger_reference_delay' on table " + name );
-
-            // Phase correction
-            double phase_correction = getDoubleValue( tuple, 13, &error );
-            if( error ) throw std::runtime_error( "Encountered error while trying to access 'phase_correction' on table " + name );
+    const std::string dbname("pmt_cables_delays_data");
+    lariov::DBFolder db(dbname, "", "", fTag, true, false);
+
+    bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
+    if( !ret ) throw std::runtime_error( "Unable to find or open " + dbname + ".db");
```
PetrilloAtWork, line 62: Have you tested this running on multiple files?
It looks like `DBFolder::UpdateData()` returns `false` if the cache was not updated, that should happen also when the cache is already current. When a new file is opened, _art_ believes a new run is being processed, so this method may end up being called also on file boundaries, not only on run boundaries.
My suggestion is to ignore the return value, or rather:
```suggestion
    mf::LogDebug(fLogCategory) << "PMT corrections" << (ret? "": " not") << " updated for run " << run;
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,92 @@
...
+
+    std::vector<unsigned int> channelList;
+    ret = db.GetChannelList(channelList);
+
+    for( auto channel : channelList ) {
+        
+        // PPS reset correction
+        double reset_distribution_delay = 0;
+        int error  = db.GetNamedChannelData( channel, "reset_distribution_delay", reset_distribution_delay );
+        if( error ) throw std::runtime_error( "Encountered error while trying to access 'reset_distribution_delay' on table " + dbname );
```
PetrilloAtWork, line 72: This is good.
Just mentioning that the current implementation always returns `0`.
Well, since it's still me and I can't be seen around saying that anything is good:
```suggestion
        if( error ) throw std::runtime_error( "Encountered error (code " + std::to_string(error) + ") while trying to access 'reset_distribution_delay' on table " + dbname );
```

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -19,32 +19,24 @@
...
+  /// Structure for single channel corrections
+  struct PMTTimeCorrectionsDB { 
 
-    double laserCableDelay=0; ///< Expect nanoseconds.
-
-    double cosmicsCorrections=0; ///< Expect nanoseconds.
+    double triggerCableDelay=0;  ///< ns
+    double resetCableDelay=0;    ///< ns
+    double laserCableDelay=0;    ///< ns
+    double cosmicsCorrections=0; ///< ns
```
PetrilloAtWork, line 39: Unless we store in the database in picoseconds, all these are actually microseconds:
```suggestion
    double triggerCableDelay=0;  ///< [&micro;s]
    double resetCableDelay=0;    ///< [&micro;s]
    double laserCableDelay=0;    ///< [&micro;s]
    double cosmicsCorrections=0; ///< [&micro;s]
```
(see e.g. `.cosmicsCorrections = -mean_residual_ns/1000` above).

icaruscode/Timing/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ set(	LIB_LIBRARIES
         wda::wda
         icaruscode_IcarusObj
         sbnobj::Common_PMT_Data
+	larevt::CalibrationDBI_IOVData
```
PetrilloAtWork, line 16: Is the dependency on `wda::wda` above still needed?

PetrilloAtWork (changes requested):
My usual rain of comments...
The important one: confirm that this works when processing multiple files with the same run. I may be wrong, but I am afraid that on the second file with the same run an exception could be thrown.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: My feeling is that return values in `DBFolder` were never properly properly set, so most of the checks I have are actually useless (at least for now). Anyway, I agree with both of your suggestions.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: Yeah, these checks are currently useless... Anyway, adding your suggestion as well.

icaruscode/Timing/PMTTimingCorrectionsProvider.h
mvicenzi: Point taken.

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -54,74 +46,57 @@ namespace icarusDB{ class PMTTimingCorrectionsProvider; }
...
- * Input
- * ------
- * 
- * 
- * Output
- * -------
- * 
- * 
+ * This module reads the PMT timing corrections from the database.
+ * Corrections are picked according to the run number being processed.
```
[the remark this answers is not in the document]
mvicenzi: Absolutely. Adding this as well.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: I repeated the test, running of multiple files (same run). The corrections are read again when the next file is opened, and no exception is thrown. I originally added this check as a way to catch missing `.db` files, but I noticed that if a file is missing `DBFolder::UpdateData()` trips internally on its own (producing an exception) before being able to return `false`. So I believe your suggestion to just log it is the best course of actions as I have no intention of changing `DBFolder`.

icaruscode/Timing/CMakeLists.txt
mvicenzi: Not anymore. I already removed the `#include` statements, but I forgot to remove it from `CMakeLists.txt`.

PetrilloAtWork (approved):
Good from my side, and another "thank you" token for @mvicenzi's stack.

mvicenzi:
> @mvicenzi should we have this change ported to the production branch too?

Yes, it should be ported as well. Adding it will get rid of the dependency on the connection to the online database. However, it is essential to update `icarus_data ` with the new `.db ` file for cosmics correction before proceeding.

mvicenzi:
I added two other commits to take into account different tags for each correction database and defined two "standard" fcl configurations (as of now).

mmrosenberg:

PR #615, opened 2023-07-31
ArtHandleTrackerManager can now handle invalid handles

PetrilloAtWork:
It was reported a segmentation fault from a module using `ArtHandleTrackerManager`. It turns out that the input of that module was defective and the handle being created was invalid, which `ArtHandleTrackerManager::getHandle()` could handle not.
Now it handles that scenario by not attempting to track an invalid handle and returning the invalid handle to the caller, similarly to the `art::Event::getHandle()` call that it is wrapping.

So that job will still fail, but with a meaningful error.
This is a bug fix of not very critical importance.

PR #616 is the twin change for production branch.


mmrosenberg:
trigger build

PR #617, opened 2023-08-08
CopyBeamTimePMTwaveforms module to support MC

PetrilloAtWork:
The `CopyBeamTimePMTwaveforms` is used for size reduction of data samples.
It makes a copy of selected PMT waveforms, only if they contain a time point.

In production, that point is chosen to be the time of opening of the beam gate. This effectively forces the preservation of a waveform that extends from -7 to +19 µs from the beam gate front.

For a potential future use in MC samples, we would like the coverage to be similar.
This module could not guarantee that, since the duration of the waveform around the beam gate is shorter.
This pull request changes the module to accept a time interval instead of a single time point, and it will copy all the waveforms that overlap with that time interval. In this way it is guaranteed that no information is lost (there may still be gaps in that interval, but they are not covered by any existing information — in fact, in the current configuration these gaps could only contain noise).

This pull request includes:
* extension of `CopyBeamTimePMTwaveforms` to support a time interval instead of time point.
* update of _data_ Stage0 configurations of this module to cover the interval [ -7 ; +19 ] µs from the beam gate opening.
* addition to _MC_ Stage0 configuration of this module (but this module is still *not* executed in any MC Stage0 or Stage1 job).
* "bonus": `CopyBeamTimePMTwaveforms` can now also copy the associations to baseline and RMS data products (without this, after waveforms are dropped the discovery of baseline of each waveform requires major hoops).
* "bonus": extended PMT digitisation module to produce `sbn::OpDetWaveformMeta` objects that preserve a bit of information for all the waveforms even when the waveforms are dropped; this is to match the behaviour of ICARUS PMT decoder.

This branch takes advantage of an utility introduced by SBNSoftware/icarusalg#71.

The branch was tested with `run2_stage0_icarus.fcl`, a variant of `run2_stage0_icarus_mc.fcl` and one of `run2_stage1_icarus_MC.fcl` (with different capitalisation... Tracy!!!).

Reviewers: not very sure here. I throw two names here, but if they push back I am not sure who else.

### DIFF
--- a/icaruscode/PMT/CopyBeamTimePMTwaveforms_module.cc
+++ b/icaruscode/PMT/CopyBeamTimePMTwaveforms_module.cc
@@ -9,6 +9,8 @@
 // ICARUS libraries
 #include "icaruscode/IcarusObj/OpDetWaveformMeta.h"
 #include "icaruscode/PMT/Data/WaveformRMS.h"
+#include "icarusalg/Utilities/TimeIntervalConfig.h"
+#include "icarusalg/Utilities/TimeInterval.h"
 #include "sbnobj/ICARUS/PMT/Data/WaveformBaseline.h"
 
 // LArSoft libraries
@@ -16,6 +18,7 @@
 #include "lardataalg/DetectorInfo/DetectorTimings.h"
 #include "lardataalg/DetectorInfo/DetectorTimingTypes.h" // electronics_time
 #include "lardataalg/Utilities/MultipleChoiceSelection.h"
+#include "lardataalg/Utilities/intervals_fhicl.h" // for intervals from FHiCL
 #include "larcorealg/CoreUtils/enumerate.h"
 #include "lardataobj/RawData/OpDetWaveform.h"
 
@@ -40,13 +43,17 @@
 
 
 //------------------------------------------------------------------------------
+using namespace util::quantities::time_literals;
+
+//------------------------------------------------------------------------------
 namespace icarus { class CopyBeamTimePMTwaveforms; }
 
 /**
  * @brief Creates a new list of PMT waveforms covering the specified time.
  * 
  * This module creates a new data product with all and only the PMT waveforms
- * from an existing data product that cover the configured time point.
+ * from an existing data product that overlap at least in part with the
+ * configured time interval.
  * Some associations are replicated: a new association is produced from the
  * elements of the new data product to the old elements associated to the
  * original PMT waveform data product; the associated data is not replicated.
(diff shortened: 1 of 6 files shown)


PetrilloAtWork:
I have found a preexisting bug in the same module, and I am piggybacking on this PR to fix it.
I'll need the fix in production too.

icaruscode/PMT/CopyBeamTimePMTwaveforms_module.cc
```diff
@@ -476,20 +629,33 @@ auto icarus::CopyBeamTimePMTwaveforms::getReferenceTime
...
   using util::quantities::points::microsecond;
   
   electronics_time const startWaveformTime
     { microsecond{ waveform.TimeStamp() } };
-  electronics_time const endWaveformTime
-    = startWaveformTime + fOpticalTick * waveform.size();
+    
+  Interval_t overlap
+    { startWaveformTime, startWaveformTime + fOpticalTick * waveform.size() };
+  overlap.intersect(time);
```
mvicenzi, line 643: What units are being used in this operation `startWaveformTime + fOpticalTick * waveform.size()`?
`startWaveformTime` is defined in microsecond a few lines above, but what about `fOpticalTick`? `detTimings.OpticalClockPeriod()` returns `auto` and `fOpticalTick` is defined as a `nanoseconds const`, so is it in nanoseconds? Am I missing a conversion somewhere?

Same question could be applied to the old code...

mvicenzi (changes requested):
I have just one concern related to the time interval which should be addressed before merging. Evertything else looks good. Great to see baselines emerging from the depths.

icaruscode/PMT/CopyBeamTimePMTwaveforms_module.cc
PetrilloAtWork: Thank you for asking :sweat_smile:

`startWaveformTime` is a variable of `electronics_time` type, which is a time point derived from `util::quantities::points::microsecond`.
`fOpticalHits` is a time interval derived from `util::quantities::intervals::nanoseconds`.
These two are derivation of a [quantity type](https://nusoft.fnal.gov/larsoft/doxsvn/html/structutil_1_1quantities_1_1concepts_1_1Quantity.html#details) `second`; each of the types derived from [`second`](https://nusoft.fnal.gov/larsoft/doxsvn/html/namespaceutil_1_1quantities.html#ae3f21a6b9f866534e769a19192532eb4) are stored in a unit and can convert to any other. So if you have `microsecond{4} + nanosecond{20}` (or, with literals, `4_us + 20_ns`), it gets converted into a `microsecond` of value `4.02` (more precisely, the `operator+(microsecond, nanoseconds)` converts the second operand into `microseconds` and then does numerical addition.

You have been saved by a failed C&P on a more detailed description of quantities, intervals, points, and `detinfo::DetectorClocksWithUnits` and [`detinfo::DetectorTimings`](https://nusoft.fnal.gov/larsoft/doxsvn/html/classdetinfo_1_1DetectorTimings.html#details).

PetrilloAtWork:
The question by @mvicenzi should be addressed in the comment.

icaruscode/PMT/CopyBeamTimePMTwaveforms_module.cc
mvicenzi: It's clear now, thank you!

mvicenzi (approved):

PR #618, opened 2023-08-09
Trigger efficiency fcl updates and change to standard running of CRT/TPC matching

jzettle:
This PR encompasses fhicl updates in order to process over some of the runs taken with the trigger adders configuration for the trigger efficiency work. Why this PR is going in now (it is harmless to anything in icaruscode itself it is just fcl file updates) is it contains an important update to the way the CRT/TPC matching code is run. The module should be running with TSMode: 1 but has been running with TSMode: 2 which is an old method of running the matching. This is important for the efficiency studies so the main purpose of the PR is the make that change.


tboone0042:
If it's not a huge thing to add, perhaps also change the default loaded to TSMode in icaruscode/CRT/CRTUtils/CRTT0MatchAlg.cc, line 21? that way there's no shot the "old" TSMode gets loaded again? Otherwise it looks good to me!

jzettle:
Made the change @tboone0042 requested and committed again

mmrosenberg:

tboone0042 (approved):
All looks good!! Thanks for doing this Jacob!

mmrosenberg:
trigger build

PR #620, opened 2023-08-11
Run-period-dependent channel mapping

PetrilloAtWork:
Due to the change in cabling of the PMT, a slightly different mapping will required from the time of the swap on.
Using the old mapping would result in some of the channels being applied the wrong corrections.

This pull request introduces in `ICARUSChannelMapService`, its provider and the implementing tools a run-period awareness, and a simple class to (hard-codedly) map run numbers to periods, `icarusDB::RunPeriods`, is also provided.

This is technically a breaking change: the service _provider_ (`ICARUSChannelMapProvider`) used to load the data at construction; now this is not true any more, and a `withRun()` or `withPeriod()` call must be executed before the object is available for use. This is automatically done by the _art_ service wrapper, ``ICARUSChannelMapService`, but users not using the service but rather the provider directly, e.g. in a _gallery_ environment, will have to make that call now.
To ensure the most backward compatibility, the service _art_ does load the run period `Runs0to2` immediately at construction, so that code that used to cache the mapping at construction or begin of job will still work with old data.
I am not aware of such code for PMT in `icaruscode`, and `Stage0` did not show any. For the other subdetectors, mapping is always the same; should that change in the future, a code survey will be needed.

I have tested the code with a new SQLite database provided by @mvicenzi, mocking up the run period boundary.
I have noticed no difference in the TPC hits, and the expected differences in the optical hits (that is, no difference with respect to the output obtained from a production release when using run period `Runs0to2`, and differences on the changed channels for the new period).

**THIS IS A DRAFT PR**, but please start reviewing it now so that when the hardware change is happening we are ready to update it and merge it. Merging must happen after:
 * the new PMT mapping DB is added to `icarus_data`
 * the cable swap has happened on hardware
 * the code has been updated to reflect the actual run number boundary between the old and the new period

Asking for review from TPC, PMT and CRT people.

### DIFF
--- a/icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
+++ b/icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
@@ -5,6 +5,9 @@
  *
  */
 
+// ICARUS libraries
+#include "icaruscode/Decode/ChannelMapping/RunPeriods.h"
+
 // Framework Includes
 #include "art/Framework/Core/EDProducer.h"
 #include "art/Framework/Services/Registry/ServiceHandle.h"
@@ -23,6 +26,8 @@
 #include <string>
 #include <iostream>
 #include <memory>
+#include <array>
+#include <cassert>
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 // implementation follows
@@ -42,9 +47,19 @@ public:
     explicit ChannelMapPostGres(fhicl::ParameterSet const &pset);
 
     /**
-     *  @brief  Destructor
+     * @brief   Prepares the object for queries pertaining the specified period.
+     * @param   period the period to be prepared for
+     * @return  whether values cached from the previous period are invalidated
+     * 
+     * Periods are defined in `icarusDB::RunPeriods` class.
+     * 
+     * If the return value is `true`, caller should follow by rebuilding all the
+     * maps they need to use.
+     * If the return value is `false`, all maps built after the previous call to
+     * `SelectPeriod()` are still current for `period`: building new ones is not
+     * necessary, but it's not harmful either (at most, wasteful).
      */
-    ~ChannelMapPostGres();
+    virtual bool SelectPeriod(RunPeriod period) override;
 
     /**
      *  @brief Define the returned data structures for a mapping between TPC Fragment IDs
@@ -84,13 +99,53 @@ public:
   virtual int BuildSideCRTCalibrationMap(SideCRTChannelToCalibrationMap&) const override;
 
 private:
+
+    /// Record of all relevant table names.
+    struct TableNames_t {
+      std::string TPCfragmentMap;
+      std::string TPCreadoutBoardMap;
+      std::string PMTfragmentMap;
+      std::string CRTsideMap;
+      std::string CRTtopMap;
+      std::string CRTsideCalibrationMap;
+    };
+
     // Recover data from postgres database
     int GetDataset(const std::string&, const std::string&, const std::string&, Dataset&) const;
     int GetCRTCaldata(const std::string&, const std::string&, Dataset&) const;
     uint32_t fNothing;     //< Nothing
 
+    /// The set of tables being served. Chosen by `SelectRun()`.
+    TableNames_t const* fCurrentTable = nullptr;
+
+    /// The list of names of tables.
+    static std::array<TableNames_t, RunPeriods::NPeriods> const TableNameSets;
+
 };
 
+//------------------------------------------------------------------------------------------------------------------------------------------
+std::array<ChannelMapPostGres::TableNames_t, RunPeriods::NPeriods> const
+ChannelMapPostGres::TableNameSets{
+  TableNames_t{
+      "icarus_hw_readoutboard"    // TPCfragmentMap
+    , "icarus_hardware_prd"       // TPCreadoutBoardMap
+    , "Pmt_placement"             // PMTfragmentMap
+    , "Feb_channels"              // CRTsideMap
+    , "topcrt_febs"               // CRTtopMap
+    , "SideCRT_calibration_data"  // CRTsideCalibrationMap
+  },
+  TableNames_t{
+      "icarus_hw_readoutboard"    // TPCfragmentMap
+    , "daq_channels"              // TPCreadoutBoardMap
+    , "Pmt_placement"             // PMTfragmentMap      TODO need the new table!
+    , "Feb_channels"              // CRTsideMap
+    , "topcrt_febs"               // CRTtopMap
+    , "SideCRT_calibration_data"  // CRTsideCalibrationMap
+  }
+};
+
+
+//------------------------------------------------------------------------------------------------------------------------------------------
 ChannelMapPostGres::ChannelMapPostGres(fhicl::ParameterSet const &pset)
 {
     fNothing = pset.get<uint32_t>("Nothing");
@@ -99,10 +154,31 @@ ChannelMapPostGres::ChannelMapPostGres(fhicl::ParameterSet const &pset)
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
-
-ChannelMapPostGres::~ChannelMapPostGres()
-{
+bool ChannelMapPostGres::SelectPeriod(RunPeriod period) {
+  
+  auto const iPeriod = static_cast<unsigned int>(period);
+  TableNames_t const* newSet = &(TableNameSets.at(iPeriod));
+  
+  if (fCurrentTable == newSet) {
+    mf::LogDebug("ChannelMapPostGres")
+      << "Period #" << iPeriod << " already selected";
+  }
+  else if (fCurrentTable) {
+    mf::LogDebug("ChannelMapPostGres") << "Switched from period #"
+      << (fCurrentTable - &(TableNameSets[0])) << " to #" << iPeriod;
+  }
+  else {
+    mf::LogDebug("ChannelMapPostGres") << "Switching to period #" << iPeriod;
+  }
+  
+  if (fCurrentTable == newSet) return false;
+  
+  fCurrentTable = newSet;
+  return true;
+  
 }
+
+//------------------------------------------------------------------------------------------------------------------------------------------
 
 // -----------------------------------------------------
 // This Function does the basic information retrieval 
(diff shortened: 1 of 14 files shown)


icaruscode/Timing/PMTWaveformTimeCorrectionExtractor.cxx
```diff
@@ -47,11 +47,11 @@ icarus::timing::PMTWaveformTimeCorrectionExtractor::Error::Error
...
 ::MultipleCorrectionsForChannel
-  (unsigned int existing, unsigned int additional)
+  (unsigned int channel, unsigned int existing, unsigned int additional)
     : Error{
-      "Attempt to overwrite correction from channel "
-      + std::to_string(existing) + " with one from channel "
-      + std::to_string(additional) + "\n"
+      "Attempt to overwrite the correction for channel "
+      + std::to_string(channel) + " (from channel " + std::to_string(existing)
+      + ") with another (from channel " + std::to_string(additional) + ")\n"
```
mvicenzi, line 54: I'm a little confused by these changes. What is this exception trying to catch?

icaruscode/Decode/ChannelMapping/RunPeriods.h
```diff
@@ -0,0 +1,84 @@
...
+  
+  /**
+   * @brief Returns the period of the specified `run`.
+   * @return the period `run` belongs to, or `NPeriods` if not supported
+   */
+  static constexpr RunPeriod withRun(int run)
+    {
+      
+      if (run < 0xBADCAFE) return RunPeriod::Runs0to2; // FIXME set the correct run number
+      if (run < 0xBADCAFE) return RunPeriod::Runs3andOn; // FIXME set the correct run number
```
mvicenzi, line 73: The run number itself is still unknown, but I imagine the second `<` should actually be a `>=`.

icaruscode/Decode/ChannelMapping/ChannelMapSQLite_tool.cc
```diff
@@ -101,11 +134,51 @@ ChannelMapSQLite::ChannelMapSQLite(fhicl::ParameterSet const &pset)
...
+  }
+  else if (fCurrentTable) {
+    mf::LogDebug("ChannelMapSQLite") << "Switched from period #"
+      << (fCurrentTable - &(TableNameSets[0])) << " to #" << iPeriod;
+  }
+  else {
+    mf::LogDebug("ChannelMapSQLite") << "Switching to period #" << iPeriod;
+  }
+  
+  if (fCurrentTable == newSet) return false;
```
mvicenzi, line 174: Is there a reason for not adding `return false` directly in the same `if` statement above? Just curious.

icaruscode/Decode/ChannelMapping/ICARUSChannelMapProvider.cxx
```diff
@@ -63,13 +102,15 @@ ICARUSChannelMapProvider::ICARUSChannelMapProvider(const fhicl::ParameterSet& ps
...
+    fReadoutBoardToChannelMap.clear();
     if (fChannelMappingTool->BuildTPCReadoutBoardToChannelMap(fReadoutBoardToChannelMap))
     {
         std::cout << "******* FAILED TO CONFIGURE CHANNEL MAP ********" << std::endl;
         throw cet::exception("ICARUSChannelMapProvider") << "POS didn't read the F'ing database again \n";
     }
 
     // Do the channel mapping initialization
+    fFragmentToDigitizerMap.clear();
     if (fChannelMappingTool->BuildFragmentToDigitizerChannelMap(fFragmentToDigitizerMap))
```
mvicenzi, line 114: Is it possible to rename this function to `BuildPMTFragmentToDigitizerChannelMap`? I think it would be clearer and also in line with the same functions for the other subsystems.

icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
```diff
@@ -84,13 +99,53 @@ class ChannelMapPostGres : virtual public IChannelMapping
...
+    static std::array<TableNames_t, RunPeriods::NPeriods> const TableNameSets;
+
+};
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+std::array<ChannelMapPostGres::TableNames_t, RunPeriods::NPeriods> const
+ChannelMapPostGres::TableNameSets{
+  TableNames_t{
+      "icarus_hw_readoutboard"    // TPCfragmentMap
+    , "icarus_hardware_prd"       // TPCreadoutBoardMap
```
mvicenzi, line 131: There is a copy and paste error here: `"icarus_hardware_prd"` is constant for all queries (it's the `dbname`). The TPC readout board map table should be simply `"daq_channels"` as below.

icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
```diff
@@ -302,7 +379,8 @@ int ChannelMapPostGres::BuildTPCFragmentIDToReadoutIDMap(TPCFragmentIDToReadoutI
 const unsigned int CHANNELSPERBOARD = 64;
 int ChannelMapPostGres::BuildTPCReadoutBoardToChannelMap(TPCReadoutBoardToChannelMap& rbChanMap) const
 {
-    const std::string  name("icarus_hardware_prd");
+    assert(fCurrentTable);
+    const std::string  name(fCurrentTable->TPCreadoutBoardMap);
     const std::string  dburl("https://dbdata0vm.fnal.gov:9443/QE/hw/app/SQ/query?dbname=icarus_hardware_prd");
     const std::string  dataType("daq_channels");
```
mvicenzi, line 385: I'm a bit confused by the usage of `name `and `dataType`. The variable that is actually being used as the table name in the URL query (`?t=`) further down in `GetDataset()` is `dataType`. My feeling is that `name` is a just a label required by `getDataWithTimeout()`.

mvicenzi (approved):
I added a few comments. The most relevant ones are related to the PostGres tool, which is not used anyway so it doesn't make much sense to spend time on it. Since everything else looks okay to me, I approve.

icaruscode/Timing/PMTWaveformTimeCorrectionExtractor.cxx
PetrilloAtWork: This is about the timing correction from the digitised waveforms (the reference probably will always be the global trigger signal), which we use to synchronise the waveforms from all the readout boards. Cable delay corrections are also added at the same time.
The action starts from the decoder, which after decoding and appropriately merging all the waveforms is now extracting time corrections. Corrections may be extracted from all the special waveforms (the ones on the 16th channel). The decoder is filling a table (`vector`) with a correction value for each regular PMT channel, by asking  the correction extractor to process each of the special waveforms ([`icarus::timing::PMTWaveformTimeCorrectionExtractor::findWaveformTimeCorrections`](icaruscode/Timing/PMTWaveformTimeCorrectionExtractor.cxx#L174)). That function extracts the correction value from one special waveform (an unassuming [`getStartSample()`](icaruscode/Timing/PMTWaveformTimeCorrectionExtractor.cxx#L195) and then fills the correction value for all the (45) regular channels in the crate that waveform belongs to.
This exception is thrown when this happens while one of the waveforms has already a correction, which is not expected. For example, _imagine_, if you can, that the mapping is wrong and a channel ID, say, `225`, appears multiple times, either in the same or in a different one. That means that the channel is registered in two different crates (or in the same crate twice). When we process the special waveform for the first crate, we assign all its 45 channels a correction value (e.g. from channel `0x6051`), including channel `225`. Then, when we meet the special waveform of the second crate (`0x6061`?), we do the same, 45 channels, and `225` again among them. There is where the problem shows. If the two instances appear under the same crate, little changes (it would show twice `0x6051` instead of mentioning `0x6051` and `0x6061`).
The message is now telling which special channel the already `existing` correction was extracted from, from which special channel we are trying to extract it again, and the regular channel we were filling the correction for.

icaruscode/Decode/ChannelMapping/RunPeriods.h
PetrilloAtWork: The idea was to have a large number, but not "infinite", so that a blatant wrong number triggers the special no-period value. My final solution might have looked like:
```
      if (run < 10452) return RunPeriod::Runs0to2;
      if (run < 0xBADCAFE) return RunPeriod::Runs3andOn;
```
I don't have strong opinions, so I can just remove that second `if` if you think it's better.

icaruscode/Decode/ChannelMapping/ChannelMapSQLite_tool.cc
PetrilloAtWork: Not a string one. I was separating the functional part (`if (fCurrentTable == newSet) return false;`) from he debugging message part, which may change or be removed (ok, more honestly, which was added later).

icaruscode/Decode/ChannelMapping/ICARUSChannelMapProvider.cxx
PetrilloAtWork: That goes against the principle of "not making this code better".

icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
PetrilloAtWork: I may have fixed this an even number of times, while odd is required. :facepalm:

SFBayLaser:
I am reminded that while I copied the SQLite file to icarus_data I neglected to push... I will do so tomorrow morning.

icaruscode/Timing/PMTWaveformTimeCorrectionExtractor.cxx
mvicenzi: Thank you! it's clear now.

mmrosenberg:

SFBayLaser:
Matthew was kind enough to release icarus_data as v09_77_00 which has the sqlite tables provided by Jaesung so that should be ready to go.

SFBayLaser (approved):
works for me with promise to fix the Postgres side too

francescopoppi (approved):
I reviewd this PR and it looks fine to me.
Soon, after an automated CRT calibration will be available, I will have to speak with Tyler and Valerio (and Jaesung) to move to a run-period-dependent channel calibration for the CRTs as done in this PR, instead of using an hard-coded date.
Thank you Gianluca.

mmrosenberg:

PR #626, opened 2023-08-31
Update laser calibration tags for run2

mvicenzi:
This PR updates the tags of the timing calibration databases (`v2r0` -> `v2r1`) to pick up the latest PMT laser corrections for Run 2. This is also necessary for the extractions of the next-level corrections for Run 2, as cosmics corrections are computed on top of the laser ones.

**Note**: This update depends on new `pmt_cable_delays_data.db` and `pmt_laser_timing_data.db` files, which should be packaged in a concurrent `icarus_data` release. These files have been produced by @jedori0228 and tested by both of us.

Asking reviews from people that are (willingly or not) involved in this business.

### DIFF
--- a/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
+++ b/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
@@ -40,7 +40,7 @@ icarusDB::PMTTimingCorrectionsProvider::PMTTimingCorrectionsProvider
 
 // -------------------------------------------------------------------------------
 
-uint64_t icarusDB::PMTTimingCorrectionsProvider::RunToDatabaseTimestamp( uint32_t run ) {
+uint64_t icarusDB::PMTTimingCorrectionsProvider::RunToDatabaseTimestamp( uint32_t run ) const{
 
    // Run number to timestamp used in the db
    // DBFolder.h only takes 19 digit (= timestamp in nano second),
--- a/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
+++ b/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
@@ -16,6 +16,7 @@
 
 // Database interface helpers
 #include "larevt/CalibrationDBI/Providers/DBFolder.h"
+#include "larevt/CalibrationDBI/IOVData/TimeStampDecoder.h"
 
 // C/C++ standard libraries
 #include <string>
@@ -68,6 +69,10 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t 
 
     bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
     mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
+    mf::LogTrace(fLogCategory)
+           << "Fetched IoV [ " << db.CachedStart().DBStamp() << " ; " << db.CachedEnd().DBStamp()
+           << " ] to cover t=" << RunToDatabaseTimestamp(run)
+           << " [=" << lariov::TimeStampDecoder::DecodeTimeStamp(RunToDatabaseTimestamp(run)).DBStamp() << "]";
 
     std::vector<unsigned int> channelList;
     if (int res = db.GetChannelList(channelList); res != 0) {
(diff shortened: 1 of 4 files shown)


mvicenzi:
The laset timing tables relared to this PR are now available in `icarus_data` v09_79_00. Thanks to @jedori0228 and @SFBayLaser .

jedori0228 (approved):
Looks good to me, we will soon have the corresponding tag for `icarus_data` (v09_79_00).

SFBayLaser (approved):
The two db files needed for this release are in icarus_data v09_79_00 which will get build and distributed with the next keep p release. As I read this PR, this goes to the develop branch so will get picked up.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -207,16 +208,17 @@ void icarusDB::PMTTimingCorrectionsProvider::readTimeCorrectionDatabase(const ar
...
-                  << std::endl; 
+            ss << key << " " 
+               << value.triggerCableDelay << "," 
+               << value.resetCableDelay << ", " 
+               << value.laserCableDelay << ", "
+               << value.cosmicsCorrections << ","
+               << std::endl; 
         }
+	mf::LogInfo(fLogCategory) << ss.str();
     }
```
PetrilloAtWork, line 222: I assume this change is to avoid the header of messagefacility on each single line. Two ways to proceed here.
You may want to create a "permanent" instance of the `mf::LogInfo` object:
```cpp
      mf::LogInfo log(fLogCategory);
      log << "Dump information from database "
        << "\nchannel, trigger cable delay, reset cable delay, laser corrections, muons corrections";
      for( auto const & [key, value] : fDatabaseTimingCorrections ){
         log << "\n" << key << " "
            << ", " << value.triggerCableDelay
            << ", " << value.resetCableDelay
            << ", " << value.laserCableDelay
            << ", " << value.cosmicsCorrections;
      }
```
This will enclose the whole message in a single header.
Or, you may want to just replace the `mf::LogInfo` with `mf::Verbatim` in the loop, or in both the look and the intro line, depending on the effect you want. Each of the [messagefacility streams](https://github.com/art-framework-suite/messagefacility/blob/23731098651df6a6087a2d4d697b4baadba5590d/messagefacility/MessageLogger/MessageLogger.h#L207-L219) comes with a version with the "decoration" and a version without it; `mf::LogVerbatim` is the twin of `mf::LogInfo`.

icaruscode/Timing/timing_corrections_tags.fcl
```diff
@@ -24,4 +24,14 @@ PMTtimingCorrectionsTags_Run2_August2023: {
...
 }
 
+# These are the standard tags for analyses on Run 1 and Run 2 data (as of September 2023)
+# These tagged versions of the databases only contain tables relevant for Run 1 and Run 2.
+# Notes: 
+#  - No cosmics corrections are available for Run 2
+PMTtimingCorrectionsTags_Run2_Sept2023: {
+  CablesTag:    "v2r1"  # tables for run>=0 (null), run>=8046, run>=9301, run>=9628 and run>=9773
+  LaserTag:     "v2r1"  # tables for run>=0 (null), run>=8046, run>=9301, run>=9628 and run>=9773
+  CosmicsTag:   "v1r0"  # tables for run>=0 (null) and run>=8046
```
PetrilloAtWork, line 34: This implies that we are applying on Run2 inconsistent laser and cosmic ray corrections.
Ideally we would have a null cosmic-ray-based correction for `run>=9301`.

PetrilloAtWork (commented):
I left a technical comment and a more important one.
While I would appreciate a statement on the latter, I believe that this is ultimately a known behaviour, and on that assumption I am not blocking the merge.
I believe that in principle every new laser-based correction update should come with a null cosmic-ray-based one for the same period, until the latter is reextracted.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: Yes, that's precisely what I was trying to achieve to easily read the printed table for better debugging. I'll follow your suggestion and use one of those.

icaruscode/Timing/timing_corrections_tags.fcl
mvicenzi: Yes, it would be better. It would also simplify my work in extracting the cosmics corrections for Run 2 by not having to subtract already applied cosmics corrections (or having to ask to specifically produce calibration ntuples with cosmics corrections disabled). I'll ask @jedori0228 how painful it would be to add it...

mvicenzi:
Following @PetrilloAtWork's comment, we decided to add a null table to avoid cosmics corrections meant for Run 1 being applied to Run 2. This updated file will need to make its way into `icarus_data` before merging this PR.

mvicenzi:
Imported several commits from the code review of #628 , in order to match the code between production and develop.

jedori0228:
`icarus_data` with tag `v09_79_01` now contains null table for Run>=9301 (table tag `v2r0`) for the cosmic timing table.

mmrosenberg:
icaruscode's icarus_data dependency has now been updated to v09_79_01

mmrosenberg:
trigger build

PetrilloAtWork (approved):
I am happy.

PR #623, opened 2023-08-29
Updated PMT channel mapping (20230829)

PetrilloAtWork:
This PR and the connected release of the updated channel mapping SQLite database in `icarus_data` `v09_78_00` reflect the completion (?) of the PMT cable swaps to have digitisers (and adders) serve contiguous PMT.

A third run period is introduced (the previous one lasting only one week). The database supports all the runs, and the code has (hard)coded the run period boundaries.
In addition, this PR contains a bug fix, a few diagnostics changes and the removal of an unused (I hope) FHiCL configuration.

*Note*: PostGreSQL database is not updated yet, and the schema for the change is not final yet. So **PostGreSQL support for runs after `10369`, August 23, 2023, is broken**.

This PR must be coupled with a release already including `icarus_data` `v09_78_00` or newer.
A `develop` PR will follow after this one passes the review.

This PR comes also with a few bug fixes on the previous code change.
Reviewers:
* @mvicenzi (who also delivered the database file)
* @SFBayLaser to confirm the configuration and version number changes
* @mmrosenberg also to confirm the version changes

PR #605, opened 2023-07-10
Removed trigger decoder V3 assertion

PetrilloAtWork:
The assertion ensured that bits that were assumed unused were indeed not used. The addition of adder information on some of those bits invalidated the check.

Reviewers: @jzettle is enjoying his rest and this is urgent, so it goes to run coordination and trigger expert.

PR #627, opened 2023-09-06
Updated PMT channel mapping (20230829) [develop]

PetrilloAtWork:
This is an almost-twin PR of #623: that was for production branch, this is for develop.
The code is the same (:cherries:) with just an adjacent-change conflict resolved.
*But* in order to make any sense, `develop` code must also need to be able to _decode_ the new runs with the new mapping, which is the reason I have also put in this PR the workaround on the trigger decoder at PR #605, which had not been proposed for `develop` yet.

Reviewers: mostly @mvicenzi, because he's going to use this branch anyway to verify that mapping.
It is reasonable, but not absolutely necessary, for this PR to wait until his checks are completed (anyway these changes are already in the _production_ branch...).


mvicenzi (approved):

mmrosenberg (approved):

mmrosenberg:
trigger build

PR #628, opened 2023-09-07
PMT timing corrections from local db + Run 1 cosmics corrections [SBN2023A]

mvicenzi:
This is twin PR of #614 meant for production.

It contains two things:
- Switch to using local database for PMT time corrections 
- Introduction of cosmics corrections. These can be disabled by setting the fcl parameter `CorrectCosmics` (from `icaruscode/Timing/timing_icarus.fcl`) to `false`.

Cosmics corrections require `icarus_data` 09_77_00 or newer, which should already be satisfied in the latest production patches.

### DIFF
--- a/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
+++ b/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
@@ -207,10 +207,10 @@ void icarusDB::PMTTimingCorrectionsProvider::readTimeCorrectionDatabase(const ar
     if( fVerbose ) {
 
         mf::LogInfo(fLogCategory) << "Dump information from database " << std::endl;
-        mf::LogInfo(fLogCategory) << "channel, trigger cable delay, reset cable delay, laser corrections, muons corrections" << std::endl;
-
+        mf::LogVerbatim(fLogCategory) << "channel, trigger cable delay, reset cable delay, laser corrections, muons corrections" << std::endl;
         for( auto const & [key, value] : fDatabaseTimingCorrections ){
-            mf::LogInfo(fLogCategory) << key << " " 
+            mf::LogVerbatim(fLogCategory) 
+                  << key << " " 
                   << value.triggerCableDelay << "," 
                   << value.resetCableDelay << ", " 
                   << value.laserCableDelay << ", "
--- a/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
+++ b/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
@@ -40,7 +40,7 @@ icarusDB::PMTTimingCorrectionsProvider::PMTTimingCorrectionsProvider
 
 // -------------------------------------------------------------------------------
 
-uint64_t icarusDB::PMTTimingCorrectionsProvider::RunToDatabaseTimestamp( uint32_t run ) {
+uint64_t icarusDB::PMTTimingCorrectionsProvider::RunToDatabaseTimestamp( uint32_t run ) const{
 
    // Run number to timestamp used in the db
    // DBFolder.h only takes 19 digit (= timestamp in nano second),
--- a/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
+++ b/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
@@ -16,6 +16,7 @@
 
 // Database interface helpers
 #include "larevt/CalibrationDBI/Providers/DBFolder.h"
+#include "larevt/CalibrationDBI/IOVData/TimeStampDecoder.h"
 
 // C/C++ standard libraries
 #include <string>
@@ -68,6 +69,10 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t 
 
     bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
     mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
+    mf::LogTrace(fLogCategory)
+           << "Fetched IoV [ " << db.CachedStart().DBStamp() << " ; " << db.CachedEnd().DBStamp()
+           << " ] to cover t=" << RunToDatabaseTimestamp(run)
+           << " [=" << lariov::TimeStampDecoder::DecodeTimeStamp(RunToDatabaseTimestamp(run)).DBStamp() << "]";
 
     std::vector<unsigned int> channelList;
     if (int res = db.GetChannelList(channelList); res != 0) {
@@ -126,6 +131,10 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadLaserCorrections( uint32_t run 
 
     bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
     mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
+    mf::LogTrace(fLogCategory)
+           << "Fetched IoV [ " << db.CachedStart().DBStamp() << " ; " << db.CachedEnd().DBStamp()
+           << " ] to cover t=" << RunToDatabaseTimestamp(run)
+           << " [=" << lariov::TimeStampDecoder::DecodeTimeStamp(RunToDatabaseTimestamp(run)).DBStamp() << "]";
 
     std::vector<unsigned int> channelList;
     if (int res = db.GetChannelList(channelList); res != 0) {
@@ -162,6 +171,10 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadCosmicsCorrections( uint32_t ru
 
     bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
     mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
+    mf::LogTrace(fLogCategory)
+           << "Fetched IoV [ " << db.CachedStart().DBStamp() << " ; " << db.CachedEnd().DBStamp()
+           << " ] to cover t=" << RunToDatabaseTimestamp(run)
+           << " [=" << lariov::TimeStampDecoder::DecodeTimeStamp(RunToDatabaseTimestamp(run)).DBStamp() << "]";
 
     std::vector<unsigned int> channelList;
     if (int res = db.GetChannelList(channelList); res != 0) {
--- a/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
+++ b/icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
@@ -27,15 +27,15 @@ icarusDB::PMTTimingCorrectionsProvider::PMTTimingCorrectionsProvider
     (const fhicl::ParameterSet& pset) 
     : fVerbose{ pset.get<bool>("Verbose", false) }
     , fLogCategory{ pset.get<std::string>("LogCategory", "PMTTimingCorrection") }
-    , fTags{ pset.get<fhicl::ParameterSet>("CorrectionsTags") }
     { 
-	fCablesTag  = fTags.get<std::string>("CablesTag", "v1r0");
-	fLaserTag   = fTags.get<std::string>("LaserTag", "v1r0");
-	fCosmicsTag = fTags.get<std::string>("CosmicsTag", "v1r0");
+        fhicl::ParameterSet const tags{ pset.get<fhicl::ParameterSet>("CorrectionsTags") };
+	fCablesTag  = tags.get<std::string>("CablesTag", "v1r0");
+	fLaserTag   = tags.get<std::string>("LaserTag", "v1r0");
+	fCosmicsTag = tags.get<std::string>("CosmicsTag", "v1r0");
 	if( fVerbose ) mf::LogInfo(fLogCategory) << "Database tags for timing corrections:\n"
 						 << "Cables corrections  " << fCablesTag << "\n"  
 						 << "Laser corrections   " << fLaserTag  << "\n"
-						 << "Cosmics corrections " << fCosmicsTag << std::endl;
+						 << "Cosmics corrections " << fCosmicsTag;
     }
 
 // -------------------------------------------------------------------------------
@@ -51,7 +51,7 @@ uint64_t icarusDB::PMTTimingCorrectionsProvider::RunToDatabaseTimestamp( uint32_
    uint64_t timestamp = runNum+1000000000;
    timestamp *= 1000000000;
 
-   if( fVerbose ) mf::LogInfo(fLogCategory) << "Run " << runNum << " corrections from DB timestamp " << timestamp << std::endl;
+   if( fVerbose ) mf::LogInfo(fLogCategory) << "Run " << runNum << " corrections from DB timestamp " << timestamp;
    
    return timestamp;
 }
@@ -72,11 +72,11 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t 
     std::vector<unsigned int> channelList;
     if (int res = db.GetChannelList(channelList); res != 0) {
       throw cet::exception
-        ( "PMTTimingCorrectionsProvider: GetChannelList() returned " + std::to_string(res) + " on run " + std::to_string(run) + " query in " + dbname);
+        ( "PMTTimingCorrectionsProvider" ) << "GetChannelList() returned " << res << " on run " << run << " query in " << dbname << "\n";
     }
     
     if (channelList.empty()) {
-      throw cet::exception("PMTTimingCorrectionsProvider: got an empty channel list for run " + std::to_string(run) + " in " + dbname);
+      throw cet::exception("PMTTimingCorrectionsProvider") << "Got an empty channel list for run " << run << " in " << dbname << "\n";
     }
 
     for( auto channel : channelList ) {
@@ -84,17 +84,17 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t 
         // PPS reset correction
         double reset_distribution_delay = 0;
         int error  = db.GetNamedChannelData( channel, "reset_distribution_delay", reset_distribution_delay );
-        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'reset_distribution_delay' on table " + dbname );
+        if( error ) throw cet::exception("PMTTimingCorrectionsProvider") << "Encountered error (code " << error << ") while trying to access 'reset_distribution_delay' on table " << dbname << "\n";
 
         // Trigger cable delay
         double trigger_reference_delay = 0;
         error  = db.GetNamedChannelData( channel, "trigger_reference_delay", trigger_reference_delay );
-        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'trigger_reference_delay' on table " + dbname );
+        if( error ) throw cet::exception( "PMTTimingCorrectionsProvider" ) << Encountered error (code " << error << ") while trying to access 'trigger_reference_delay' on table " << dbname << "\n";
 
         // Phase correction
         double phase_correction = 0;
 	error = db.GetNamedChannelData( channel, "phase_correction", phase_correction );
-        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'phase_correction' on table " + dbname );
+        if( error ) throw cet::exception( "PMTTimingCorrectionsProvider" ) << "Encountered error (code " << error <<  ") while trying to access 'phase_correction' on table " << dbname << "\n";
    
         /// This is the delay due to the cables connecting the 'global' trigger crate FPGA to the spare channel of the first digitizer in each VME crates. 
         /// The phase correction is an additional fudge factor 
@@ -130,11 +130,11 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadLaserCorrections( uint32_t run 
     std::vector<unsigned int> channelList;
     if (int res = db.GetChannelList(channelList); res != 0) {
       throw cet::exception
-        ( "PMTTimingCorrectionsProvider: GetChannelList() returned " + std::to_string(res) + " on run " + std::to_string(run) + " query in " + dbname);
+        ( "PMTTimingCorrectionsProvider" ) << "GetChannelList() returned " << res << " on run " << run << " query in " << dbname << "\n";
     }
     
     if (channelList.empty()) {
-      throw cet::exception("PMTTimingCorrectionsProvider: got an empty channel list for run " + std::to_string(run) + " in " + dbname);
+      throw cet::exception("PMTTimingCorrectionsProvider") << "got an empty channel list for run " << run << " in " << dbname << "\n";
     }
 
     for( auto channel : channelList ) {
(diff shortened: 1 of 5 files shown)


icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
-            /// pmt_laser_delay: delay from the Electron Transit time inside the PMT 
-            /// and the PMT signal cable 
-            /// corrections are saved in ns, but icaruscode wants us
-            /// Corrections are additive! 
-            fDatabaseTimingCorrections[channel_id].laserCableDelay = -t_signal/1000.; 
-
-            releaseTuple(tuple);
-        }
+    bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
+    mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
```
PetrilloAtWork, line 128: Same question as for the reset delay table above.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,107 @@
...
-            
-            // PPS reset correction
-            double reset_distribution_delay = getDoubleValue( tuple, 11, &error );
-            if( error ) throw std::runtime_error( "Encountered error '" + std::to_string(error) + "' while trying to access 'reset_distribution_delay' on table " + name );
-
-            // Trigger cable delay
-            double trigger_reference_delay = getDoubleValue( tuple, 10, &error );
-            if( error ) throw std::runtime_error( "Encountered error while trying to access 'trigger_reference_delay' on table " + name );
+    std::vector<unsigned int> channelList;
+    if (int res = db.GetChannelList(channelList); res != 0) {
```
PetrilloAtWork, line 73: C++17 is supposed to understand that if there is a variable definition and no explicit condition, the condition should be that the conversion of the defined variable to `bool` be `true`, that would be `bool(res) == true` here, so equivalent to `res != 0`:
```suggestion
    if (int res = db.GetChannelList(channelList)) {
```
Note that I am not necessarily recommending this syntax, which is somehow less readable: I let it to your judgement (I might end up leaving it as is myself).

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,107 @@
...
+        
+        // PPS reset correction
+        double reset_distribution_delay = 0;
+        int error  = db.GetNamedChannelData( channel, "reset_distribution_delay", reset_distribution_delay );
+        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'reset_distribution_delay' on table " + dbname );
+
+        // Trigger cable delay
+        double trigger_reference_delay = 0;
+        error  = db.GetNamedChannelData( channel, "trigger_reference_delay", trigger_reference_delay );
+        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'trigger_reference_delay' on table " + dbname );
```
PetrilloAtWork, line 92:
```suggestion
        if( error ) throw cet::exception( "PMTTimingCorrectionsProvider" ) << Encountered error (code " << error << ") while trying to access 'trigger_reference_delay' on table " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,107 @@
...
+
+        // Phase correction
+        double phase_correction = 0;
+	error = db.GetNamedChannelData( channel, "phase_correction", phase_correction );
+        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'phase_correction' on table " + dbname );
```
PetrilloAtWork, line 97:
```suggestion
        if( error ) throw cet::exception( "PMTTimingCorrectionsProvider" ) << "Encountered error (code " << error <<  ") while trying to access 'phase_correction' on table " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
-
-            releaseTuple(tuple);
-        }
+    bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
+    mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
 
+    std::vector<unsigned int> channelList;
+    if (int res = db.GetChannelList(channelList); res != 0) {
+      throw cet::exception
+        ( "PMTTimingCorrectionsProvider: GetChannelList() returned " + std::to_string(res) + " on run " + std::to_string(run) + " query in " + dbname);
```
PetrilloAtWork, line 133:
```suggestion
        ( "PMTTimingCorrectionsProvider" ) << "GetChannelList() returned " << res << " on run " << run << " query in " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
+    }
+    
+    if (channelList.empty()) {
+      throw cet::exception("PMTTimingCorrectionsProvider: got an empty channel list for run " + std::to_string(run) + " in " + dbname);
```
PetrilloAtWork, line 137:
```suggestion
      throw cet::exception("PMTTimingCorrectionsProvider") << "got an empty channel list for run " << run << " in " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
     }
 
+    for( auto channel : channelList ) {
+        
+        // Laser correction
+        double t_signal = 0;
+        int error  = db.GetNamedChannelData( channel, "t_signal", t_signal );
+        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 't_signal' on table " + dbname );
```
PetrilloAtWork, line 145:
```suggestion
        if( error ) throw cet::exception( "PMTTimingCorrectionsProvider" ) << "Encountered error (code " << error << ") while trying to access 't_signal' on table " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
-    int error = ConnectToDataset( name, run, dataset );
+void icarusDB::PMTTimingCorrectionsProvider::ReadCosmicsCorrections( uint32_t run ) { 
 
-    if (error) throw(std::exception());
+    const std::string dbname("pmt_cosmics_timing_data");
+    lariov::DBFolder db(dbname, "", "", fCosmicsTag, true, false);
 
-    for( int row=4; row < getNtuples(dataset); row++ ) {
+    bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
+    mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
```
PetrilloAtWork, line 164: As above.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
-
-            /// pmt_laser_delay: delay from the Electron Transit time inside the PMT 
-            /// and the PMT signal cable 
-            /// corrections are saved in ns, but icaruscode wants us
-            /// Corrections are additive! 
-            fDatabaseTimingCorrections[channel_id].cosmicsCorrections = t_signal/1000.; 
+    std::vector<unsigned int> channelList;
+    if (int res = db.GetChannelList(channelList); res != 0) {
+      throw cet::exception
+        ( "PMTTimingCorrectionsProvider: GetChannelList() returned " + std::to_string(res) + " on run " + std::to_string(run) + " query in " + dbname);
```
PetrilloAtWork, line 169:
```suggestion
        ( "PMTTimingCorrectionsProvider" ) << "GetChannelList() returned " << res << " on run " << run << " query in " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
+    }
 
-            releaseTuple(tuple);
-        }
+    if (channelList.empty()) {
+      throw cet::exception("PMTTimingCorrectionsProvider: got an empty channel list for run " + std::to_string(run) + " in " + dbname);
```
PetrilloAtWork, line 173:
```suggestion
      throw cet::exception("PMTTimingCorrectionsProvider") << "Got an empty channel list for run " << run << " in " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -128,72 +121,73 @@ void icarusDB::PMTTimingCorrectionsProvider::ReadPMTCablesCorrections( uint32_t
...
+    }
 
+    for( auto channel : channelList ) {
+        
+        // Cosmics correction
+ 	double mean_residual_ns = 0;
+	int error = db.GetNamedChannelData( channel, "mean_residual_ns", mean_residual_ns );
+        if( error ) throw cet::exception( "Encountered error (code " + std::to_string(error) + ") while trying to access 'mean_residual_ns' on table " + dbname );
```
PetrilloAtWork, line 181:
```suggestion
        if( error ) throw cet::exception( "PMTTimingCorrectionsProvider" ) << "Encountered error (code " << error << ") while trying to access 'mean_residual_ns' on table " << dbname << "\n";
```

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -207,15 +201,13 @@ void icarusDB::PMTTimingCorrectionsProvider::readTimeCorrectionDatabase(const ar
...
     ReadLaserCorrections(run.id().run());
-
-    //ReadCosmicsCorrections(run.id().run());
+    ReadCosmicsCorrections(run.id().run());
 
     if( fVerbose ) {
 
         mf::LogInfo(fLogCategory) << "Dump information from database " << std::endl;
-        mf::LogInfo(fLogCategory) << "channel, trigger cable delay, reset cable delay, laser corrections, muons correctsions" << std::endl;
+        mf::LogInfo(fLogCategory) << "channel, trigger cable delay, reset cable delay, laser corrections, muons corrections" << std::endl;
```
PetrilloAtWork, line 210: You may consider here to merge the two messages, 
```cpp
      mf::LogInfo(fLogCategory) << "Dump information from database"
         << "\nchannel, trigger cable delay, reset cable delay, laser corrections, muon corrections";
```
and possibly to treat the message as described before for the one similar to this.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,107 @@
...
+    : fVerbose{ pset.get<bool>("Verbose", false) }
     , fLogCategory{ pset.get<std::string>("LogCategory", "PMTTimingCorrection") }
-    , fClocksData{ art::ServiceHandle<detinfo::DetectorClocksService const>()->DataForJob() }
-    { }
-
+    , fTags{ pset.get<fhicl::ParameterSet>("CorrectionsTags") }
+    { 
+	fCablesTag  = fTags.get<std::string>("CablesTag", "v1r0");
+	fLaserTag   = fTags.get<std::string>("LaserTag", "v1r0");
+	fCosmicsTag = fTags.get<std::string>("CosmicsTag", "v1r0");
```
PetrilloAtWork, line 34: The `CorrectionsTags` (shouldn't it be `CorrectionTags` in English?) table does not need to be stored, since its information is fully extracted into other data members:
```suggestion
    { 
        fhicl::ParameterSet const tags{ pset.get<fhicl::ParameterSet>("CorrectionsTags") };
	fCablesTag  = tags.get<std::string>("CablesTag", "v1r0");
	fLaserTag   = tags.get<std::string>("LaserTag", "v1r0");
	fCosmicsTag = tags.get<std::string>("CosmicsTag", "v1r0");
```
Chances are that you can even refer to the parameters directly:
```suggestion
    { 
	fCablesTag  = tags.get<std::string>("CorrectionsTags.CablesTag", "v1r0");
	fLaserTag   = tags.get<std::string>("CorrectionsTags.LaserTag", "v1r0");
	fCosmicsTag = tags.get<std::string>("CorrectionsTags.CosmicsTag", "v1r0");
```
Some of `fhicl::ParameterSet` methods support this, I don't remember in which circumstances though.

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -19,32 +19,24 @@
...
 #include "cetlib_except/exception.h"
 
-#include "lardata/DetectorInfoServices/DetectorClocksService.h"
-
+// Local
 #include "icaruscode/Timing/PMTTimingCorrections.h"
 
 // Database interface helpers
-#include "wda.h"
+#include "larevt/CalibrationDBI/Providers/DBFolder.h"
```
PetrilloAtWork, line 26: I don't see anything in this header that needs `DBFolder`. It is (correctly) included into the source file already.
```suggestion
```

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -54,74 +46,62 @@ namespace icarusDB{ class PMTTimingCorrectionsProvider; }
...
  * Configuration parameters
  * -------------------------
- * 
- * 
- * 
- * Multithreading
- * ---------------
- * 
- * 
+ * * `Tag` (default: `false`): Tag for database versioning
```
PetrilloAtWork, line 56: Need fix.
```suggestion
 * * `CorrectionsTags`: tags to select the correction versions:
 *     * `CablesTag` (default: `v1r0`): correction for cable delay.
 *     * `LaserTag` (default: `v1r0`): first order PMT time correction, from laser data.
 *     * `CosmicsTag` (default: `v1r0`): second order PMT time correction, from cosmic rays.
```

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -54,74 +46,62 @@ namespace icarusDB{ class PMTTimingCorrectionsProvider; }
...
-
-        unsigned int fTimeout;
-
-        bool fCorrectCablesDelay;
+        static constexpr PMTTimeCorrectionsDB CorrectionDefaults {}; ///< Default values
 
         bool fVerbose = false; ///< Whether to print the configuration we read.
-  
         std::string fLogCategory; ///< Category tag for messages.
+	fhicl::ParameterSet fTags; ///< List of database tags
```
PetrilloAtWork, line 97: This data member is not needed (see above).
```suggestion
```

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -130,8 +110,8 @@ class icarusDB::PMTTimingCorrectionsProvider : public PMTTimingCorrections {
                 return (it == fDatabaseTimingCorrections.end())? CorrectionDefaults: it->second;
             }
 
-        int ConnectToDataset(const std::string& name, 
-            uint32_t run, Dataset& dataset ) const;
+	/// Convert run number to internal database
+	uint64_t RunToDatabaseTimestamp(uint32_t run);
```
PetrilloAtWork, line 114: Please mark this function as `const`.

icaruscode/Timing/timing_icarus.fcl
```diff
@@ -1,11 +1,13 @@
...
 
 icarus_pmttimingservice:
 {
-    # service name: IPMTTimingCorrectionService
+    # service name:   IPMTTimingCorrectionService
     service_provider: PMTTimingCorrectionService
-    DatabaseURL:   "https://dbdata0vm.fnal.gov:9443/icarus_con_prod/app/data?"
-    Timeout:   1000
+    CorrectionsTags:  @local::PMTtimingCorrectionsTags_Run2_August2023
+    Verbose:          false
```
PetrilloAtWork, line 10: Have mercy: set this to `false` for normal usage.

icaruscode/Timing/timing_corrections_tags.fcl
```diff
@@ -0,0 +1,27 @@
...
+
+# These are the standard tags for analyses on Run 1 and Run 2 data (as of August 2023)
+# These tagged versions of the databases only contain tables relevant for Run 1 and Run 2.
+# Notes: 
+#  - No cosmics corrections are available for Run 2
+#  - Laser and cable corrections do not cover the entire Run 2 period
+PMTtimingCorrectionsTags_Run2_August2023: {
+  CablesTag:    "v2r0"  # tables for run>=0 (null), run>=8046 and run>=9773
+  LaserTag:     "v2r0"  # tables for run>=0 (null), run>=8046 and run>=9773
+  CosmicsTag:   "v1r0"  # tables for run>=0 (null) and run>=8046
```
PetrilloAtWork, line 24: As discussed on aside, a null cosmic-ray-based correction table for `run >= 9773` should be useful.

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -19,32 +19,24 @@
 #include "fhiclcpp/types/Sequence.h"
 #include "cetlib_except/exception.h"
```
PetrilloAtWork, line 20: Once is enough (actually, even too much since the header does not use exceptions, but it's not bad practice to include this since in some sense thrown exceptions are part of the service interface).
```suggestion
```
Also, regrettably we are not using FHiCL validation here, so the `fhiclcpp/types` headers are not needed, while `fhiclcpp/ParameterSet.h` is.
I think `ServiceHandle.h` is also superfluous.

icaruscode/Timing/PMTTimingCorrectionsProvider.h
```diff
@@ -19,32 +19,24 @@
...
 #include "icaruscode/Timing/PMTTimingCorrections.h"
 
 // Database interface helpers
-#include "wda.h"
+#include "larevt/CalibrationDBI/Providers/DBFolder.h"
 
 // C/C++ standard libraries
-#include <memory> // std::unique_ptr<>
-#include <optional>
 #include <string>
```
PetrilloAtWork, line 29: We also want, and always did,
```cpp
#include <map>
```
And to be pedantic, which I always am, also `<cstdint>` for `uint64_t` and similar. Although to be even more pedantic, it would be C header, `<stdint.h>`, since we are not using the namespace (`std::uint64_t`).

PetrilloAtWork (changes requested):
Some general comments:
1. backward compatibility: how is the old code going to behave with the database having more than one period? my glance at the old code suggests that there is a fundamental change in the database structure and the old code would not be able to parse the new database at all. For the timing corrections this would to be a ~major problem~ showstopper, since it means that the whole ICARUS code before this PR won't be able to process any data (at least as long as the timing correction service is configured in the job).
2. Exception messages should be addressed. It does affect the workflow options (in particular, exceptions from particular categories might be ignored, so it has value for all the exceptions from a single source to have the same category)

mvicenzi:
> Some general comments: 
> 1. backward compatibility: how is the old code going to behave with the database having more than one period? my glance at the old code suggests that there is a fundamental change in the database structure and the old code would not be able to parse the new database at all. For the timing corrections this would to be a ~major problem~ showstopper, since it means that the whole ICARUS code before this PR won't be able to process any data (at least as long as the timing correction service is configured in the job).

Regarding backward compatibility, I see no issues here:

1. The old code reads the time corrections from the online PostGres database, which was already time-sensitive and whose structure has not changed. The old software used the run number in its query to select the appropriate correction tables. This behavior still works, so the old software is still able to query and use corrections. No changes were made in the database structure, nor in the way it selects the run periods (which is built-in).
Moreover, since we are also in parallel updating the PostGres database by adding new tables for more recent periods, the old software would in principle be able to get the right corrections for both old and recent runs.
2. The above applies for what concerns the "cable" and "laser" corrections. The old software doesn't attempt to access the "cosmics" corrections since the database itself (both PostGres and Sqlite) didn't exist at the time, but this doesn't cause any backward compatibility issue.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: Thank you for the suggestions. I was unsure on how to read tags from a table, so I went and saved the table as well. I will simplify the code as you suggests.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
```diff
@@ -10,114 +10,107 @@
...
     // and reset distribution
-    const std::string name("pmt_cables_delays_data");
-    Dataset dataset;
-    int error = ConnectToDataset( name, run, dataset );
+    const std::string dbname("pmt_cables_delays_data");
+    lariov::DBFolder db(dbname, "", "", fCablesTag, true, false);
 
-    if (error) throw(std::exception());
+    bool ret = db.UpdateData( RunToDatabaseTimestamp(run) ); // select table based on run number   
+    mf::LogDebug(fLogCategory) << dbname + " corrections" << (ret? "": " not") << " updated for run " << run;
```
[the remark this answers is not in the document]
mvicenzi: I was never able to trigger a `return false` from `db.UpdateData`, even when running over multiple files from the same run (which is when you would expect the data to be already present). My decision was then to simply log the event in case issues ever arise.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: I'd rather keep the expanded syntax. It would be embarassing if future me ended up not understading his own code.

icaruscode/Timing/PMTTimingCorrectionsProvider.cxx
mvicenzi: I will do the same as in #626

icaruscode/Timing/timing_corrections_tags.fcl
mvicenzi: Yes, this will be added with #626 and eventually ported to production as well

PetrilloAtWork (approved):
My questions and concerns were fully addressed.
Thank you.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_72_00 SBNSoftware/sbncode@v09_72_00_05 SBNSoftware/sbnanaobj@v09_20_06_02 SBNSoftware/sbndaq-artdaq-core@v1_06_00of0 SBNSoftware/sbnobj@v09_16_00_01 SBNSoftware/icarus_signal_processing@v09_63_00_02 SBNSoftware/icarusalg@v09_72_00_04 SBNSoftware/icarusutil@v09_66_02

PR #622, opened 2023-08-23
Run-period-dependent channel mapping [SBN2023A]

PetrilloAtWork:
This is the production version (i.e. the important version) of PR #620.
They contain the very same commits. :cherries: 

It is imperative that this goes into release as quick as possible once it's ready ~(and now it is _NOT_ ready yet)~ so that production can keep up.

The cable swapping which this request reflects is expected to happen momentarily (in the coming hours), after which we can update the hard-coded run number boundary of the change, and when _that_ commit is also in, dris PR is ready to go.
~Until then, it's draft.~

PR #630, opened 2023-09-12
Update PostGres tool for new PMT mapping

mvicenzi:
As [promised](https://github.com/SBNSoftware/icaruscode/pull/620#issuecomment-1690226045) to @SFBayLaser in #622 , this PR updates `PostGresChannelMap_tool` to handle the time-period awareness now required by the PMT channel mapping. 

The FNAL DB team has changed the structure of the `pmt_placements` table, adding a timestamp (`period_active`) which can be used to select the time period. Since the database itself was changed, this is a breaking change: the old software is no longer able to read the PMT table correctly. However, it is important to note that PostGres is no longer actively used, and it is kept only as a backup option.

To read off the new table structure:
- The query for the database was changed, appending a SQL `where` clause: `"&w=period_active:eq:"+fCurrentPMTTimestamp`
- For unknown reasons, the new PMT table has a different column ordering: the hardcoded column numbers were changed to match the right data.

### DIFF
--- a/icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
+++ b/icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
@@ -137,15 +137,14 @@ bool ChannelMapPostGres::SelectPeriod(RunPeriod period) {
   
   if (fCurrentPMTTimestamp == newPMTTimestamp) {
     mf::LogDebug("ChannelMapPostGres")
-      << "Period #" << iPeriod << " already selected";
+      << "Period " << newPMTTimestamp << " already selected";
   }
   else if ( !fCurrentPMTTimestamp.empty() ) {
-    mf::LogDebug("ChannelMapPostGres") << "Switched from period #"
-      << std::find(PMTTimestampSet.begin(), PMTTimestampSet.end(), fCurrentPMTTimestamp) 
-      << " to #" << iPeriod;
+    mf::LogDebug("ChannelMapPostGres") << "Switched from period "
+      << fCurrentPMTTimestamp << " to " << newPMTTimestamp;
   }
   else {
-    mf::LogDebug("ChannelMapPostGres") << "Switching to period #" << iPeriod;
+    mf::LogDebug("ChannelMapPostGres") << "Switching to period " << newPMTTimestamp;
   }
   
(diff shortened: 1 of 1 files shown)


mmrosenberg:
trigger build

icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
```diff
@@ -165,23 +133,24 @@ ChannelMapPostGres::ChannelMapPostGres(fhicl::ParameterSet const &pset)
...
-  if (fCurrentTable == newSet) {
+  if (fCurrentPMTTimestamp == newPMTTimestamp) {
     mf::LogDebug("ChannelMapPostGres")
       << "Period #" << iPeriod << " already selected";
   }
-  else if (fCurrentTable) {
+  else if ( !fCurrentPMTTimestamp.empty() ) {
     mf::LogDebug("ChannelMapPostGres") << "Switched from period #"
-      << (fCurrentTable - &(TableNameSets[0])) << " to #" << iPeriod;
+      << std::find(PMTTimestampSet.begin(), PMTTimestampSet.end(), fCurrentPMTTimestamp)
```
PetrilloAtWork, line 144: Doesn't
```suggestion
      << std::find(PMTTimestampSet.begin(), PMTTimestampSet.end(), fCurrentPMTTimestamp) 
```
return an iterator? what does this print?
I suggest that the period "time stamps" be printed instead of the number. The SQLite did not do that just because there was nothing easily available to print, while here we do have it.

Same comment for the previous branch if this `if`.

PetrilloAtWork (approved):
I left a minor comment on the debug output.
While I would have preferred something more sortable than `23aug2023` (which comes after `22sep2023` and before `27feb1988`), I can live with it.

icaruscode/Decode/ChannelMapping/ChannelMapPostGres_tool.cc
mvicenzi: I agree it makes much more sense to directly print the timestamp strings themselves. New commit deals with that.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #632, opened 2023-09-20
Update to the CRT calibration Ntuple

francescopoppi:
This PR includes a new feature to the CRT Hit calibration ntuple.
Originally only the sum of the PEs in one CRT Hit was stored in the HitTree. This PR includes 3 new branches (arrays of size 32) to the HitTree.
1st Branch is PEs, for the Top CRT (for which all the 32 channels of the FEB are recorded) each index of the PEs array correspond to the channel of the FEB, the entry is the corresponding calibrated PE value. The Side CRTs only store information of the channels that are actively part of the reconstructed CRT Hit, this is due to how the Side CRT works (multiple channels from different FEB participate in a hit). For the Side CRT you loose correspondance between channel and index, this is nonetheless fixed with the inclusion of a 2nd branch (array): Chans.
2nd Branch is Chans: for the Top CRT, since it is not needed, it is filled with default value (-1). For the Side CRT each index correspond to the same index of PEs and Macs branch. Stores information of the channel number.
3rd Branch is Macs:  for the Top CRT, since it is not needed, it is filled with default value (-1). For the Side CRT each index correspond to the same index of PEs and Chans branch. Stores information of the Mac that is associated with the i-th channel, i-th pe.

### DIFF
--- a/icaruscode/CRT/CRTDataAnalysis_module.cc
+++ b/icaruscode/CRT/CRTDataAnalysis_module.cc
@@ -218,6 +218,8 @@ namespace crt {
     float    fZErrHit; ///< stat error of CRT hit reco Z (cm)
     uint64_t    fT0Hit; ///< hit time w.r.t. PPS
     Long64_t    fT1Hit; ///< hit time w.r.t. global trigger
+    float     fHitPE[32];
+    int       fHitMac[32];
     int       fHitReg; ///< region code of CRT hit
     int       fHitSubSys;
     int       fNHit; ///< number of CRT hits for this event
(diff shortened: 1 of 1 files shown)


aheggest (approved):
looks good to me! thanks for accounting for the side CRT reco

francescopoppi:
I included few more suggestions I received from @tboone0042

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #633, opened 2023-09-22
Omnibus pull request mainly aimed at signal processing 

SFBayLaser:
Several topics to address with this PR:
1) in conjunction with icarus_signal_process include new options for waveform smoothing aiming to address the observed low frequency oscillation in data. The interface for this is a bit of a kludge at the moment but will allow more significant testing to proceed. 
2) Included an analysis tool to output a tuple allowing studies of ROI and Hit finding efficiency
3) Updates to the wirecell configuration files to address issues observed with time offsets in waveforms, at this point should have deconvolved waveforms aligned with raw waveforms from simulation and the SimChannel truth information. Also added output from the new "DepoFluxWriter" to get SimChannel information from that module which includes the "origTrackID" field needed by the ML effort.
4) Included a test version of the gauss hit finder which will use ChannelROIs as input instead of Wire objects. Ultimately, if tests successful, will aim to push the updates back to larsoft, so this is for testing only at this point.


SFBayLaser:
Note that this will need the latest version of icarus_data which has updated field response files

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mrmooney (approved):
Hi - this looks good to me.  However, please make sure we are set to use the data-driven field response for signal simulation in MC and for the 2D deconvolution in both data and MC.  I'm not sure if that is included in this, but if not, please work to include this in the next release.

gputnam (approved):
Approved after some discussion over slack. The main thing is to ensure that the icarus::Hit object does not make it into any production.

mmrosenberg:
trigger build

PR #634, opened 2023-09-25
Added cosmic distance variable

icaromx:
This PR changes one of the fcl files to enable that we save a variable that calculates the shortest distance from a shower to the nearest cosmic onto cafs. Currently, this variable goes unfilled in cafs but is calculated at the stage1 level, so I made changes to the cafemake_def.fcl file in order to start storing these values.


mmrosenberg:
trigger build

PR #635, opened 2023-09-26
Quick fix of typo in configuration file + module to convert icarus::Hits to recob::Hits

SFBayLaser:
The first breaks the 2D deconvolution so need this fix as patch.
Second will allow studies by others of test gauss hit finder which uses ChannelROIs as input.


SFBayLaser:
trigger build

mmrosenberg:
trigger build

PR #636, opened 2023-09-30
Add a CRT Hit flag to categorize side CRT Z position rectonstruction

aheggest:
This PR adds a CRT Hit flag to help identify some cases where the Z-position reconstruction for side CRT hits isnt so good.
The Side CRT Hit are made of coincidences between inner and outer layers, with two layers being readout on opposite ends by two different FEBs (For East and West Side CRT walls *only*, North and South Side CRT walls use cut modules and readout on a single end). 
In some cases, the Side CRT Hit Z position can be reconstructed off the bounds of the module wall, due to the FEB readouts on opposite ends being too far apart in time (T0 on inner and outer layers being > 54ns apart). There are some other cases where the Side CRT Z-Position is defaulted to the center of the module, which is also flagged here. 
This PR is able to flag the badly reconstructed Z-positions (flag = 1,2,3) so the analyzer can choose whether or not to remove those CRT Hits entirely,  which one might want to do if the analysis is CRT position dependent.
We can use the CRT light signal amplitude (PE) to help assign a better Z position for these flagged CRT Hits, but this will have to wait for a new PR.

### DIFF
--- a/icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
+++ b/icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
@@ -342,13 +342,13 @@ vector<pair<sbn::crt::CRTHit, vector<int>>> CRTHitRecoAlg::CreateCRTHits(
 // Function to make filling a CRTHit a bit faster
 sbn::crt::CRTHit CRTHitRecoAlg::FillCRTHit(
     vector<uint8_t> tfeb_id, map<uint8_t, vector<pair<int, float>>> tpesmap,
-    float peshit, uint64_t time0, Long64_t time1, int plane, double x,
+    float peshit, double flag, uint64_t time0, Long64_t time1, int plane, double x,
     double ex, double y, double ey, double z, double ez, string tagger) {
   CRTHit crtHit;
   crtHit.feb_id = tfeb_id;
   crtHit.pesmap = tpesmap;
   crtHit.peshit = peshit;
-  crtHit.ts0_s_corr = time0 / 1'000'000'000;
+  crtHit.ts0_s_corr = flag;
   crtHit.ts0_ns = time0 % 1'000'000'000;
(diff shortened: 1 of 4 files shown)


tboone0042 (approved):
All looks good to me, thanks for doing this Anna! I appreciate the added comments and cleaning up the GT stuff for Side CRT!

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
```diff
@@ -441,16 +322,17 @@ vector<pair<sbn::crt::CRTHit, vector<int>>> CRTHitRecoAlg::CreateCRTHits(
...
+      << " , nHitM: " << nHitM << '\n'
+      << "  nMisC: " << nMissC << " , nMisD: " << nMissD
+      << " , nMisM: " << nMissM << '\n'
+      << " CRT Hits by region" << '\n';
     auto cts = regCounts.begin();
-    mf::LogInfo("CRT") << " CRT Hits by region" << '\n';
     while (cts != regCounts.end()) {
-      mf::LogInfo("CRT") << "reg: " << (*cts).first
-                         << " , hits: " << (*cts).second << '\n';
+      std::cout  << "reg: " << (*cts).first
```
francescopoppi, line 334: Are you sure you want this as a std::cout and not as a Log Message?
Just wanna make sure it does not overflow the console with print messages

icaruscode/CRT/CRTUtils/CRTHitRecoAlg.cc
```diff
@@ -441,16 +322,17 @@ vector<pair<sbn::crt::CRTHit, vector<int>>> CRTHitRecoAlg::CreateCRTHits(
...
+		 << " , hits: " << (*cts).second << '\n';
```
francescopoppi, line 335: same as above

PR #638, opened 2023-10-12
This commit fixes a bug in the CRT-PMT matching analyzer module. Prev…

francescopoppi:
…iously there was the requirment for a certain number of opHits to be above threshold to consider the flash valid. The bug did not initialze some vectors when the nopHitAboveThreshold was lower then the requirments. Since in the future threshold will change, I removed the strong skipping-the-flash condition and I added the number of opHits above threhsold as a variabile for the analyzers to cut on.

### DIFF
--- a/icaruscode/CRT/CRTPMTMatchingAna_module.cc
+++ b/icaruscode/CRT/CRTPMTMatchingAna_module.cc
@@ -246,6 +246,7 @@ class icarus::crt::CRTPMTMatchingAna : public art::EDAnalyzer {
   double fOpFlashTime_us;  // Time of the optical Flash w.r.t. Global Trigger
   double fOpFlashTimehit_us;
   double fOpFlashTimeAbs;
+  int fnOpHitAboveThreshold; // Number of Optical Hits above threshold
   bool fInTime;  // Was the OpFlash in beam spill time?
   bool fInTime_gate;
   bool fInTime_beam;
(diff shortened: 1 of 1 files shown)


icaruscode/CRT/CRTPMTMatchingAna_module.cc
```diff
@@ -495,9 +497,11 @@ void icarus::crt::CRTPMTMatchingAna::analyze(art::Event const& e) {
       }
       geo::Point_t flash_pos = flashCentroid.middlePoint();
       t_m = t_m / nPMTsTriggering;
-      if (nPMTsTriggering < fnOpHitToTrigger) {
+      // F. Poppi: I will leave this commented in case we want to revert it to the original A. Scarpelli suggestion. 
+      // I'd rather have this as a variable to cut when doing the analysis.
+      /*if (nPMTsTriggering < fnOpHitToTrigger) {
```
PetrilloAtWork, line 502: Doesn't Clang get upset that `fnOpHitToTrigger` is set but not used?
We'll see...

PetrilloAtWork: If you leave this commented out, it may be worth either documenting why you should _not_ restore it as is, or change the code so that it would work even with this line.
The latter can be achieved by using temporary vectors and then move the content to the ones bound to the tree later, like filling a `std::vector<double> opHitX` first and finally  calling `append(fOpHitX, opHitX);` with
```cpp
template <typename Dest, typename Src>
void append(Dest& dest, Src const& src) { dest.insert(dest.end(), src.begin(), src.end()); }
```

PetrilloAtWork (approved):
I have left a couple of comments because it's me after all.
But the fix seems sound.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_78_06

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_78_06

PR #647, opened 2023-12-12
Feature/cerati crtpmtfilter fcl

cerati:
This PR adds a fcl to run the CRT-PMT filter and copy forward the TPC RAW data only for events that pass the filter.
A producer module to make this copy is added as well.

The goal is to be able to reduce the size of the RAW data we need to process, starting from the data we already collected and currently store on disk.

### DIFF
--- a/fcl/reco/Stage0/Run2/stage0_run2_icarus_crtpmtfilter.fcl
+++ b/fcl/reco/Stage0/Run2/stage0_run2_icarus_crtpmtfilter.fcl
@@ -1,0 +1,41 @@
+###
+## This fhicl file is used to filter events that do not pass the "medium" level
+## of the CRT-PMT filter. Only events passing are saved, in the same RAW format as before.
+## A second root file is saved with the CRT-PMT information for all events,
+## to allow studies on the filter performance.
+##
+#include "stage0_icarus_driver_common.fcl"
+
+process_name: CrtPmtFilter
+
+## Define the path we'll execute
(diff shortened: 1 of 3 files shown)


SFBayLaser:
Does it make more sense to keep the noise filtered RawDigits vs the original daq data? Meaning will the compressed RawDigits take less space that daq -or- is it more or less and even split?

cerati:
> Does it make more sense to keep the noise filtered RawDigits vs the original daq data? Meaning will the compressed RawDigits take less space that daq -or- is it more or less and even split?

This fcl does not run noise filtering, the goal is to store the original daq data to be able to reprocess it from scratch. Is noise filtering stable enough that we'll never need to re-run it? In that case we could have another fcl that does that.

SFBayLaser:
Ah, my mistake, I read "daq" (and read past its definition) and assumed the decoder.

SFBayLaser:
Last question, what is run time for this? Next iteration of noise filtering has decode step down to 30 sec/event or less which is a fraction of the root output stage.... if that holds true on gpvms when why not run decode step and save RawDigits? Saves time for the processing... (though that will be dominated by 2D decon).

cerati:
On icarubuild02 it takes 11 seconds per event on average for 50 events, including the root output which takes 4 seconds.

francescopoppi (approved):

mmrosenberg:
trigger build

SFBayLaser (approved):

PR #650, opened 2023-12-14
This commit fixes a bug in the CRT-PMT matching analyzer module. 

francescopoppi:
Previously there was the requirment for a certain number of opHits to be above threshold to consider the flash valid. The bug did not initialze some vectors when the nopHitAboveThreshold was lower then the requirments. Since in the future threshold will change, I removed the strong skipping-the-flash condition and I added the number of opHits above threhsold as a variabile for the analyzers to cut on.  This PR was previously approved by reviewe.


mmrosenberg:
trigger build

PR #651, opened 2023-12-14
Update to Energy Scale Calibration

gputnam:
Updates calibration factors and implements EMB recombination.


mmrosenberg:
trigger build

mrmooney (approved):
Looks fine to me, thanks!

SFBayLaser (approved):
Is there a better way to handle the gains for the electronic responses? I ask because we need to match these to the deconvolution now and I wonder if they should be in a parameter file or something common to both simulation and deconvolution?

PR #653, opened 2023-12-21
Change trigger efficiency NTuple hit-level dE/dx information to output of standard calorimetry modules

jzettle:
One-line change to trigger efficiency modules to extract the hit-level dE/dx information from the standard calorimetry modules in the NTuples used for the trigger efficiency analyses instead of a homebrew function in the module. Tested extensively that the production works and values are output using the production branch v09_72_00_05p04 by Riccardo Triozzi and shown to work. Can test again directly with this version if requested by review.

### DIFF
--- a/icaruscode/Analysis/trigger/TimeTrackTreeStorageCRT_module.cc
+++ b/icaruscode/Analysis/trigger/TimeTrackTreeStorageCRT_module.cc
@@ -1145,7 +1145,7 @@ sbn::selHitInfo sbn::TimeTrackTreeStorage::makeHit(const recob::Hit &hit,
         hinfo.oncalo = true;
         hinfo.pitch = c->TrkPitchVec()[i_calo];
         hinfo.dqdx = c->dQdx()[i_calo];
-        hinfo.dEdx = dEdx_calc(hinfo.dqdx, fMODA, fMODB, fWion, fEfield);
+        hinfo.dEdx = c->dEdx()[i_calo];
         hinfo.rr = sortRange(c->ResidualRange()[i_calo]);
         break;
       } // for i_calo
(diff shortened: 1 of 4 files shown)


mmrosenberg:
trigger build

jzettle:
The CI failure for c14:prof (clang) mentions in the log it failed from: 
```
4945: /scratch/workspace/icarus_ci/label_exp/SLF7/label_exp2/swarm/ICARUS/srcs/icarus_signal_processing/icarus_signal_processing/Denoising.h:248:10: error: private field 'fOutputStats' is not used [-Werror,-Wunused-private-field]
4946:     bool fOutputStats;
4947:          ^
4948: 1 error generated.
```
That is not related to anything that I changed so I'm not sure the right path to resolve. Unless I found a red herring...

mmrosenberg:
@jzettle the c14 builds have been failing for some time now (for other experiments as well) due to a known larsoft issue that we are still waiting on a fix for. The e26 gcc build was successful, so I think we're good, but I'll take another look at the clang failure to make sure no additional issues were introduced by other recent PRs. Either way don't worry about it for now

jzettle:
@mmrosenberg Sounds good, thanks for clarifying that. I did want to check to make sure I didn't identify something on my end.

mmrosenberg:
Taking a closer look, the clang build was actually supposed to be successful, it's the CI tests that are expected to fail due to that larsoft issue I mentioned. But the build failure was introduced by a recent icarus_signal_processing PR (which I missed there), so @jzettle I can confirm that this has nothing to do with your updates - I will coordinate a fix separately. Thanks for bringing this to my attention!

PetrilloAtWork (approved):
The change is ok.
However, they make a few variables and configuration parameters unused and obsolete.
I have somewhere in a branch changes that will remove that code and configuration. You may also consider to do that here and now thought.

jzettle:
I removed the now-obsolete function and fhicl parameters that went into it. I thought to do the quickest thing first but I agree that if we don't need it we should remove it

jzettle:
I seem to have forgotten to remove the obsolete parameters in the constructor. Trying again

jzettle:
trigger build

mmrosenberg:
trigger build

PR #654, opened 2023-12-21
New tags for Run 3 laser corrections

mvicenzi:
This PR updates the PMT calibration database tags to pick up the most recent corrections which should cover the beginning of Run 3. In particular:

- Cable delays are updated given the summer work on the laser system (`v2r4`).
- New laser corrections are included following the replacement of the signal cables and the new gain equalization (`v2r2`).

**Note**: This change requires a concurrent release of `icarus_data` containing the updated database files.

-------

@PetrilloAtWork Addressing here your comment from below: the cosmics-based corrections downstream are computed starting from the laser-correct ophit times and added on top. Assuming they get invalidated if you change the laser corrections is a conservative assumption we make. Comparing actual data, the issue is not as bad as we have claimed in the past. If you look at the plot below, you can see the cosmics residuals from run 9595 (first calibration period of RUN-2) corrected with: the correct cosmics corrections for that period (green), cosmics corrections from a wrong RUN-2 period (blue) and cosmics corrections from RUN-1 (orange). No relevant hardware change between green and blue, several between green and orange (adders + new voltages). Despite the laser corrections having changed, the same cosmics corrections could be applied but the price you pay are longer tails.
CORRECTION_VALIDITY_TEST

### DIFF
--- a/fcl/configurations/calibration_database_GlobalTags_icarus.fcl
+++ b/fcl/configurations/calibration_database_GlobalTags_icarus.fcl
@@ -6,7 +6,7 @@ BEGIN_PROLOG
 
 ICARUS_Calibration_GlobalTags: {
   @table::TPC_CalibrationTags_Oct2023
-  @table::PMT_CalibrationTags_Run2_Dec2023
+  @table::PMT_CalibrationTags_Run3_Dec2023
   @table::CRT_CalibrationTags_Oct2023
 }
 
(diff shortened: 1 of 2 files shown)


mmrosenberg:
trigger build

PetrilloAtWork:
@mvicenzi, can you repeat here (in the first post above) your arguments about this change invalidating the downstream cosmic-ray-based time correction?

jedori0228 (approved):
I see the updated tags from `icarus_data` `v09_82_00`, so we can now merge this PR.

mvicenzi:
> @mvicenzi, can you repeat here (in the first post above) your arguments about this change invalidating the downstream cosmic-ray-based time correction?

I've updated the first post to address your question @PetrilloAtWork

PetrilloAtWork (approved):

mmrosenberg:
trigger build

SFBayLaser (approved):
I am ok to approve this

PR #657, opened 2024-01-17
add wcls jsonnet for refactored g4

cerati:
This PR adds a new jsonnet file for the new refactored g4. The jsonnet file is identical to the standard one except for the input label to the wclsDepoFluxWriter, which is set to 'ionization'. The fcl file is updated consequently. This is needed to be able to test the new G4 in larger scale productions.


SFBayLaser (approved):
Looks ok to me. 
I hope that once we make this change we can start purging the old versions...

mmrosenberg:
trigger build

PR #660, opened 2024-01-22
apply offset to simulated CRT Hit times as the CRT Hits are being produced

aheggest:
This pull request applies fGlobalT0Offset, a fhicl parameter that should be applied to simulated CRT Hit times only, to the CRTHitProducer. This offset gets [added to the true time in the CRTDetSimAlg](https://github.com/SBNSoftware/icaruscode/blob/develop/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc#L616) before passing the true time to a timestamp generator. This offset should be subtracted back off so the simulated CRT Hit distribution is centered around 0, ranging from -1.5 ms to +1.5 ms. 
This offset is currently applied to the simulated CRT Hit times in the CAFs and the CRT-PMT Matching producer, but is not applied to _all_ simulated CRT hit times (for ex., this offset is missing in calibration ntuples). This is a flaw in the workflow; we should have this offset applied as high up in the chain as possible, as the CRT Hits are being made, so that the CRT hit time distribution looks the same regardless of what type of file is being looked at (see [docdb 34586](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=34586) for more info).

Keeping this PR as a draft while I continue to validate my changes and ask for reviews from reviewers.

### DIFF
--- a/fcl/caf/cafmakerjob_icarus.fcl
+++ b/fcl/caf/cafmakerjob_icarus.fcl
@@ -50,7 +50,9 @@ physics:
 # MCT0Offset for MC ()
 # ref1 : https://github.com/SBNSoftware/icaruscode/blob/v09_37_02_01/icaruscode/CRT/crtsimmodules_icarus.fcl#L11
 # ref2 : https://github.com/SBNSoftware/icaruscode/blob/v09_64_01/icaruscode/CRT/crttruehitproducer.fcl#L8
-physics.producers.cafmaker.CRTSimT0Offset: @local::standard_icarus_crtsimalg.GlobalT0Offset
+# physics.producers.cafmaker.CRTSimT0Offset: @local::standard_icarus_crtsimalg.GlobalT0Offset
+# instead, apply this^ offset in CRTHitRecoAlg.cc, see https://github.com/SBNSoftware/icaruscode/pull/660
+
 # Blinding not needed for MC
 physics.producers.cafmaker.CreateBlindedCAF: false
 # MC times are already measured w.r.t. the beam gate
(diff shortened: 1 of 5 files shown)


francescopoppi (commented):
Thank you @aheggest, I would kindly ask you to add a few plots as comment to this PR or load them in a docDB and reference the document both in a comment here and in a comment in the code.
Especially I would like to have reported:
A) time distribution of the CRT Hits in MC
B) plot of the CRT-PMT matching data product time difference
C) plot of A) and B) in the CAFs.
Would it be possibile?

tboone0042:
Hi Anna! Things are looking good to me, I'd also be curious if you have some plots demonstrating  that this does not produce timestamps that are out of the norm from previous analyses, along the lines of what Francesco is asking? It'd be good to confirm that the resulting timestamps aren't changed from this in a way that is incorrect. Other than that, things look good to me! Thanks for putting this together!

tboone0042 (approved):
I'll put my approval down, though it'd still be good to see some example plots like Francesco mentioned!

aheggest:
thanks @francescopoppi and @tboone0042! I will add these reference plots.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

aheggest:
@francescopoppi please see https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=34763 for the requested plots. please let me know if you have any questions! I can update the plots in anyway if needed. I'll also add this as a reference comment in the code

francescopoppi (approved):
Thanks Anna for the plots

mmrosenberg:
trigger build

PR #661, opened 2024-01-23
Feature/hhausner test compressed tpc fragments for pull request

hausnerh:
### Description

Update TPC decoder tools to read ADC values using the overlays instead of indexing a raw pointer to to the artdaq::Fragment. This is necessary for the decoder tools to function with compressed TPC data as they are
- non-uniform in size
- store ADC differences instead of ADC values
using the ICARUS PhysCrateOverlay's adc_val alleviates this concern as it can be modified to properly read from the compressed Fragments.

Additionally I've created some tools to test and debug the compressed data in the `icaruscode/TPC/Compression` directory. These include a module to convert uncompressed data into a compressed format without access to the updated overlays. However the ability to processes compressed data will require an updated offline version of sbndaq-artdaq-core to access the new additions to the PhysCrateOverlays.

### Related Repository Branches

To set the TPC compression in the board readers we need the [hhausner_setTPCCompression branch of sbndaq-artdaq](https://github.com/SBNSoftware/sbndaq-artdaq/pull/135). The default behavior is to run in the uncompressed mode, but this allows us to configure the A2795 boards to produce compressed fragments.

The updates to the ICARUS PhysCrateOverlay are in the [feature/hhausner_TPCCompression_forPullRequest branch of sbndaq-artdaq-core](https://github.com/SBNSoftware/sbndaq-artdaq-core/pull/96). These changes need to be available for to icaruscode. But for the time being do not impact the ability to process uncompressed data.


mmrosenberg:
trigger build SBNSoftware/sbndaq-artdaq-core#96

cerati (approved):
I do not have the expertise to review ilne-by-line changes. But I have tested these changes offline and confirm that they work as intended

hausnerh:
@SFBayLaser what would be the best way to get the updated PhysCrateFragment overlays into the offline build of sbndaq-artdaq-core? That will be necessary for offline processing of compressed data.

dtorretta56 (commented):
I have no expertises on the icaruscode to be able to approve or reject this PR.

icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
```diff
@@ -585,20 +585,15 @@ void DaqDecoderICARUSTPCwROI::processSingleFragment(size_t
...
-
-        // Get the pointer to the start of this board's block of data
-        const icarus::A2795DataBlock::data_t* dataBlock = physCrateFragment.BoardData(board);
-
         // Copy to input data array
         for(size_t chanIdx = 0; chanIdx < nChannelsPerBoard; chanIdx++)
         {
            icarus_signal_processing::VectorFloat& rawDataVec = channelArrayPair.second[chanIdx];
+           for (size_t tick = 0; tick < nSamplesPerChannel; ++tick)
+             rawDataVec[tick] = physCrateFragment.adc_val(board, chanIdx, tick);
```
SFBayLaser, line 593: Does this new method preserve the sign of the ADC value (note the "-" sign in the removed code)

SFBayLaser (commented):
Similar question for the other modules modified in same way. 
With answer to that I can sign off.

SFBayLaser:
> @SFBayLaser what would be the best way to get the updated PhysCrateFragment overlays into the offline build of sbndaq-artdaq-core? That will be necessary for offline processing of compressed data.

Sorry I did not see this sooner. I believe at this point the sbn art daq release is in progress (or done)? So question no longer relevant...

icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
hausnerh: Ah, no it doesn't. The function `adc_val(board, chanIdx, tick)` returns the ADC as a positive value. It's not entirely clear to me why we want to store the ADCs as negative values, but since I don't want to introduce unintended behaviors I can add in the `-`.

hausnerh: I notice `icaruscode/Decode/DecoderTools/TPCDecoder_tool.cc`, unlike the other decoder tools, doesn't have the `-`. Should I leave that one as is?

hausnerh:
> > @SFBayLaser what would be the best way to get the updated PhysCrateFragment overlays into the offline build of sbndaq-artdaq-core? That will be necessary for offline processing of compressed data.
> 
> Sorry I did not see this sooner. I believe at this point the sbn art daq release is in progress (or done)? So question no longer relevant...

Correct, you can disregard that question

icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
SFBayLaser: When we saw the first waveforms from the TPC they had negative polarity which is not the way that “standard” LArSoft tools were expecting to see things - hence the minus sign you saw on the decoder tools we actually use. If the new function is return a positive polarity waveform then all is good. Otherwise we will need the minus sign.

hausnerh: So I don't know about the polarity, but the values being read out of the fragments are negative when I use `rawDataVec[tick] = -dataBlock[chanIdx + tick * nChannelsPerBoard]`, but are positive when I use `rawDataVec[tick] = physCrateFragment.adc_val(board, chanIdx, tick)`. I tested this by running `stage0_run2_icarus.fcl` on `data_dl1_fstrmOffBeamBNBMINBIAS_run11418_6_20240118T024159.root` with the following debug check:
```
         // Copy to input data array
         for(size_t chanIdx = 0; chanIdx < nChannelsPerBoard; chanIdx++)
         {
            icarus_signal_processing::VectorFloat& rawDataVec = channelArrayPair.second[chanIdx];
            for (size_t tick = 0; tick < nSamplesPerChannel; ++tick)
            {
              rawDataVec[tick] = -physCrateFragment.adc_val(board, chanIdx, tick);
              float advValue = physCrateFragment.BoardData(board)[chanIdx + tick * nChannelsPerBoard];
              if (tick == 7)
                mf::LogDebug(fLogCategory)
                  << "FragmentID " << std::hex << fragmentID << std::dec << ", Crate " << crateName<< ",  board " << board 
                  << ", channel " << chanIdx << ", sample " << tick << " has ADC value " << advValue << " | " << rawDataVec[tick];
            }
 
           // Keep track of the channel
           channelArrayPair.first[chanIdx] = channelPlanePairVec[chanIdx];
         }
```
Which shows the methods give opposite signs:
```FragmentID 1700, Crate WW11,  board 6, channel 2, sample 7 has ADC value 2020 | -2020```
So if we want to preserve the current function of the Decoder Tools we should sill have the `-`.

mmrosenberg:
trigger build

icaruscode/Decode/DaqDecoderICARUSTPCwROI_module.cc
SFBayLaser: If I understand the last comment then we need to introduce a "-" sign to preserve current behavior. This is important for the deconvolution since it expects to see waveforms with polarity as it currently returned. Is it easy (and possible) to overlay a waveform near a signal for the two methods to make sure we have things right?

SFBayLaser (approved):
I believe the sign change is correct so will approve. Thanks very much!

mmrosenberg:
trigger build

PR #662, opened 2024-01-25
materials for the simulation of dirts BNB events added

cfarnese:
I added here few things that can be used to produce dirt events for the BNB beam! The code I must confess is not so "beautiful" but I commit it in any case to try to have it for the new release! In particular there are :

- a couple of fhicl files in fcl/gen/genie to generate the dirts events

- a filter module and its fhicl to remove the neutrino interactions with the vertex inside the active volume or that did not deposit energy in the detector

### DIFF
--- a/icaruscode/Filters/FilterDirts_module.cc
+++ b/icaruscode/Filters/FilterDirts_module.cc
@@ -1,0 +1,578 @@
+/**
+ * @file   icaruscode/Filters/FilterDirts_module.cc
+ * @brief  Select only neutrino events with the primary vertex outside the active volume and that deposit energy in the active volume (dirt events)
+ * @authors Christian Farnese (farnese@pd.infn.it),
+ *
+ * @date   Jan 10, 2024
+ * 
+ *
+ */
+
+
+// ICARUS libraries
+#include "icarusalg/Utilities/WeakCurrentType.h"
+
+// LArSoft libraries
+#include "larcore/Geometry/Geometry.h"
+#include "larcore/CoreUtils/ServiceUtil.h" // lar::providerFrom()
+#include "lardataalg/MCDumpers/MCDumperUtils.h" // sim::TruthXXXName()
+#include "larcorealg/Geometry/GeometryCore.h"
+#include "larcorealg/Geometry/ROOTGeometryNavigator.h"
+#include "larcorealg/Geometry/GeoNodePath.h"
+#include "larcorealg/Geometry/TransformationMatrix.h"
+#include "larcorealg/Geometry/geo_vectors_utils.h" // geo::...::makeFromCoords()
+#include "larcorealg/Geometry/BoxBoundedGeo.h"
+#include "larcorealg/CoreUtils/StdUtils.h" // util::begin(), util::end()
+#include "larcorealg/CoreUtils/enumerate.h"
+#include "larcoreobj/SimpleTypesAndConstants/geo_vectors.h" // geo::Point_t
+#include "lardataobj/Simulation/SimPhotons.h"
+#include "lardataobj/Simulation/SimChannel.h"
+
+// nutools libraries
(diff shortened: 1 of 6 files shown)


SFBayLaser:
trigger build

SFBayLaser:
Interesting question... should we require "beauty" before we sign off on a PR?

cfarnese:
I believe usually the code should be clear and as simple as possible... I must confess in this case I prefer to make something "quick and dirt" because the main goal was to arrive in time and unfortunately the many activities don't provide me the proper time to make the code clearer.. sorry for this!

mmrosenberg:
trigger build

SFBayLaser (approved):
Ok, quick and dirty… but approving!

icaruscode/Filters/FilterDirts_module.cc
```diff
@@ -0,0 +1,578 @@
...
+#include "larcorealg/Geometry/GeometryCore.h"
+#include "larcorealg/Geometry/ROOTGeometryNavigator.h"
+#include "larcorealg/Geometry/GeoNodePath.h"
+#include "larcorealg/Geometry/TransformationMatrix.h"
+#include "larcorealg/Geometry/geo_vectors_utils.h" // geo::...::makeFromCoords()
+#include "larcorealg/Geometry/BoxBoundedGeo.h"
+#include "larcorealg/CoreUtils/StdUtils.h" // util::begin(), util::end()
+#include "larcorealg/CoreUtils/enumerate.h"
+#include "larcoreobj/SimpleTypesAndConstants/geo_vectors.h" // geo::Point_t
+#include "lardataobj/Simulation/SimPhotons.h"
```
PetrilloAtWork, line 28: Unused?
```suggestion
```

icaruscode/Filters/FilterDirts_module.cc
```diff
@@ -0,0 +1,578 @@
...
+ */
+
+
+// ICARUS libraries
+#include "icarusalg/Utilities/WeakCurrentType.h"
+
+// LArSoft libraries
+#include "larcore/Geometry/Geometry.h"
+#include "larcore/CoreUtils/ServiceUtil.h" // lar::providerFrom()
+#include "lardataalg/MCDumpers/MCDumperUtils.h" // sim::TruthXXXName()
```
PetrilloAtWork, line 18: Unused?
```suggestion
```

PetrilloAtWork (changes requested):
The name of `SimChannel` data product needs to be moved into a configuration parameter.
The rest are suggestions.

PetrilloAtWork (approved):
The comments from my previous review persist.
The submitter has explained me the role of `GenInTimeSorter`, which indeed hides a bug.
The sorter will take all generated stuff, cosmic rays from CORSIKA and neutrino interactions from GENIE, and will split them in "in time" and "out of time". The first category is then assumed to have been GEANT'd already, while the second category will be processed through GEANT4 in the next stage (e.g. `intime_g4_icarus.fcl`) and merged with the GEANT4 output of this stage.
The problem is that in-time cosmic rays, if any, will be assumed to have been processed already, while that's not the case (here GEANT4 only processed the neutrino events), resulting in the neutrino events never having a cosmic ray close in time.
The probability of such overlap is not large, though (10-kHz cosmic rays appear in a 2-µs gate once every 50 times).

With the submitter we agreed that the points raised in my review will be addressed ideally next week, and anyway in time for the production release. In the meanwhile, this PR should be merged as is in time for the _"intermediate"_ validation release.

**To the release managers: please do not delete the branch after merging.**

cfarnese:
I Introduce with the last commits almost all the corrections suggested by Gianluca! I tested this morning the code and it provides compatible results with respect to the previous version!

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #668, opened 2024-01-31
Beginning fcl paths for ICARUS detector variation samples

jzettle:
Start of the idea for preparing for making ICARUS detector variation samples. This PR encapsulates the idea for these samples and creates a set of fcls to go back and take stage1 inputs and drop all products after the largeant part of the G4 stage. The fcls in here inelegantly split the G4 stage into two parts: one that runs the largeant step and one that runs the rest of the stage (happy to take suggestions on this). This is in order to preserve the G4 steps for a given event when we run detector variations along with the identical set of generated events.  Then there are fcl to rerun the original stages providing a template for the variations. These fcls exist solely due to the art requirement that a stage cannot include the same `process_name` or else it does not run. 

This PR is made now in order to help the production team test workflows using a tagged release working chiefly with Mateus Carneiro (who I don't clearly see in here). I tried to inform @SFBayLaser and @PetrilloAtWork (who I also tag as reviewers) so hopefully they are not too surprised. I will commit a better naming scheme as Gianluca requested privately. Other suggestions/thoughts are welcome.

In order to implement this scheme we would need to save the data products from the G4 stage up to and including the largeant step in the stage1 files which means saving those as well as part of a production.

### DIFF
--- a/fcl/CMakeLists.txt
+++ b/fcl/CMakeLists.txt
@@ -9,5 +9,6 @@ add_subdirectory(caf)
 add_subdirectory(services)
 add_subdirectory(utilities)
 add_subdirectory(compatibility)
+add_subdirectory(syst_variations)
 
 install_fhicl()
--- a/fcl/syst_variations/CMakeLists.txt
+++ b/fcl/syst_variations/CMakeLists.txt
@@ -1,0 +1,1 @@
+install_fhicl()
--- a/fcl/syst_variations/caf_variations_standard.fcl
+++ b/fcl/syst_variations/caf_variations_standard.fcl
@@ -1,0 +1,3 @@
+#include "cafmakerjob_icarus_detsim2d.fcl"
+
+process_name: CAFmakerVar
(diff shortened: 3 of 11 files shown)


mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PetrilloAtWork (commented):
~Three~ Four comments:
1. some of the FHiCL files seem to have a name not closely related from their original counterpart (I should have marked them in the comments);
2. not clear why many G4 configurations were copied instead of included and modified;
3. documentation should be updated where appropriate; especially, where there are hard-coded numbers, a source of them should be quoted (a SBN DocDB number is well enough, although more extensive description does not hurt)
4. I did not pay too much attention to the file names... considering that (a) they grow longer and longer and (b) there is value, in my opinion, in preserving the original file name, you could consider having output file names `%ifb_%p.root`, omitting the creation timestamp `%tc`: that cuts some information but also shortens the file names a lot (one needs to make sure that each variation has a different process name though, or file names will collide).

jzettle:
@PetrilloAtWork I think I addressed everything I see here. Let me know if I did anything unsatisfactorally

mmrosenberg:
trigger build

PetrilloAtWork (approved):
Looks good, thank you!

mmrosenberg:
trigger build

Issue #449, opened 2022-09-16
Warning in PMT digitisation: "Check on sampled photoelectron waveform template failed!"


PetrilloAtWork:
PMT digitisation simulation emits these warnings:
```
%MSG-w PMTsimulationAlg:  SimPMTIcarus:opdaq@BeginModule  16-Sep-2022 17:41:50 CDT run: 1 subRun: 0 event: 4
Check on sampled photoelectron waveform template failed!
 => high tail at the end of sampling (842 ns) is still at -0.0351564 #
Shape parameters:Pulse 'SPR202202' (v. 1, n/a):
    average waveform of channel 258 of run XXXX
    from 804x 1 ns samples, peak at 55.1 ns with amplitude -32.8509 #
    start at 39.1 ns, gain 9.7e+06 (integral: -638.122 #)

%MSG
```
The single photon response is supposed to be defined widely enough that it has already faded out at the end of the interval of definition. With our current data-driven SPR (`202202`), this is not the case and the warning is triggered.
The best solution is not clear (whether to work around the warning or to change the SPR).

The relevant code is somewhere in [`icaruscode/PMT/Algorithms`](https://github.com/SBNSoftware/icaruscode/tree/develop/icaruscode/PMT/Algorithms).

wesketchum:
@PetrilloAtWork is this still an issue?

PR #447, opened 2022-09-14
Set PFO Characterisation BDT to persist features.

brucehoward-physics:
Marking as draft for now -- I want to make the PRs to do a CI-based test first.

PR #673, opened 2024-02-07
Introduce fhicl to enable adjustment of the ICARUS simulation according to trigger emulation 

woodtp:
This the companion detsim fhicl that inserts the module described in [sbncode PR #421](https://github.com/SBNSoftware/sbncode/pull/421) into the ICARUS simulation chain. It is a modification of [`icaruscode/fcl/detsim/detsim_2d_icarus.fcl`](https://github.com/SBNSoftware/icaruscode/blob/v09_82_01/fcl/detsim/detsim_2d_icarus.fcl) (also includes the contents of [`icaruscode/fcl/detsim/detsim_2d_icarus_fitFR.fcl`](https://github.com/SBNSoftware/icaruscode/blob/v09_82_01/fcl/detsim/detsim_2d_icarus_fitFR.fcl)) which inserts `shiftIonization` between optical simulation + trigger emulation and the tpc + crt simulation. To update products based on the applied time shifts, the trigger simulation is re-run after `shiftIonization`. In summary, the order of the detector simulation becomes:
1. `rns`
2. `opdaq`
3. trigger simulation
4. `shiftIonization`
5. re-simulate trigger
6. `daq`
7. `crtdaq`

The `AdditionalOffset` parameter was added to provide extra flexibility for tuning the time shift.


woodtp:
I've added modified fcls and WC jsonnets that make the new detsim procedure compatible with [#684 Move to Refactored LArG4](https://github.com/SBNSoftware/icaruscode/issues/684).

mmrosenberg:
trigger build SBNSoftware/sbncode#421

brucehoward-physics:
I think even though the module was merged, Tony and I were hoping for @PetrilloAtWork to weigh in on whether this whole thing is good-to-go (as both trigger expert and whether we're doing the right thing with at least the photon system products). So though the sbncode PR was merged, maybe we should wait for more confirmations before making this the default workflow (and if there is any conversation that needs to happen as we go towards the new production on the foreseen workflow).

PetrilloAtWork (changes requested):
With @woodtp we have verified that optical reconstruction is performed on the unshifted waveforms.
He is out to fix that in the configuration, after which we'll check the flash time again.
Also, I suggested that the trigger emulation (the "final" one, performed on the shifted waveforms and expected to always give a result close to 0) be executed in `MCstage0` rather than earlier in `DetSim` stage, for consistency with the standard workflows.

woodtp:
> With @woodtp we have verified that optical reconstruction is performed on the unshifted waveforms. He is out to fix that in the configuration, after which we'll check the flash time again. Also, I suggested that the trigger emulation (the "final" one, performed on the shifted waveforms and expected to always give a result close to 0) be executed in `MCstage0` rather than earlier in `DetSim` stage, for consistency with the standard workflows.

@PetrilloAtWork I have implemented this change. I also found the places in stage0 where I needed to replace `opdaq` as the input module with `shiftIonization`. I tested the new detsim and stage0 fcls for both standard and refactored LArG4 workflows, and things appear to be working as expected now.

jzennamo:
trigger build https://github.com/SBNSoftware/sbncode/pull/421

woodtp:
I have amended the module responsible for shifting simulation products based on the trigger time to now also include shifts to `sim::AuxDetIDE`, and I have updated the detsim fhicl files to incorporate these changes and override the input label to `crtdaq`. I have confirmed that the CRT Hit times are now shifted in response.

fcl/reco/Stage0/Run2/stage0_run2_icarus_mc_shiftForTrigger_refactored.fcl
```diff
@@ -0,0 +1,9 @@
+#include "stage0_run2_icarus_mc.fcl"
+
+
+physics.producers.mcophit.SimPhotonsProducer: "shiftIonization"
+physics.producers.ophit.InputModule: "shiftIonization"
+physics.producers.opdetonbeam.Waveforms: "shiftIonization"
+physics.producers.pmtbaselines.OpticalWaveforms: "shiftIonization"
+physics.producers.pmtfixedthr.OpticalWaveforms: "shiftIonization"
+physics.producers.triggersimgates.BeamGateTag: "shiftIonization"
```
brucehoward-physics, line 9: This seems to be the same as `stage0_run2_icarus_mc_shiftForTrigger.fcl` -- is it missing a `_refactored` in the include in line 1 maybe?

fcl/detsim/detsim_2d_icarus_shiftForTrigger.fcl
```diff
@@ -0,0 +1,98 @@
...
+    outputCommands: [
+      "keep *",
+      "drop *_ionization_*_*",
+      "drop *_pmtlvdsgatesinit_*_*",
+      "drop *_pmtfixedthrinit_*_*",
+      "drop *_pmttriggerwindowsinit_*_*",
+      "drop *_triggersimgatesinit_*_*"
+      ]
+  }
+}
```
brucehoward-physics, line 98: Does this fhicl need something like ```physics.producers.daq.wcls_main.configs: ["pgrapher/experiment/icarus/wcls-multitpc-sim-drift-simchannel-shifted.jsonnet"]```
that is in the other files?

I guess no since the shift was tested? Or maybe just there is a local change not on the remote. (( Just checking while quickly looking through the changes! I can also think if I'm missing something here. ))

fcl/reco/Stage0/Run2/stage0_run2_icarus_mc_shiftForTrigger_refactored.fcl
woodtp: I decided, here, to copy and modify [stage0_run2_icarus_mc_refactored.fcl](https://github.com/SBNSoftware/icaruscode/blob/525f06b16ff203e4bf13f4b265f855d3d5d333ba/fcl/reco/Stage0/Run2/stage0_run2_icarus_mc_refactored.fcl), and have a parallel version rather than continuing the chain of includes. If it's preferred, I can modify this fhicl but the difference would simply be to add `_refactored` to the include on line 1. The original fhicl is only two lines: including the non-refactored stage0 fhicl and pointing `mcophit.SimPhotonsProducer` to `pdfastsim`.

brucehoward-physics:
In addition to the 2 comments given above, putting down that this also wants https://github.com/SBNSoftware/sbncode/pull/447

fcl/reco/Stage0/Run2/stage0_run2_icarus_mc_shiftForTrigger_refactored.fcl
brucehoward-physics: Is it cleaner if we actually just make
```
stage0_run2_icarus_mc_shiftForTrigger_refactored.fcl
```
literally just be a one-liner?
```
#include stage0_run2_icarus_mc_shiftForTrigger.fcl
```

Maybe invoking @PetrilloAtWork for his insight on the intentions of fhicl maintainers?

fcl/detsim/detsim_2d_icarus_shiftForTrigger.fcl
woodtp: 100% this is an oversight. Will push an update promptly.

woodtp: Done: https://github.com/SBNSoftware/icaruscode/pull/673/commits/dca13e09946c903191ed046e3e382629effe95fb

brucehoward-physics: Cool: thanks!

fcl/reco/Stage0/Run2/stage0_run2_icarus_mc_shiftForTrigger_refactored.fcl
woodtp: Now that I stare at these two files side-by-side they are actually identical, and we may not need to have separate fhicls here.

woodtp: The only reason they were separate previously, was because mcophit was (incorrectly) pointed to `pdfastsim`, but now that I fixed this they are completely redundant.

brucehoward-physics: Yeah that's why I was initially asking if something was missing 😅 - But the parallelism argument is fair.

aheggest (approved):
looks good, thanks Tony and Bruce

fcl/reco/Stage0/Run2/stage0_run2_icarus_mc_shiftForTrigger_refactored.fcl
woodtp: I tested that either file works for whichever version of the simulation chain is used, which makes sense as they are identical. Therefore, I've chosen to remove the redundant fhicl for simplicity. See: https://github.com/SBNSoftware/icaruscode/pull/673/commits/b8919bc2a9d80c33fa1bd72910c226743582fd11

PetrilloAtWork: > Maybe invoking @PetrilloAtWork for his insight on the intentions of fhicl maintainers?

I do not have such an insight, unfortunately. My expectation is that `refactored` is a temporary label, and that after a (short?) while everything with `run3` will be implicitly refactored, but I can't guess about Run2 — I think it depends a lot on what the new MC production will use.

woodtp:
After consulting with @mmrosenberg, I have updated the two base detsim fhicls with the option to shift the sim::AuxDetIDEs disabled. This is because https://github.com/SBNSoftware/sbncode/pull/447 will likely not make it in at the same time as this PR. I'll make another PR that will go in with a patch release, which will re-enable this option.

> [!CAUTION]
> Using these fhicl files as they are, without shifting `sim::AuxDetIDEs` will result in the CRT-PMT matching being **_BROKEN_**.

woodtp:
At @SFBayLaser's request, I have removed the new fhicls and worked the shifting procedure directly into the standard workflows.

> [!WARNING]
> This PR must not be merged without same time as [sbncode #447](https://github.com/SBNSoftware/sbncode/pull/447) and [sbncode #449](https://github.com/SBNSoftware/sbncode/pull/449).

fcl/detsim/detsim_2d_icarus.fcl
```diff
@@ -15,21 +16,39 @@ services: {
...
+      "drop *_pmttriggerwindowsinit_*_*",
+      "drop *_triggersimgatesinit_*_*"
+      ]
+  }
 }
+
+#legacy G4 configs
+services.Geometry.GDML: "icarus_complete_20220518_overburden.gdml"
+services.Geometry.ROOT: "icarus_complete_20220518_overburden.gdml"
+physics.producers.daq.wcls_main.configs: ["pgrapher/experiment/icarus/wcls-multitpc-sim-drift-simchannel-shifted.jsonnet"]
```
PetrilloAtWork, line 54: This is confusing.
This PR should not need to affect the GEANT4 model... it sounds like the default has changed to the different model and here we manually revert to another one that is not default any more.
Is upstream `detsim_2d_icarus.fcl` broken? Can this be clarified? Also question for @cerati and @SFBayLaser, I suppose.

icaruscode/PMT/Trigger/trigger_emulation_icarus.fcl
```diff
@@ -363,19 +363,79 @@ icarus_standard_triggersim: {
...
+    shifted: {
+       module_type: AdjustSimForTrigger
+       InputTriggerLabel: "emuTriggerUnshifted"
+       InitAuxDetSimChannelLabel: "largeant"
+       InitBeamGateInfoLabel: "triggersimgatesinit"
+       InitSimEnergyDepositLabel: "ionization"
+       InitSimPhotonsLabel: "largeant"
+       InitWaveformLabel: "opdaq"
+       ShiftAuxDetIDEs: true
+       ShiftBeamGateInfo: true
```
PetrilloAtWork, line 419: Can you remind me the rationale behind this?
```
       ShiftBeamGateInfo: true
```

I remember having discussed that the beam gate object (`sim::BeamGateInfo`) is on simulation time scale and therefore should not be shifted, but the issue was probably more involved than just this.

PetrilloAtWork (changes requested):
Two pending questions that I would like follow up with.
Also my fault that it took me so long to come back to this, that I have forgotten things.

I can't review the JSONnet stuff, somebody else will have to weigh in on that. I cautionary leave this in "request changes" state until that happens.

fcl/detsim/detsim_2d_icarus.fcl
woodtp: To be clear about the purpose of this commit (3510f6ae8ce0cb7b48f92e43162c0034a4154d14), I had previously removed these lines, which would break the legacy LArG4 configurations. Here, I am restoring what's currently present on `develop` ([detsim_2d_icarus.fcl](https://github.com/SBNSoftware/icaruscode/blob/develop/fcl/detsim/detsim_2d_icarus.fcl)) with an override to point WireCell to the jsonnet where I have altered the input tags for the simulated energy deposits.

For the newer LArG4 workflow, I've modified  `detsim_2d_icarus_refactored.fcl` in the same commit I mention above.

icaruscode/PMT/Trigger/trigger_emulation_icarus.fcl
woodtp: If I understood your previous advice correctly, I am shifting the `sim::BeamGateInfo` produced by `triggersimgatesinit`, which is then provided `emuTrigger` (the shifted one). I changed the input labels and give `triggersimgates` a `DummyProducer` [here](https://github.com/SBNSoftware/icaruscode/pull/673/files#diff-d3f990401fb3e7d310310cc20642a4b0cb120f1301e3bb0b17a4eb1f1104835fR71-R72).

fcl/detsim/detsim_2d_icarus.fcl
```diff
@@ -14,28 +15,38 @@ services: {
...
 }
 
 #legacy G4 configs
 services.Geometry.GDML: "icarus_complete_20220518_overburden.gdml"
 services.Geometry.ROOT: "icarus_complete_20220518_overburden.gdml"
-physics.producers.daq.wcls_main.configs: ["pgrapher/experiment/icarus/wcls-multitpc-sim-drift-simchannel.jsonnet"]
-physics.producers.crtdaq.G4ModuleLabel: "largeant"
-physics.producers.opdaq.InputModule: "largeant"
+physics.producers.daq.wcls_main.configs: ["pgrapher/experiment/icarus/wcls-multitpc-sim-drift-simchannel-shifted.jsonnet"]
+physics.producers.crtdaq.G4ModuleLabel: "shifted"
```
woodtp, line 52: @PetrilloAtWork, I have resolved the merge conflicts, which I think makes it clearer what is actually being changed by this PR in these lines.

SFBayLaser (commented):
It looks like we only need to resolve the two comments between Gianluca and Tony and this is good to go?

icaruscode/PMT/Trigger/trigger_emulation_icarus.fcl
PetrilloAtWork: Thank you for providing me some output to better understand what happens.

In principle, `sim::BeamGateInfo` is expected to be on simulation time scale (something that says `0` when the generator is ready to put signal in), which for neutrino samples is usually the time of arrival of the spill. So our generators put in the `generator` data product a gate that goes from -0.2 µs to say 2.0 µs (to make up for a 2.2 µs BNB gate). At the beginning of digitisation stage, this simulation time scale still matches the electronics time scale, which is what everything detected and reconstructed is based on (this is the scale that is trigger-based in data and in simulation was "beam-gate" based).
The module `TriggerSimulationOnGates` runs trigger emulation on a input "beam" gate, and it needs that gate in the optical time scale, which is the time scale the PMT waveforms are on. So that module converts the input gate into optical time scale assuming that
1. the input gate is on simulation time scale, because that's the standard;
2. the optical time scale the PMT waveforms are based on corresponds to the electronics time scale (de facto convention, and enforced by data decoding);
3. the electronics time scale starts with the simulation time scale (barred the offset `1500`), because there is no place in LArSoft where their relation is stored and this one is the "standard" practice.

This gate conversion results therefore trivial (like, shifting ahead of `1500` µs).

Now the configuration in this PR has changed the practice at point (3) so that the electronics time scale, and the optical one, match the trigger time scale, not the simulation time scale any more. The old gate conversion of `TriggerSimulationOnGates` won't suffice any more.

The ~solution~ _workaround_ in this PR is to break the standard at point (1) and provide `TriggerSimulationOnGates` a gate that, after the trivial transformation, encloses the appropriate time interval. This gate is in a "physics interaction" time scale (making it up here and now) that has the 0 at the time of the interaction (allegedly equivalent to the one of the trigger).

The module `TriggerSimulationOnGates` in `v09_89_01_01` (and probably earlier) has an additional degree of freedom: it allows to specify which time scale the input gate is in. This should allow the module to perform the appropriate transformation to optical time scale. So the _solution_ to the issue would be to use the new module with the appropriate time reference in the configuration (which should be something like `BeamGate` or `Simulation`). In practice,
1. this PR has no access to that new code
2. it is not guaranteed that the new code has already the needed conversion among the available options
3. furthermore, additional test would be needed to verify that the conversion feature works as expected, since it's quite new and it may need more testing.

The only thing we need to pay attention to is that with the new `TriggerSimulationOnGates`, the default beam gate reference reproduces the behaviour of the old one.

PetrilloAtWork (approved):
I don't have any further demand.
Thank you!

SFBayLaser (approved):

mmrosenberg:
trigger build

mmrosenberg:

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_90_00 SBNSoftware/sbncode@v09_90_00 SBNSoftware/sbnanaobj@v09_23_00_01 SBNSoftware/sbndaq-artdaq-core@v1_09_00of1 SBNSoftware/sbnobj@v09_19_01

PR #674, opened 2024-02-09
Feature/cerati refactored g4fcls

cerati:
This PR adds some missing fcl files for the new refactored G4, like cosmics_sce and all intime versions.


mmrosenberg:
trigger build

SFBayLaser (approved):
I am not sure I am familiar enough with the "new" G4 to make specific comments. My question might be more on organization... thinking we might want to create sufolders for "old" and "new" to separate the current fhicls from the old. That would allow an "easy" dropping of old in the future when we simply stop including the old folder in releases.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_82_02

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_82_02

PR #675, opened 2024-02-10
Update the drift velocity in the WireCell simparams.jsonnet

woodtp:
This PR updates the `drift_speed` within `icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet` to reflect measurements by @gputnam (see [SBN-doc-23346-v1](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=23346&filename=driftV.pdf&version=1)).

See also [icarusalg PR #81](https://github.com/SBNSoftware/icarusalg/pull/81)

#### Question for reviewers: 
Here, I am using a value based on the measurement for the West cryostat. Since I don't believe the drift speed can be set for each cryostat independently, would it be better to choose a value somewhere between the two measured drift speeds of each cryostat?

### DIFF
--- a/icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
+++ b/icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
@@ -12,7 +12,7 @@ base {
     // Electron lifetime
     lifetime : 3.5*wc.ms,
     // Electron drift speed, assumes a certain applied E-field
-    drift_speed : 1.5756*wc.mm/wc.us, // at 500 V/cm
+    drift_speed : 1.572*wc.mm/wc.us, // at 500 V/cm
   },
   daq: super.daq {
 
--- a/icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
+++ b/icaruscode/TPC/ICARUSWireCell/icarus/simparams.jsonnet
@@ -12,7 +12,7 @@ base {
     // Electron lifetime
     lifetime : 3.5*wc.ms,
     // Electron drift speed, assumes a certain applied E-field
-    drift_speed : 1.572*wc.mm/wc.us, // at 500 V/cm
+    drift_speed : 1.5714*wc.mm/wc.us, // at 500 V/cm
   },
   daq: super.daq {
 


woodtp:
Leaving this PR in draft mode while @brucehoward-physics and I perform some additional validation.

cfarnese (approved):

SFBayLaser:
This is a new measrement? Do we need to change to it right away?
It turns out that wirecell has more than one place that the drift velocity needs to be updated or else we will get inconsistencies.

woodtp:
> This is a new measrement? Do we need to change to it right away? It turns out that wirecell has more than one place that the drift velocity needs to be updated or else we will get inconsistencies.

@SFBayLaser, this is the same measurement from @gputnam that has already been incorporated in the 2D deconvolution. @brucehoward-physics's and I intended to also include this in the simulation by updating the WireCell jsonnet(s) and the `DriftVelFudgeFactor` within `icaruscode/fcl/services/detectorproperties_icarus.fcl`, which is done in the companion [PR #81](https://github.com/SBNSoftware/icarusalg/pull/81) in `icarusalg`.

Here, I've only updated the `drift_speed` in the `simparams.jsonnet` as it doesn't seem to overridden elsewhere. If there are additional places, could you please point me to those, if possible, so that I can update them as well?

SFBayLaser:
Unfortunately, the drift speed is also embedded in the field response files and some wirecell modules use that value.  A few  months ago we went through and updated the drift velocity in all places to the value 1.5756 cm/us (so this is what is used in the 2D deconvolution). So you have to pardon my confusion about changing yet again. Is this an updated measurement from Gray?

Because the drift speed is embedded in the field response files...  it would be my preference if we could push this PR for now and give time to sort out all the files that need to be udpated. Let me know if this has to be done right away.

brucehoward-physics:
Hi Tracy -- Gray's measurements that Tony references are from 2021 and are closer to 1.571-1.572, so we were going based on that.

1.57565 is a value that one gets if one asks the detector properties service what the drift velocity is based on input settings/values and the model we have chosen to use and not really a _measurement_ if I'm thinking about the right thing. _Is this the provenance of the 1.5756 number, or is there a more recent measurement? Or are we baking out another effect to arrive at this number?_

It's a fairly small difference, but for things drifting from the cathode to the anode we think this probably leads to a few (~5?) tick difference. So if we're right it's still small, but there. Anyway, just throwing in some thoughts and trying to understand better where numbers come from!

SFBayLaser:
Ok, confirmed with Gray that we should move to new value. But we have to push to the next release because I need to modify the response files that are in icarus_data.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_84_00 SBNSoftware/sbncode@v09_84_00_01 SBNSoftware/sbnanaobj@v09_22_01 SBNSoftware/sbndaq-artdaq-core@v1_08_00of4 SBNSoftware/sbnobj@v09_18_01_01 SBNSoftware/icarusalg#81

PR #676, opened 2024-02-12
SCE Configurations for 2D Drift Sim hack

jzennamo:
This adds fcl files to use the new SCE maps which flip the position x-offsets in the right-hand TPCs. 

The newest maps are in `icarus_data` and require `feature/jzennamo_SCE_2d_drift_sim_hack`

These need to replace all the production fcls.


jzennamo:
Also, validation will be done by Lane once the grid jobs stop failing

PetrilloAtWork (changes requested):
I am not sure how valuable my review is going to be, since one should test all this stuff carefully (which I am not doing).

Some comments:
1. I am seeing the refactored in-time configuration for the first time. Why was it chosen to have a complete simulation of in-time and out-of-time particles and merge it later, as opposed to simulating the minimal in-time information to feed the filter, and then merge with the rest and _then_ proceed with ionisation, photons etc.? I suppose it has to do with what is defined to be that "minimal" information, but I would hear your take on it.
2. Some of the configurations do not have the `_icarus` tag in the name. Considering that we may have in some configurations `sbndcode` set up, I recommend all configurations to be so tagged.
3. About the in-time: I think the purpose is to have an enriched sample that includes all triggered events. In that case, considering that the NuMI trigger gate is 10.1 µs long (from the beam time reference, it's about -0.4 µs to 9.7 µs, if I recall correctly), we should have larger intervals than the ones configured in the filters. BNB: roughly -0.4 µs to 1.8 µs.
4. I would have preferred all the recombination parameter variations to be collected in the same configuration file, and then included one in each of the various job configurations. That allows an overview of all the variations, and it should also make it easier to spot inconsistencies or holes. Please consider such change.
5. I am scared of this becoming a monster of unmaintainability, but I don't have solutions to that. Just rambling.

cerati:
I agree with the sentiment of @PetrilloAtWork about long term maintenance but I would like to suggest that a separate PR for that will follow later and for now we focus on getting the files in. The definition of a consistent set of fcl files needs to be done at the SBN level

jzennamo:
OK, hi everyone! I changed _how_ we do this fix after learning more about the SCE offsets in ICARUSCode. 

What I have done is decided not to use the new SpaceCharge maps which Mike made and instead hard-code a `-1` and a fcl flag for calling this during the 2D drift simulation workflow.

jzennamo:
The big change can be found in this file:
`icaruscode/TPC/Simulation/SpaceCharge/SpaceChargeICARUS.cxx`

jzennamo:
This has been tested and it works as expected:

Test at simulation:
image (3)

Test at reconstruction (from Lane) : 
refac_offsets

I think this is ready to merge.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #689, opened 2024-02-16
Updated DetectorPropertiesService from Jaesung and me

brucehoward-physics:
The T0 shift to be used on hits!

The work from Tony and me more recently is to get this handled in sim moving forward, but for analysis using already-produced files, it can be nice to be able to treat the hits as if they were shifted back in time a bit as the time of the event should be the time of the trigger and not the beginning of the beam spill. Largely copies the DetectorProperties(Service, Standard) but edits a few things as needed to insert the additional shift.


PetrilloAtWork (commented):
Two preliminary comments:
1. shouldn't this be in a directory not bound to TPC? like `icaruscode/Utilities`?
2. have you considered _deriving_ from the standard service provide class instead of copying it? That reduces the need for maintenance, since I don't think anybody is going to successfully track the original service and port bug fixes and interface changes (the latter will likely break compatibility so they may be easier to spot). I haven't downloaded the branch so I don't exactly know which are the changes, so I can't state what I suggested is possible.

icaruscode/TPC/Utilities/DetectorPropertiesICARUSClockOffsetMC.h
```diff
@@ -0,0 +1,294 @@
...
+//
+// \author brebel@fnal.gov
+//
+// Separation of service from Detector info class:
+// jpaley@fnal.gov
+//
+// Update for ICARUS by Bruce Howard and Jaesung Kim
+////////////////////////////////////////////////////////////////////////
+#ifndef DETINFO_DETECTORPROPERTIESICARUS_H
+#define DETINFO_DETECTORPROPERTIESICARUS_H
```
PetrilloAtWork, line 14: Please use a full specification:
```suggestion
#ifndef ICARUSCODE_TPC_UTILITIES_DETECTORPROPERTIESICARUSCLOCKOFFSETMC_H
#define ICARUSCODE_TPC_UTILITIES_DETECTORPROPERTIESICARUSCLOCKOFFSETMC_H
```

icaruscode/TPC/Utilities/DetectorPropertiesICARUSClockOffsetMC.cxx
```diff
@@ -0,0 +1,483 @@
...
+    double const x_ticks_coefficient = 0.001 * driftVelocity * samplingRate;
+
+    double extraOffset = 0.;
+    if ( clock_data.TriggerTime() > std::numeric_limits<double>::lowest() + std::numeric_limits<double>::epsilon() &&
+         clock_data.TriggerTime() < std::numeric_limits<double>::max() - std::numeric_limits<double>::epsilon()
+    ){
+      extraOffset = clock_data.TriggerTime()-clock_data.BeamGateTime();
+    }
+
+    double const triggerOffset = trigger_offset(clock_data) + clock_data.TPCClock().Ticks(extraOffset);
```
PetrilloAtWork, line 366: Here is where the difference is.
Incidentally, it should be possible to check the existence of the trigger time with an equality comparison, since in case of no trigger the exact value from `numeric_limits` is assigned. But ok, more care won't hurt.

icaruscode/TPC/Utilities/CMakeLists.txt
```diff
@@ -49,6 +49,34 @@ cet_build_plugin(SignalShapingICARUSService art::service LIBRARIES ${icarus_util
...
+  larcoreobj::SimpleTypesAndConstants
+  canvas::canvas
+  messagefacility::MF_MessageLogger
+  fhiclcpp::types
+  fhiclcpp::fhiclcpp
+  cetlib_except::cetlib_except
+  PRIVATE
+  cetlib::cetlib
+  ROOT::Core
+  ROOT::Hist
```
PetrilloAtWork, line 72: It does not look like ROOT is needed.
```suggestion
```

PetrilloAtWork (approved):
I left a few comments because it's me.
It is not clear to me which are all the consequences of this change to the tick offsets; I count on the expertise of the submitters.

brucehoward-physics:
> It is not clear to me which are all the consequences of this change to the tick offsets; I count on the expertise of the submitters.

This is definitely not the optimal way to achieve this affect. That would be to, for example, use what Tony's been putting together. But that applies things at the sim level so not an option here. This works since we are primarily using the TPC for analysis with this service and then shift the ticks in interpreting the hits in reco.

We do not intend to use this updated service at CAFMaking, as it could reduce the quality of the truth match. This means that the slight shifts would be reflected in calibration e.g. for electron lifetime corrections, but should be small effect on the level of shift we're talking about (which should be ~couple of cm at max).

The implications for CRT and PMT systems here (by setting the Trigger Module Label) are not immediately clear to me at least without giving it some thought. But, this is meant to be used for reprocessing for analysis primarily using the TPC system.

Just writing down some of the thoughts to this end.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_72_00 SBNSoftware/sbncode@v09_72_00_06 SBNSoftware/sbnanaobj@v09_20_06_03 SBNSoftware/sbndaq-artdaq-core@v1_06_00of0 SBNSoftware/sbnobj@v09_16_00_02 SBNSoftware/icarus_signal_processing@v09_63_00_02 SBNSoftware/icarusalg@v09_72_00_06 SBNSoftware/icarusutil@v09_66_02

mmrosenberg:
CI builds must be triggered manually to get the old c7/e20 qualifiers that were used to build the larsoft version needed for this production release. The manually triggered CI tests passed. Links to the tests:

https://dbweb8.fnal.gov:8443/LarCI/app/ns:icarus/build_detail/phase_details?build_id=icarus_ci/9974&platform=Linux%20slf7&phase=ci_tests&buildtype=slf7%20c7:prof

https://dbweb8.fnal.gov:8443/LarCI/app/ns:icarus/build_detail/phase_details?build_id=icarus_ci/9975&platform=Linux%20slf7&phase=ci_tests&buildtype=slf7%20e20:prof

PR #690, opened 2024-02-16
stage0/stage1 fhicl updates plus WC mod for memory issue

SFBayLaser:
This updates the stage0 and stage1 fhicl files for handling the various cases to deal with - 1D vs 2D deconvolution, running larcv for both data and MC, etc. This also includes modification to the wirecell deconvolution configuration file to implement the memory management issue encountered previously.

Finally, these updates will keep both 1D and 2D deconvolution output through hit finding - "gaushit1dTPCxx" and "gaushit2dTPCxx" hit collections. 

NOTE: the above could be a breaking change for users who simple use "gaushitTPCxx" for finding hits.


SFBayLaser:
trigger build

PR #697, opened 2024-02-21
Moveing SystTools from icaruscode to sbncode

jedori0228:
Moving `icarsucode/fcl/caf/SystTools` to `sbncode/SBNEventWeight/jobs/SystTools`. The universes are already thrown in fcl, so it is important to have same configs between SBND and ICARUS. In https://github.com/SBNSoftware/sbncode/pull/433 I added back this directory in sbncode.

Also added cafmaker fcls that only runs systtools and not flux, which is appropriate for NuMI samples; for NuMI, we evaluate flux systematics at the analysis level (`sbnana`).


PetrilloAtWork (approved):
Looks good.

mmrosenberg:
trigger build

PR #694, opened 2024-02-18
Feature/howard for nu mi2023 a from gputnam configs

brucehoward-physics:
Cherry-picking in what I think are the relevant commits from Gray's feature branch used in the reprocessing.

A few minor edits where necessary to resolve a cherry-pick conflict may be present.

Actually I notice a few things here that relate to GENIE systematics, e.g. config_GENIEReWeightMultisigma_icarus_v2.fcl -- tagging @jedori0228 in case that is already in a commit he has on our board or this or others aren't needed or etc.

Issue #693, opened 2024-02-16
NuMI XSec 2024 Updated Production Patch: Data calibration changes to "run" from "timestamp"


brucehoward-physics:

brucehoward-physics:
This also comes from Gray's feature branches so marking it as "ready" -- assume we will just want this too.

PR #699, opened 2024-02-24
Use run number for TPC calibration

jedori0228:
- We are passing time stamp to TPC dqdx and YZ calibration DB querying; drift correction module was correctly using run number. 
- Using SQL db as default for TPC calibration


mvicenzi (approved):
Good catch!

mvicenzi:
Does this solve #693 as well?

jedori0228:
> Does this solve #693 as well?

Current develop is same as #694; using PostGres as default, and passing timestamp for YZ and dQ/dX module. We need to cherry-pick [88cc345](https://github.com/SBNSoftware/icaruscode/pull/699/commits/88cc345e687bd41ad15055574d4377a958efc18e) and [2566732](https://github.com/SBNSoftware/icaruscode/pull/699/commits/256673220f4ed64461b594d3124584abd874e75f) there as well. @brucehoward-physics

cfarnese (approved):
Important correction to run data events!

mmrosenberg:
trigger build

PetrilloAtWork (commented):
Seen, nothing looks wrong, I trust the submitter and the other reviewers on the database actually accepting a run number as a "time" parameter instead of a timestamp.

Issue #14, opened 2020-06-13
Update the enable time of PMT simulation


PetrilloAtWork:
The current setting is for 2.8 ms, while the planned one is 2.0 ms.
Need to be careful to set the offsets correctly (it should be starting 1 ms before the beam gate opening time).

SFBayLaser:
Is there an update to this issue?

PetrilloAtWork:
Not really. In fact, I had forgotten about it.
Let's pull in somebody in DAQ, like @wesketchum , to ask if the information that the readout of PMT is enabled from -1 ms to +1 ms with respect to the expected beam gate opening (I can't find Angela Fava user here, or I would bother her too).
My understanding was that at -1 ms the readout board caches are flushed, and 2 ms after that what is in the cache is transferred (if a trigger has happened).

wesketchum:
Good q. I'm not sure what the PMT guys have decided on the timing. It's configurable, so I think it could be more ... I'll try to followup.

SFBayLaser:
Is there an update to this issue?

wesketchum:
Pinging again, and maybe pinging myself. What is needed to close this out?

PetrilloAtWork:
Mostly some clarity about what the readout/DAQ configuration is going to be.
I think we are going to get there with the first commissioning of the multi-window PMT acquisition (i.e. storing interesting PMT waveforms even out of the beam gate), which should happen in one-two months time scale.

PR #702, opened 2024-02-27
Workaround on FHiCL parameter set bug

PetrilloAtWork:
The bug is discussed in issue #14 of art-framework-suite/fhicl-cpp (https://github.com/art-framework-suite/fhicl-cpp/issues/14). The workaround is that `util::readConfigurationFromArtPrincipal` t emits a warning if an expected configuration is not present, instead of throwing an exception. That puts the load on the caller for checking that the configuration that is needed is actually present (in some of the cases we have seen, it was, so no harm done).

This workaround will cover only the trigger and PMT configuration extraction.
Lookup of the configuration is also performed in some standard LArSoft services (`DetectorClocks` and `DetectorProperties`), where no issue has been reported, and in `CAFMaker` of SBNSoftware/sbncode, that this workaround does not cover.
It _is_ possible to port the same type of workaround in `CAFMaker`. On the other end, this will require a new release of `sbncode`, and it may compete in timeline with a new release of `fhiclcpp` and LArSoft stack with the bug fixed, after which the workaround is not necessary any more. So there is a decision to be taken there.

### DIFF
--- a/icaruscode/Utilities/ReadArtConfiguration.h
+++ b/icaruscode/Utilities/ReadArtConfiguration.h
@@ -14,6 +14,7 @@
 // framework libraries
 #include "canvas/Persistency/Provenance/ProcessConfiguration.h"
 #include "canvas/Persistency/Provenance/ProcessHistory.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
 #include "fhiclcpp/ParameterSet.h"
 #include "fhiclcpp/ParameterSetID.h"
 #include "fhiclcpp/ParameterSetRegistry.h" // also defines ParameterSetID hash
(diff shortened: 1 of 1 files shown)


cerati (approved):
works!

PetrilloAtWork:
I _think_ that it is still of interest merging this request, because the issue caused some files to have "corrupted" history saved, and processing those files would still fail.
On the other end, one can also decide that it's too transient and mark the affected data to be junked and reprocessed.

mmrosenberg:
trigger build

PetrilloAtWork:
Although the bug that originated this pull request has been fixed upstream, there are still data files around that are affected by it, and this patch makes the code more robust in my opinion, so I would still like it to be added into a release.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #705, opened 2024-02-29
Feature/howard for nu mi2023 a from gputnam configs plus few extras

brucehoward-physics:
Picking up a few more things Gray pointed out beyond what's in #694 . This should go in after that, as this picks up those commits and adds this.


SFBayLaser (approved):
I will leave my standard lament that variations require separate fhicl files which only serve to make things more challenging to understand. Sigh... well, that shipped sailed many years ago  I suppose.

PR #707, opened 2024-02-29
Enhancement of the CRTPMT Analysis module

francescopoppi:
This Commits adds several addition to the analysis module: the first addition is the implementation of three different timing for the opHits, namely the StartTime, the RiseTime and the PeakTime. The second addition is the inclusion of vector of OpHitsArea, Width and PE. Finally, this commit adds some variables on the Filter Test Tree which reflects the different filtering logic and their possible results.


mmrosenberg:
trigger build

PetrilloAtWork (approved):
Approved.

mmrosenberg:
trigger build

PR #716, opened 2024-03-12
restore legacy G4

cerati:
This PR is meant to restore the legacy G4 functionality (needs SBNSoftware/icarusalg#84).
The main problem is that the legacy G4 does not support multiple sensitive AuxDet volumes corresponding to a single logical volume (see e.g. https://github.com/LArSoft/larsim/blob/develop/larsim/LegacyLArG4/AuxDetReadoutGeometry.cxx#L139) and the refactored gdml uses a single logical volume for all strips in the same module type.

What we do in this PR is to:
* ensure that the volume name lookup works for both legacy and refactored gdml (with and without "_" in the volume names). See changes in icaruscode/CRT and icaruscode/CRT/CRTUtils
* all the other changes are to fcl files:
   * change the default fcls and input labels to the refactored g4, as it should be the new standard
   * ensure the legacy g4 and detsim use the legacy gdml and the legacy input labels
   * ensure the legacy stage0 fcl for MC use the legacy input label for mcophit (using the new gdml is fine for reco as the geometry is the same)
   * comment out mcophit from stage1 fcls as it is not used

What this PR does not do:
* move all the legacy fcls to a separate directory
* rename the refactored fcls to be the new standard (i.e. what production uses)
These should probably be done once there is consensus that we can use refactored for production.

### DIFF
--- a/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
+++ b/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
@@ -496,7 +496,7 @@ namespace icarus{
 	      path += "/";
 	    }
 	  }
-	  if (path.find(adGeo.Name())!=std::string::npos && path.find("strip"+stripn)!=std::string::npos) break;
+	  if (path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) break;
 	  else path = "";
 	}
 
--- a/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
+++ b/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
@@ -488,6 +488,9 @@ namespace icarus{
 	string stripn = std::to_string(adsid);
 	if (stripn.length()==1) stripn = "0"+stripn;
 
+	// whether we are using the "legacy" geometry naming scheme
+	bool const isLegacy = std::string(adsGeo.TotalVolume()->GetName()).find("_")!=std::string::npos;
+
         string path = "";
 	for (size_t ip=0;ip<paths.size();ip++) {
 	  for (size_t inode=0; inode<paths.at(ip).size(); inode++) {
@@ -497,7 +500,7 @@ namespace icarus{
 	    }
 	  }
 	  if ( (path.find(adGeo.Name())!=std::string::npos && path.find("strip"+stripn)!=std::string::npos) || //refactored
-	       (std::string(adsGeo.TotalVolume()->GetName()).find("_")!=std::string::npos && path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) ) //legacy
+	       (isLegacy && path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) ) //legacy
 	    break;
 	  else path = "";
 	}
--- a/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
+++ b/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
@@ -496,7 +496,7 @@ namespace icarus{
 	      path += "/";
 	    }
 	  }
-	  if (path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) break;
+	  if ( (path.find(adGeo.Name())!=std::string::npos && path.find("strip"+stripn)!=std::string::npos) || path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) break;
 	  else path = "";
 	}
 
--- a/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
+++ b/icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
@@ -496,7 +496,9 @@ namespace icarus{
 	      path += "/";
 	    }
 	  }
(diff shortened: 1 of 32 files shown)


cerati:
The intended behavior of fcl changes has been verified using @jzennamo's `fcl_check.sh` script. Thanks Joseph!

icaruscode/CRT/crtsimmodules_icarus.fcl
```diff
@@ -85,7 +85,7 @@ icarus_crtsim:
 {
   module_type: "icaruscode/CRT/CRTDetSim"
 
-  G4ModuleLabel: "largeant"
+  G4ModuleLabel: "genericcrt"
```
jzennamo, line 88: Is this the intended behavior?

jzennamo (approved):

icaruscode/CRT/crtsimmodules_icarus.fcl
cerati: yes, this is part of making refactored G4 the default in the upstream fcl files. 

Here the confusion may be that what is referred as "G4ModuleLabel" is not the G4 module in the refactored scheme.

cerati:
> My reading is that this makes "refactored" closer to the default handling, is that true?
Yes. And at the same time revives the legacy

icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
```diff
@@ -496,7 +496,9 @@ namespace icarus{
 	      path += "/";
 	    }
 	  }
-	  if (path.find(adGeo.Name())!=std::string::npos && path.find("strip"+stripn)!=std::string::npos) break;
+	  if ( (path.find(adGeo.Name())!=std::string::npos && path.find("strip"+stripn)!=std::string::npos) || //refactored
+	       (std::string(adsGeo.TotalVolume()->GetName()).find("_")!=std::string::npos && path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) ) //legacy
```
PetrilloAtWork, line 500: Isn't whether the volume name contains an underscore known since the beginning?
I think's is faster and more readable to cache that information:
```cpp
  // whether we are using the "refactored" geometry naming scheme
  bool const isRefactored = std::string(adsGeo.TotalVolume()->GetName()).find("_") != std::string::npos;
```
```suggestion
	       (isRefactored && path.find(adsGeo.TotalVolume()->GetName())!=std::string::npos) ) //legacy
```

cerati: Thanks for the suggestion. I assume the variable should be called "isLegacy" rather than "isRefactored" thoughj

cerati:
> I would have preferred the definition of a complete geometry legacy configuration in `geometry_icarus.fcl`, instead of
> 
> ```
> services.Geometry.GDML: "icarus_complete_20220518_overburden.gdml"
> services.Geometry.ROOT: "icarus_complete_20220518_overburden.gdml"
> ```

Looks like there is already a "legacy" geometry in `geometry_icarus.fcl`, so to avoid confusion I would not make another one. Also, the refactored is not really a different geometry, it's just a different gdml.

SFBayLaser (approved):
I am mildly concern about merge conflicts with other PRs that are also making fcl changes... but I guess we will see what happens and believe git will do the right thing in merging. 
Its likely too late now but I wonder if appending "refactored" to the fcls was the right choice. I was looking at a list of fcl files and wondering what "refactored" meant...

cerati:
Right, we should have a followup PR with a cleanup of the fcls...

icaruscode/CRT/CRTUtils/CRTDetSimAlg.cc
PetrilloAtWork: Possible. I am lost with respect to which has `"_"` and which does not.

cerati: Legacy has "_"

PetrilloAtWork:
> > I would have preferred the definition of a complete geometry legacy configuration in `geometry_icarus.fcl`, instead of
> > ```
> > services.Geometry.GDML: "icarus_complete_20220518_overburden.gdml"
> > services.Geometry.ROOT: "icarus_complete_20220518_overburden.gdml"
> > ```
> 
> Looks like there is already a "legacy" geometry in `geometry_icarus.fcl`, so to avoid confusion I would not make another one. Also, the refactored is not really a different geometry, it's just a different gdml.

I think it's time to retire that legacy configuration. Also, a new one may have a more descriptive name (`geometry_service_GEANTrefactoring`?).

cerati:
> > > I would have preferred the definition of a complete geometry legacy configuration in `geometry_icarus.fcl`, instead of
> > > ```
> > > services.Geometry.GDML: "icarus_complete_20220518_overburden.gdml"
> > > services.Geometry.ROOT: "icarus_complete_20220518_overburden.gdml"
> > > ```
> > 
> > 
> > Looks like there is already a "legacy" geometry in `geometry_icarus.fcl`, so to avoid confusion I would not make another one. Also, the refactored is not really a different geometry, it's just a different gdml.
> 
> I think it's time to retire that legacy configuration. Also, a new one may have a more descriptive name (`geometry_service_GEANTrefactoring`?).

that can be done, but in a followup PR once we agree what needs to be preserved in fcls in release

mmrosenberg:
trigger build SBNSoftware/icarusalg#84

PR #721, opened 2024-03-25
Two updates: bug fix for keep up processing, adjust interplane time offsets and decon timing for truth matching

SFBayLaser:
1) A bug was reported in the keepup processing fhicl file that caused a loss of purity monitor output. This was because the input RawDigit data product names were wrong, set back to default.
1) Adjusted the interplane time offsets and the overall deconvolution offset (in the finall inverse FFT) to get simulation and reconstruction times to match for truth matching.


cfarnese (approved):
I take a look in particular to the part related to the purity and it seems to me good!

mmrosenberg:
trigger build

PR #722, opened 2024-03-27
Script maintenance

PetrilloAtWork:
This is a composite pull request with maintenance of some scripts:
 * removed obsolete overburden-related helper scripts
 * converted `runFilesFromSAM.py` into Python (it was Bash), now way faster
 * `sortDataLoggerFiles.py` to support ICARUS data file new name convention (note that *this script has also duplicate data file filtering functionalities*)
 * updated `manageDataRunDefinitions.py` to support minimum file size and more (in intermediate commits it is called with an old name `createDataRunDefinitions.py` by mistake)

Reviewers:
* @SFBayLaser (wildcard)
* @mvicenzi from DAQ (feel free to delegate to somebody else)


mvicenzi (approved):
These are very useful and should be vastly advertised!

Semi-related question: no script to check the file(s) locality (disk or tape or both)?
In my workflows I do it while or after building the list so I know whether to launch prestaging or not.

mmrosenberg:
trigger build

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #727, opened 2024-04-02
Internal fix to PMT simulation configuration

PetrilloAtWork:
This is a minor fix to the PMT simulation configuration.
Some noise tool configurations were not correctly referenced but rather bypassed.
And one of the bypass had an obsolete tool name (so the configuration without electronics noise was effectively broken).

I have verified that `detsim_1d_icarus.fcl` in `develop` and on this branch come out the same (i.e. this is just an internal fix).

Reviewers: two people more or less informed on ICARUS PMT simulation.


mmrosenberg:
trigger build

mvicenzi (approved):
This is just a refactoring of the configuration. Approving!

amenegol:
Ok for me. Approved.

mmrosenberg:
trigger build

PR #730, opened 2024-04-12
More updates to enable detector variation MC samples

jzettle:
This PR is meant to update the work on adding detector variation fcls to its most recent form that I have been working with and testing. It adds support for the refactored G4 version which I had not been previously using to generate samples so it has been lightly tested with icaruscode v09_87_00 to the degree that it completes through the procedure for a very small sample that has been generated with the refactored LArG4 but further support to validate things would be incredibly welcome. I had been using the legacy LArG4 in the samples that I have produced for the initial work and that was difficult to move away from without needing to repeat the entirety of the previous work. I have also validated that the variation indeed changes something to the level of checking fhicl-dump and what art is actually running during the process but any low-level validation checks are also useful to really confirm that something is different, which is not something I always know how to do myself.

The reviewers I chose here are more for informing a set of people about this PR to check that I have based things on the correct set of initial fhicls for the refactored LArG4 and if there are any comments on the path forward.

In order to run this we need to have saved the stage1 files from a given production with the g4 stage data products. The idea here is this splits the g4 stage into 2 steps to reduce randomness but we may not be able to run it like this if something needs to be launched imminently.

Comments welcome and please reach out to me if needed.


mmrosenberg:
trigger build

cerati (approved):
I reviewed the usage of "refactored" fcl files and it looks correct to me.

SFBayLaser (approved):
I continue to yearn for a way to do this without the massive proliferation of fcl files...

mrmooney (approved):
Detector variations to be carried out were discussed offline, and determined to be fine

mmrosenberg:
trigger build

mmrosenberg:
trigger build

PR #732, opened 2024-04-27
Update for PMT gain calibration tools

mvicenzi:
This PR contains a set of updates to the gain estimation tools in `icaruscode/PMT/Calibration`. 

These tools allow to extract the PMT gains using "background" optical hits in the single photo-electron range from `offbeam{bnb,numi}minbias` waveforms. This update improves both the selection and the fitting procedure to make the estimation more robust.

- `PMTBackgroundphotonsCalibration_module.cc` now allows time-filtering to exclude train of optical hits that follow large signals. Histogram binning can now be tuned from fhicl file. The entire collection of selected optical hits is now also saved in the output file: this is needed to apply an amplitude cut on the ophits before fitting the charge distribution.
- New fitting functions have been added in `CaloTools`: a parent class (`SPEChargeResponse`) and two derived classes with the actual implementations (`SPEChargeIdealResponse` and `SPEChargeLogisticResponse`).
- The fitting manager `FitBackgroundPhotons.cc`has been updated to handle the new functions and allow better flexibility in setting the initial fit parameters. It can now also save the fit results in a .pdf file.
- Large revamping of the fitting macro `fitPulseDistribution.cc` with plenty of control via input parameters. The default procedure now uses the amplitude histogram to define a channel-by-channel threshold on the optical hits that contribute to the charge distribution. The new charge histogram is fitted by convoluting the ideal SPE charge response with a logistic function. The parameters of the logistic function are extracted and fixed by the ratio between the charge distribution before and after the threshold cut. This method appears to be more robust than relying on hard-coded fit ranges to remove noise pedestals. If needed, the legacy behavior is still available through the appropriate input arguments.


mmrosenberg:
trigger build

mmrosenberg:

mmrosenberg:
trigger build

cfarnese (approved):
The code seems to me ok! I tested it and it compiles properly and it runs correctly! I suggest to merge this code into develop!

PR #733, opened 2024-04-29
larg4_icarus.fcl update for ML

yeonjaej:
larg4_icarus.fcl from previous PR https://github.com/SBNSoftware/icaruscode/pull/731/ seems to have been overwritten by other PRs during the merge for the current release. This is the correct config for ML production.

PR #734, opened 2024-04-29
Fix ML chain for ICARUS systematics and standard productions

jzettle:
This PR is meant to try to fix the issue seen in v09_89_01 with Supera running when attempting to test things for the variation. Also includes small changes needed for ML running in a couple of the variation fcl. This PR supersedes https://github.com/SBNSoftware/icaruscode/pull/733 and was made in coordination with @yeonjaej ensuring I had included the same changes she did in her PR. Making this against develop as that and the v09_89_01 tag are identical according to github.

Still testing this with the neutrino-only samples to make sure the issue is fixed to the level we can quickly. Also notifying @cerati and @SFBayLaser about this. Will comment results of testing over the rest of the night/tomorrow morning and we can discuss how to patch the release if everything is successful. 

Apparently I cannot request @yeonjaej as a reviewer...?

Relies also on the sbncode PR built against the tag v09_89_01 in sbncode: https://github.com/SBNSoftware/sbncode/pull/460

### DIFF
--- a/fcl/g4/larg4_icarus.fcl
+++ b/fcl/g4/larg4_icarus.fcl
@@ -91,11 +91,11 @@ physics.producers.largeant.StoreDroppedMCParticles: false
 # ------------------------------------------------------------------------------
 # Configure mcreco to read SEDLite instead of SED and MCParticleLite in addition to MCParticle
 physics.producers.mcreco.G4ModName: "simdrift"
-physics.producers.mcreco.SimChannelLabel: "simdrift" #"sedlite"
+physics.producers.mcreco.SimChannelLabel: "sedlite"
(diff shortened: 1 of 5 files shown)


SFBayLaser (approved):
I am not in favor of keep ALL output products at stage 0 in "stage0_variation_icarus_savewires.fcl" but we have not time to do this correctly right now. The rest seem to be fine.

jzettle:
That one was my bad, I thought I had to change something to save the wire information in the calibration tuples to validate some of the variations and found it easier with the complete time crunch to save everything from stage0 instead of testing which product from stage0 was the correct one to not drop. Happy to review and update this later once there is more time

mmrosenberg:
trigger build SBNSoftware/sbncode#460

PR #736, opened 2024-05-02
Updates for Doxygen documentation generation

PetrilloAtWork:
* LArSoft has moved to a different server, and since we somehow link to their "tag file", we need an update 
* ROOT tag file is now compressed (Gzip), but (my) Doxygen does not support that: the documentation update script now decompresses tag files
* bug fix (it seems Alma Linux 9 Bash is more picky than Scientific Linux 7)


mmrosenberg:
trigger build

SFBayLaser (approved):
There is nothing that stands out to me here.... so good to go from my perspective.

mmrosenberg:

PR #739, opened 2024-05-21
Set SaveGENIEEventRecord as true be default cafmaker fcl

jedori0228:
From https://github.com/SBNSoftware/sbncode/pull/459, which is merged to develop, I added a feature to save GENIE event record to the CAFMaker output. Here I set the default value of a boolean, `SaveGENIEEventRecord`, to `true`; this is crucial to constantly adopt new cross-section systematics without re-running CAFMaker job. More importantly, we now stopped updating [LarSoft/nusystematics](https://github.com/larsoft/nusystematics) but moved to [NuSystematics/nusystematics](https://github.com/NuSystematics/nusystematics), and there are some missing dials that we cannot currently evaluate from `lar` based jobs. We can add those once we have GENIE record saved together with StandarRecord.

### DIFF
--- a/fcl/caf/cafmaker_defs.fcl
+++ b/fcl/caf/cafmaker_defs.fcl
@@ -330,6 +330,7 @@ cafmaker.TriggerLabel: "emuTrigger"
 cafmaker.FlashTrigLabel: "" # unavailable
 cafmaker.SimChannelLabel: "largeant"
 cafmaker.SystWeightLabels: ["genieweight", "fluxweight"]
+cafmaker.SaveGENIEEventRecord: true # save GENIE event record by default. Turn this off for data cafmaker fcl
 cafmaker.TPCPMTBarycenterMatchLabel: "tpcpmtbarycentermatch"
 cafmaker.TrackHitFillRREndCut: 30 # include entire PID region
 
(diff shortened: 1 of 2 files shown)


mmrosenberg:

jzettle (approved):
This sounds like a positive addition to me! I approve if this helps with getting the right systematics in place from the NuSystematics tools

brucehoward-physics:
As is known, I love this addition!

But, I do have a question. How much does this increase the file sizes @jedori0228 ?

Just wondering if we want to turn it on by default or have it as an option with a separate CAF that production requests could use.

jedori0228:
> As is known, I love this addition!
> 
> But, I do have a question. How much does this increase the file sizes @jedori0228 ?
> 
> Just wondering if we want to turn it on by default or have it as an option with a separate CAF that production requests could use.

I forgot that I have a supporting document related to this PR: https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=36869

The increment in the file size was 0.5% (when compared to a CAF that has no weights).

brucehoward-physics:
That seems pretty small given that CAFs are already smaller. Approved.

brucehoward-physics (approved):

henrylay97 (approved):
Great stuff!

mmrosenberg:
trigger build

PR #740, opened 2024-05-22
Trigger decoder update to unpack the new trigger diagnostics

PetrilloAtWork:
Changes in the trigger information conveyed via DAQ fragments are documented in [SBN DocDB 36264](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=36264) and link therein.

The new information needs to be extracted and stored into a data product (see SBNSoftware/sbnobj#110 for the data product extension).
These changes enable the ICARUS trigger decoder tool V3 to decode the new format of data (whose version number is by chance also `3`).

## Changes

The changes all pertain the cryostat-specific information about global trigger candidates in that cryostat; this information lands on a `sbn::ExtraTriggerInfo::CryostatInfo` object. The information of each cryostat is treated independently from the other. The changes are the following:
 * LVDS bits (including the adders'), the cryostat trigger count and the trigger logic are now unpacked for a cryostat also for minimum bias events; in fact, unpacking happens if either:
     * There is a global trigger in the cryostat (as will be reported in `triggerLocationBits`); this happens only for light-based triggers, not minimum bias ones.
     * The `Flag_XXXX` datum is `1` for that cryostat; this may happen in minimum bias events too, and should always happen when there is a global trigger in the cryostat. However, in rare cases (< 0.01%) it was observed that there is no trigger location (`triggerLocationBit` `0`) while there was obviously a trigger — the event _was_ acquired, after all. In those cases, the `Flag_XXXX` was still set and the decoding should still happen.
 * based on `Flag_XXXX`, the content of `Delay_XXXX` is converted into nanoseconds via the **new configuration parameter `FPGAtickLength`** (defaulting to `25` ns to reflect the settings of the hardware since its installation) and stored into the new data member `sbn::ExtraTriggerInfo::CryostatInfo::beamToTrigger`.

As a remainder, all this information pertains the first trigger candidate, with the exception of the count which records how many more candidates arrived after the first one. The only reason why the first candidate of a cryostat is not be promoted to global trigger is that a global trigger has already happened, either from the other cryostat or because of a minimum  bias gate.

## Backward compatibility and testing

The new decoder was successfully tested with old data which does not include the new information, and with new data which does, for both minimum bias and light-based trigger events.
The old decoder was also tested with the new data, in which case it correctly decodes the available information and ignores the new one, which is left default-constructed (like there were no trigger in the cryostat). The old code also does not decode anything of the cryostat information for minimum bias events.

All changes to the fragment data format and to the offline code (data product and decoding tool) are **non-breaking both backward and forward**.

## Reviewers

I summon for review one among:
* @jzettle, maintainer of the trigger decoding tool;
* @dtorretta56, author of the upstream changes;
* @mvicenzi, as DAQ expert.

### DIFF
--- a/icaruscode/Decode/DecoderTools/TriggerDecoderV3_tool.cc
+++ b/icaruscode/Decode/DecoderTools/TriggerDecoderV3_tool.cc
@@ -196,6 +196,11 @@ namespace daq
    *                and <C1P3> = `0000'0nnn`
    *            * west wall: `0000'0000'00ss'snnn` with <C2P3> = `0000'0sss`
    *                and <C3P3> = `0000'0nnn`
+   *        * `beamToTrigger`: nanoseconds from the beam gate opening to the
+   *            first trigger in the cryostat. The value is converted from ticks
+   *            (`FPGAtickLength`) and is set to
+   *            `sbn::ExtraTriggerInfo::NoTrigger` (a value larger than the beam
+   *            gate duration) if no trigger is available in the cryostat.
    *     
    *     Information may be missing. If a count is not available, its value is
    *     set to `0` (which is an invalid value because their valid range starts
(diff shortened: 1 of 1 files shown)


mvicenzi (approved):
Looks good to me.

dtorretta56:
OK to proceed in implementing this.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_90_00 SBNSoftware/sbnobj#110

jzettle (approved):

mmrosenberg:
trigger build

mmrosenberg:
trigger build SBNSoftware/sbnobj#110

PetrilloAtWork:
trigger build

PetrilloAtWork:
Pleeease merge this PR into release.

PR #741, opened 2024-05-24
Fix wrong comment in PMTsimulationAlg

PetrilloAtWork:
I just got fed up of reading that wrong information, probably because it was me to write it.

So the photon library does not include any quantum efficiency; the quantum efficiency is applied before writing the scintillation photons (`sim::SimPhotons[Lite]`) in `G4` stage.

Reviewer is who picks the shortest straw... @amenegol.


mmrosenberg:

mmrosenberg:
trigger build

PR #744, opened 2024-06-14
Updated fhicls for the reprocessing release

brucehoward-physics:
These fhicls are slight modifications (as NEW FHICLS) to the rereco (stage1+cafmaker joined) fhicls in the release already. I believe these are going to be the fhicls we want to run, so once these are in I think we can cut a tag and proceed to test things on the grid. If it turns out other slight modifications are needed, we could put a patch and/or submit a custom fhicl to the grid nodes? (This is aimed at Matt)

For the reviewer: the changes are intended basically to
(a) turn on saving the art root output -- we'll possibly need to loop back through when Geant4Reweight is set up to remake the CAFs, so having "stage1" output is useful.

(b) Save more of the calo points to the CAFs

(c) turn off the T0 thing in the calorimetry in a few places. Since T0 tags should only exist for clear cosmics, I don't think this should impact much in the analysis itself (though I think at one point these were being used to try to validate some things with cosmic muons)

I did see that I could actually run these fhicls on a couple of MC files and a data file from Run2, didn't try a Run1 or off-beam. We might run cafmaker with the current version of G4Reweight, so these fhicls do not modify that part of it from the way it was, even though those are not becoming part of this v09_72_00_07 tag. It's possible we'd want a v09_72_00_07pXX when that is updated.

Adding @PetrilloAtWork and @jzettle as reviewers in hopes that one of them has a moment to take a look. And tagging @jedori0228 in case he wants to add any comments or realizes something is missing. And tagging @mmrosenberg as release manager. Thanks all!


jzettle (approved):
Nothing looks incorrect to me and these are standalone fcls so it does not affect the default running conditions

jzettle:
So to comment on one part of your other question: at least you can submit custom fhicls to the production jobs. We had to do this for the BNB variation samples as there was an issue that I would have never seen with the variation "scrub" stage because it only came up when Mateus tried to work with the metadata from those jobs so this is an option

fcl/caf/cafmakerjob_icarus_data_rereco_numi_Run1_ReprocRelease.fcl
```diff
@@ -0,0 +1,28 @@
+#include "cafmakerjob_icarus_data_rereco_numi.fcl"
+
+services.Geometry: @local::icarus_geometry_services_no_overburden_legacy_icarus_v3.Geometry
```
PetrilloAtWork, line 3: Why no-overburden? Run1 had the overburden.
I am not sure it matters, since as far as I know the overburden is only used by GEANT4.

fcl/caf/cafmakerjob_icarus_data_rereco_numi_Run1_ReprocRelease.fcl
```diff
@@ -0,0 +1,28 @@
...
+         "drop *_caloskimCalorimetryCryoW_*_*"
+      ]
+      saveMemoryObjectThreshold: 0
+   }
+}
+
+physics.producers.cafmaker.TrackHitFillRREndCut: 9999999
+
+physics.producers.pandoraCaloGausCryoE.T0ModuleLabel: ""
+physics.producers.pandoraCaloGausCryoW.T0ModuleLabel: ""
```
PetrilloAtWork, line 28: Why did you disable it?
From your comments, it sounds like it does not help; does it hurt?

fcl/caf/cafmakerjob_rereco_numi_icarus_systtools_and_fluxwgt_ReprocRelease.fcl
```diff
@@ -0,0 +1,79 @@
...
+
+services:
+{
+  TFileService:           { }
+  IICARUSChannelMap:      @local::icarus_channelmappinggservice
+                          @table::icarus_wirecalibration_services
+  SpaceChargeService: @local::icarus_spacecharge
+  IFBeam: {}
+}
+services.ParticleInventoryService:  @local::standard_particleinventoryservice
```
PetrilloAtWork, line 15: Aestetics: you can merge this in the `services` table above. Easier to read.

fcl/caf/cafmakerjob_rereco_numi_icarus_systtools_and_fluxwgt_ReprocRelease.fcl
```diff
@@ -0,0 +1,79 @@
...
+      outputCommands: [
+         "keep *_*_*_*",
+         "drop *_caloskimCalorimetryCryoE_*_*",
+         "drop *_caloskimCalorimetryCryoW_*_*"
+      ]
+      saveMemoryObjectThreshold: 0
+   }
+}
+
+physics.producers.cafmaker.TrackHitFillRREndCut: 9999999
```
PetrilloAtWork, line 79: What about calorimetry T0 corrections?

PetrilloAtWork (commented):
Might have been better to create a subdirectory with these special configurations.
Just saying.

Left some questions around.

fcl/caf/cafmakerjob_icarus_data_rereco_numi_Run1_ReprocRelease.fcl
brucehoward-physics: I believe was seeing association errors with it. There might be a way around them but since I don't think it is necessary to have, I just gave the empty string.

fcl/caf/cafmakerjob_icarus_data_rereco_numi_Run1_ReprocRelease.fcl
brucehoward-physics: This actually comes from the fhicl I copied in this reprocessing release (same name but without the `_ReprocRelease`), but in any case this fhicl is for data reprocessing: _hopefully_ no_overburden vs with overburden wouldn't matter

fcl/caf/cafmakerjob_rereco_numi_icarus_systtools_and_fluxwgt_ReprocRelease.fcl
brucehoward-physics: Mmm can double check, but I guess this didn't throw the same errors.

brucehoward-physics:
Adding comments from Slack thread RE Geometry:

```
This came from Gray's initial reprocessing and is in the fhicl I copied then edited to make that one. I assume it's because the Run 1 data was processed via keepup in v09_65 or something like this and maybe that's before the v4 geometry? But I'm not positive so if that's not true then definitely let me know!

I think in the GitHub you also asked about the no_overburden part. Again this comes from Gray, but this fhicl is for data reprocessing, hopefully no_overburden vs with overburden wouldn't matter
```

fcl/caf/cafmakerjob_rereco_numi_icarus_systtools_and_fluxwgt_ReprocRelease.fcl
brucehoward-physics: I don't see where this one lists a specific module for T0 while the data versions did? 🤔 That could be why it was complaining in one and not the other. 😅 Not 100% sure why it was complaining though. Tagging @gputnam in case they can comment, but i'm not sure it matters?

brucehoward-physics:
Just added the missing bit to turn on the trigger time shift for MC stage1 redo. Oops. Set up a few files to test.

brucehoward-physics:
Updated upon first testing finding a problem -- forgot to feed the right modules for each separated job. Now finishing second test run of MC files and passed location of files to @jedori0228 to look at.

jedori0228:
I'm validating the fhicls by running test files and comparing them to Gray's reprocessed files. I found few parts neet to be updated when validating MC fhicls:  https://github.com/SBNSoftware/icaruscode/compare/feature/howard_NuMI2023A_reprocFhicl...feature/howard_NuMI2023A_reprocFhicl_more?expand=1
I'll also to the similar checks on data, and then will merge this branch into this PR

PetrilloAtWork (approved):
I think I dropped all the possible issues I have spotted, but I can't usually tell whether they are real problems or not.
Given that you have paid attention on them, I trust you also found the appropriate way forward.

mmrosenberg:
CI tests were manually triggered with c7 and e20 qualifiers and the following revisions:

"LArSoft/larcorealg@v09_10_00 LArSoft/larcoreobj@v09_07_00 LArSoft/lardataalg@v09_13_04 LArSoft/lardataobj@v09_14_00 LArSoft/larfinder@v09_00_01 LArSoft/larvecutils@v09_02_00 LArSoft/lar*@LARSOFT_SUITE_v09_72_00_01 SBNSoftware/sbndaq-artdaq-core@v1_06_00of0 SBNSoftware/sbncode@v09_72_00_07 SBNSoftware/sbnanaobj@v09_20_06_04 SBNSoftware/sbnobj@v09_16_00_03 SBNSoftware/icarus_signal_processing@v09_63_00_02 SBNSoftware/icarusalg@v09_72_00_06 SBNSoftware/icarusutil@v09_66_02 SBNSoftware/icaruscode#744"

Progress can be tracked at these links:

e20:
https://dbweb8.fnal.gov:8443/LarCI/app/ns:icarus/build_detail/phase_details?build_id=icarus_ci/10169&platform=Linux%20slf7&phase=setup_environment_cvmfs&buildtype=slf7%20e20:prof

c7:
https://dbweb8.fnal.gov:8443/LarCI/app/ns:icarus/build_detail/phase_details?build_id=icarus_ci/10168&platform=Linux%20slf7&phase=setup_environment_cvmfs&buildtype=slf7%20c7:prof

mmrosenberg:
The manually triggered CI tests were successful (see previous comment)

PR #749, opened 2024-07-03
Fixes from Clang compilation errors

PetrilloAtWork:
Clang 14 (`c14:prof`) complains about a few lines of code.
While the merit of the complains is arguable (not plain wrong, nor right), "fixing" the code is cheap, and needed if we want to compile with Clang.
The fixes are simple and technical.

Reviewers: who is the Collaboration C++ expert?

### DIFF
--- a/icaruscode/Decode/ChannelMapping/Legacy/ChannelMapPostGres_tool.cc
+++ b/icaruscode/Decode/ChannelMapping/Legacy/ChannelMapPostGres_tool.cc
@@ -29,7 +29,7 @@
 #include "fhiclcpp/types/TableFragment.h"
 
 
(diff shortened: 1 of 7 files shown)


SFBayLaser:
If you translate from C++ to FORTRAN II then I will be happy to review.

mmrosenberg:
trigger build SBNSoftware/icarusalg#87

PetrilloAtWork:
Forced push to update this branch to `master` (`develop` is a bit too ahead and is currently failing compilation for other reasons).
I have added fixes from a few new warnings from Clang.
I have also fixed what I think is a *bug* in the code, which made the relevant unit test fail above but only under Clang (I am not exactly sure why, and some undefined behaviour is certainly involved).

PR #750, opened 2024-07-17
Add default fcl params to gnocchi table

absolution1:
Add in missing prameters to the gnocchi fcl block


mmrosenberg:
trigger build SBNSoftware/sbncode#469

PR #757, opened 2024-10-05
Remove unused TTTResetDelay, improve documentation

mvicenzi:
This PR cleans up the PMT decoder by removing an unused and confusing option for correcting the cable delays in the propagation of the PPS signals that reset the TTT counters. This option is a leftover of an old implementation in which the delays were supposed to be read from board-by-board `TTTResetDelay` values in `CAEN_V1730_setup_icarus.fcl`. 

The reasons for its removal are several:
1. This archaic option was never finalized or used.
2. The TTT reset delay corrections are now handled in a much better way via a calibration database. 
3. It's confusing to have two places in the code where the TTT reset delays can be inserted.

The proposed changes completely remove any reference to `TTTResetDelay` from the decoder. In addition, some of the documention is updated to better reflect the waveform timestamping process.

---

**NOTE:**
The standard timestamping procedure (in place since before Run-1) does not require to explicitly correct for these TTT reset delays. By using digitized copies of the global trigger signal to determine the position of the global trigger timestamp in the PMT readout buffers, all synchronization offsets due to delays in the PPS distribution are automatically taken care of.

However, if time corrections based on the digitized trigger signal are not available, a backup option still exists to use the TTT reset delays directly: in this case, the delays are fetched from the `pmt_cable_delays` database by the `IPMTTimingCorrectionService`.

### DIFF
--- a/icaruscode/Decode/DaqDecoderICARUSPMT_module.cc
+++ b/icaruscode/Decode/DaqDecoderICARUSPMT_module.cc
@@ -293,13 +293,11 @@ namespace icarus { class DaqDecoderICARUSPMT; }
  *   in the same TAI scale as the global trigger; the difference between TTT
  *   and the global trigger pins down the end of the readout buffer in the
  *   electronics time scale;
- * 3. a delay is subtracted to the timestamp, which encompasses all fixed delays
- *   occurring in the trigger signal transportation; the most prominent is the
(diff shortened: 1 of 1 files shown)


mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_91_02

PR #760, opened 2024-10-15
Bug fix to CRT PMT Matching producer

francescopoppi:
This Commit fixes a bug that we found in the CRT PMT Matching produce…. The classification was evaluated on the wrong flash time (it used the first op hit peak time as a reference. This is wrong, one could think of using the first op hit start time, but definetly not the first op hit peak time. Using the flash time is also ok, and the important thing is that this is now consistent with the evaluation of the CRT-PMT time differences.


aheggest (approved):
This fix looks good to me!

francescopoppi:
@aheggest Added the two corrections you asked.

PetrilloAtWork (approved):
It looks good, thank you.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_92_00

PR #786, opened 2024-12-10
Removal of the requirement of a certain number of opHit to be above a given threshold

francescopoppi:
As discussed here: https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=39068
There was one more issue which needed to be addressed in the CRTPMT matching data product.
The requirement of a specific number of opHits to be above threshold does have some reasoning behind and it is a good way to reject fake flashes, but this is a decision that the analyzers should do, not the data product.
I just comment this part because for our current production we can survive with skipping this requirement.
I probably will follow with opening an issue about the addition of the number of op hit above threshold in the data product (sbnobj) and in the Standard Record (sbnanaobj). At that point I will address this once again in the CRTPMTMatching Producer.

### DIFF
--- a/icaruscode/CRT/CRTPMTMatchingProducer_module.cc
+++ b/icaruscode/CRT/CRTPMTMatchingProducer_module.cc
@@ -582,12 +582,14 @@ namespace sbn::crt {
           << firstOpHitPeakTime << " us, centroid: " << flash_pos << " cm";
       }
       
+      /*
       if (nPMTsTriggering < fnOpHitToTrigger) {
         mf::LogTrace("CRTPMTMatchingProducer")
           << "  => skipped (only " << nPMTsTriggering << " < " << fnOpHitToTrigger
           << " hits above threshold)";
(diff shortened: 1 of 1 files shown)


PetrilloAtWork (approved):
If this is the way you want to go, it's ok and approved.
A different way is to _configure_ the module to require "at least 0" triggering PMT.

mmrosenberg:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_89_01 SBNSoftware/sbndaq-artdaq-core@v1_09_00of1 SBNSoftware/sbncode@v09_89_01_02 SBNSoftware/sbnanaobj@v09_23_00_02 SBNSoftware/sbnobj@v09_19_00_01 SBNSoftware/icarus_signal_processing@v09_88_00_02 SBNSoftware/icarusalg@v09_89_01_01 SBNSoftware/icarusutil@v09_88_00_02

PR #787, opened 2024-12-10
Removal of the requirement of a certain number of opHit to be above a given threshold - Develop

francescopoppi:
As discussed here: https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=39068
There was one more issue which needed to be addressed in the CRTPMT matching data product.
The requirement of a specific number of opHits to be above threshold does have some reasoning behind and it is a good way to reject fake flashes, but this is a decision that the analyzers should do, not the data product.
I just comment this part because for our current production we can survive with skipping this requirement.
I probably will follow with opening an issue about the addition of the number of op hit above threshold in the data product (sbnobj) and in the Standard Record (sbnanaobj). At that point I will address this once again in the CRTPMTMatching Producer.

### DIFF
--- a/icaruscode/CRT/CRTPMTMatchingProducer_module.cc
+++ b/icaruscode/CRT/CRTPMTMatchingProducer_module.cc
@@ -581,14 +581,14 @@ namespace sbn::crt {
           << ") cm [" << hits.size() << " op.hits] -> first time: "
           << firstOpHitPeakTime << " us, centroid: " << flash_pos << " cm";
       }
-      
+      /*
       if (nPMTsTriggering < fnOpHitToTrigger) {
         mf::LogTrace("CRTPMTMatchingProducer")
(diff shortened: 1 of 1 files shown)


PetrilloAtWork:
Twin of #786 .

PetrilloAtWork (approved):
Same comment as for #786 .

jas1005:
trigger build larsoft@v10_24_00 LArSoft/lar*@LARSOFT_SUITE_v10_24_00 SBNSoftware/sbnalg@v10_24_00 SBNSoftware/sbnobj@v10_23_00 SBNSoftware/sbnanaobj@v10_23_00 SBNSoftware/sbndaq-artdaq-core@v1_10_06_01 SBNSoftware/sbncode@v10_24_00 SBNSoftware/icarusalg@v10_24_00 SBNSoftware/icarusutil@v10_24_00 SBNSoftware/icarus_signal_processing@v10_24_00

PR #803, opened 2025-02-11
Adding CAF Fcl for for Cosmics

jzennamo:
Currently, if you try to use CAF Maker on cosmic MC, it breaks because it tries to build the GENIE record. 

The simplest solution is to add a new CAF maker fcl that disables the GENIE Event Record storage. 

This is needed by the CI


jedori0228 (approved):

PetrilloAtWork (approved):
Looks good.

PR #809, opened 2025-03-07
remove legacy G4 files, introduce standard fcl directory

cerati:
This PR does 3 things:
1) removal of the legacy G4 fhicl files
2) creating of a "standard" folder pointing to the standard production fhicl files
3) centrally suppress warning messages from SimDriftElectrons, which currently overwhelm log files

Format of fcl files in the standard directory is:
`standard_${data/mc}_${workflow}_${step}_icarus.fcl`


SFBayLaser (commented):
I can't go through every file but I did not see any that were deleted that looked like a  mistake. 
One question: should we remove "refactored" from fcl names since that won't have meaning going forward?

cerati:
eventually yes, but for now I would not do it to avoid breaking CI tests and current production configs. Once they all switch to the standard fcls it will be easier to rename

leoaliaga:
Hi @SFBayLaser, @jzennamo: Tracy’s question about the fcl naming was addressed, and there don’t seem to be any concerns about the files being removed. Are we ready to approve this PR?

SFBayLaser (approved):
Yes, I believe this is ready to go

leoaliaga:
Trigger build

PR #812, opened 2025-03-20
Python interfaces to LArSoft [4/5]

PetrilloAtWork:
Python interface update required some changes in the geometry infrastructure.
Part of that is in `icarusalg` and subject of SBNSoftware/icarusalg#93, and part is subject of this one.

Update to geometry infrastructure
----------------------------------

ICARUS geometry infrastructure inherits a convoluted design from when it was necessary to support two channel mappings, in a way that LArSoft did not facilitate.
Now LArSoft side has changed, leaving more flexibility for the channel mapping, and on the other end the need for the non-split-wire readout channel mapping has been beyond legacy since a while, and already effectively discontinued.
So I am removing one layer of indirection from the (v10) geometry infrastructure.
I am also renaming one of the services introduced in v10 from `IcarusWireReadout` to `WireReadoutICARUS`, since that is the most common pattern in ICARUS (exceptions still apply though — just one less with this PR).

Pull request
-------------

This pull request depends on SBNSoftware/sbnalg#1 and SBNSoftware/icarusalg#93 and must be merged at the same time as those two.
Reviewers:
 * @SFBayLaser as ICARUS software coordinator


SFBayLaser (approved):
Works as expected when  added on top of icaruscode v10_04_06.

PR #818, opened 2025-04-24
Removing bad channels from use in coherent noise corrections

SFBayLaser:
This update allows the ability to remove "bad" channels (from channels database) from consideration in computing coherent noise subtraction. 
This also adds the ability to reject channels that have too low a truncated rms, or too large, so they won't be used either. 
Both have fhicl parameters to turn on/off


gputnam (approved):
Approved. Discussions sorted out over slack.

PR #823, opened 2025-05-04
Update cafmakerjob_icarus_data.fcl

nathanielerowe:
Use new product labels. To be merged in alongside: https://github.com/SBNSoftware/sbncode/pull/523.


PetrilloAtWork (commented):
I have a suggestion for a different approach.

1. leave the module labels the standard ones (`bnbinfo`, `numiinfo`, etc.)
2. change the job configurations in `sbncode` so that the modules are called with matching names (`bnbinfo` etc.).

I can't see any reason for the `bnbinfo` label to be different between the two experiments, just like there is no reason for, say, CRT hit labels to be different.

leoaliaga:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_06_00 SBNSoftware/sbnalg@v10_06_00_01 SBNSoftware/sbncode@v10_06_00_01

PR #827, opened 2025-05-20
DNNROI GPU Config

nathanielerowe:
Add GPU config options for running on Polaris. Speeds up from 80s (DNN@FNAL, CPU) -> 25s (DNN@ANL, GPU) out of the box.


nathanielerowe:
@leoaliaga Do those CI checks always fail?

PR #830, opened 2025-05-22
Fix sim label for PDFastSim

mvicenzi:
SimPhotons should be generated from pre-SCE energy deposits.
It looks like this has not been the case ever since we switched to the refactored G4 steps.

I'm deeply confused by our fhicl structure.
Please advise if you think this should be overriden somewhere else.


PetrilloAtWork (approved):
Conceptually correct.
I haven't checked for typos though (so I assume this configuration was exercised at least once).

leoaliaga:
trigger build larsoft@v10_06_00 LArSoft/lar*@LARSOFT_SUITE_v10_06_00 SBNSoftware/sbnalg@v10_06_00_01 SBNSoftware/sbncode@v10_06_00_01

mvicenzi:
This change was included in #892

PR #831, opened 2025-06-02
Patches to 2025A Data Production

gputnam:
Collecting fixes and needed updates here:
-Use correct bad channel database tag

https://github.com/SBNSoftware/icaruscode/pull/827 is also needed.


SFBayLaser:
I am not aware that the proposed change to "long hits" has been fully vetted and approved. For example, have both the pandora and SPINE folks looked at the impact to the output of their respective reconstructions and signed off? 

This potentially has a major impact to the reconstruction for particular categories of tracks. While nobody has ever said this scheme was perfect it is currently the known devil and I have not seen anything showing either no impact to reconstruction or improvement. So far as I know it solves a computation problem for one track in one event in the single failing job amongst several.

jzennamo (approved):

PR #834, opened 2025-06-18
Fix ctoffset to match RawDigit/Deconvolved collection hit time

gputnam:
Set ctoffset to match peak time between rawdigit and deconvolved on collection plane in data.

See plot on right. The green line is the collection plane, after ctoffset is set correctly.
image


SFBayLaser (approved):
This value produced the green curve

leoaliaga:
trigger build larsoft@v10_06_00 LArSoft/lar*@LARSOFT_SUITE_v10_06_00 SBNSoftware/sbnalg@v10_06_00_01 SBNSoftware/sbncode@v10_06_00_01

PR #853, opened 2025-09-30
CRT Hit Truth Label Override

gputnam:
Depends on https://github.com/SBNSoftware/sbncode/pull/586.


francescopoppi (approved):
This label is fine, I approve this PR. Just one suggestion: would it be possible to add a comment which address what the shift is respect to?

PR #871, opened 2025-11-19
Trigger simulation modules now support input tags with process name

PetrilloAtWork:
These modules are special in that they compose the list of input tags from a base tag and an instance name from a list. The simplistic tag creation logic did not account for base input tags with process names.

Now the base tag can be specified as a full input tag and an exception is raised if there is a conflict between the input tag instance and the instance in the list (if both are not empty — I expect the instance name of the input tag to be always empty; otherwise, a new protocol needs to be developed). Both module label and process name are correctly propagated.

The fix has been applied to all modules known to suffer from the issue; several of them are not commonly used, and a couple are actually deprecated and should be removed.

This pull request is against the main code branch (`develop`).

Reviewers:
* @mvicenzi who reported the issue and needed a fix
* @rtriozzi as a person informed on trigger simulation

### DIFF
--- a/icaruscode/PMT/Trigger/LVDSgates_module.cc
+++ b/icaruscode/PMT/Trigger/LVDSgates_module.cc
@@ -879,7 +879,7 @@ art::InputTag icarus::trigger::LVDSgates::makeTag
     ? art::InputTag{ thresholdStr }
     : defModule.label().empty()
       ? throw (art::Exception(art::errors::Configuration)
-        << "Default module label (`TriggerGatesTag`) specified"
+        << "No default module label (`TriggerGatesTag`) specified"
            ", and it's needed for threshold '" << thresholdStr << "'.\n")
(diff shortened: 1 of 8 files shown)


rtriozzi (approved):
These changes make sense to me, thanks Gianluca!

mvicenzi (approved):
Fix works, so I approve!

jas1005:
trigger build larsoft@v10_24_00 LArSoft/lar*@LARSOFT_SUITE_v10_24_00 SBNSoftware/sbnalg@v10_24_00 SBNSoftware/sbnobj@v10_23_00 SBNSoftware/sbnanaobj@v10_23_00 SBNSoftware/sbndaq-artdaq-core@v1_10_06_01 SBNSoftware/sbncode@v10_24_00 SBNSoftware/icarusalg@v10_24_00 SBNSoftware/icarusutil@v10_24_00 SBNSoftware/icarus_signal_processing@v10_24_00

PR #875, opened 2025-12-04
Add CRT-PMT flash classification to TPC-PMT flash-matching producer

rtriozzi:
This PR (based off of `v10_04_04`) adds the flash classification from the CRT-PMT matching (via the `crtpmt` label) to the best-matched flash in the TPC-PMT barycenter flash-matching producer. This information can be propagated to the CAFs, and enables for using the CRT-PMT matching information at the slice-level in selections. Relevant PRs will follow.

It also adds an option (via `MinimizeZDistance`) to choose the flash that is best-matched to the TPC charge by minimizing only the distance along the longitudinal direction. By default, the code still chooses the best-matched flash by minimizing the longitudinal-vertical distance.

___
### Associated PRs

- https://github.com/SBNSoftware/sbnobj/pull/157
- https://github.com/SBNSoftware/sbnanaobj/pull/179
- https://github.com/SBNSoftware/sbncode/pull/613

___
### Review

Tagging for review @francescopoppi and @PetrilloAtWork as the CRT & light gurus. Thanks!

### DIFF
--- a/icaruscode/PMT/OpReco/TPCPMTBarycenterMatchProducer_module.cc
+++ b/icaruscode/PMT/OpReco/TPCPMTBarycenterMatchProducer_module.cc
@@ -25,6 +25,7 @@
 #include "art/Framework/Services/Registry/ServiceHandle.h"
 #include "art/Persistency/Common/PtrMaker.h"
 #include "canvas/Persistency/Common/FindOne.h"
+#include "canvas/Persistency/Common/FindOneP.h"
 #include "canvas/Persistency/Common/FindMany.h"
 #include "canvas/Persistency/Common/FindManyP.h"
 
@@ -50,6 +51,8 @@
 #include "lardataobj/AnalysisBase/T0.h"
 #include "lardataobj/RawData/TriggerData.h"
 #include "sbnobj/Common/Reco/TPCPMTBarycenterMatch.h"
+#include "icaruscode/CRT/CRTUtils/CRTPMTMatchingUtils.h"
+#include "sbnobj/Common/CRT/CRTPMTMatching.hh"
 
 //ROOT includes
 #include "TTree.h"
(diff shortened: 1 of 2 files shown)


icaruscode/PMT/OpReco/TPCPMTBarycenterMatchProducer_module.cc
```diff
@@ -543,15 +552,23 @@ void TPCPMTBarycenterMatchProducer::produce(art::Event& e)
...
+          }
+        }
+        else {
+          thisFlashCenterY = flash.YCenter();
+          thisFlashCenterZ = flash.ZCenter();
+          thisDistance = std::hypot( (thisFlashCenterY - fChargeCenterY), (thisFlashCenterZ - fChargeCenterZ) );
+          if ( thisDistance < minDistance ) {
+            minDistance = thisDistance;
+            matchIndex = m;
+          }
```
PetrilloAtWork: Extract the common code paths from the `if` branches:
```suggestion
        // Find index of flash that minimizes barycenter distance...
        double const thisFlashCenterZ = flash.ZCenter();
        if ( fMinimizeZDistance ) { // ... along Z
          thisDistance = std::abs(thisFlashCenterZ - fChargeCenterZ);
        }
        else { // ... in the YZ plane
          thisDistance = std::hypot( (flash.YCenter() - fChargeCenterY), (thisFlashCenterZ - fChargeCenterZ) );
        }
        if ( thisDistance < minDistance ) {
          minDistance = thisDistance;
          matchIndex = m;
        }
```
For example, this helps noticing that the minimisation code is the same, and only the metric changes.
Note that the suggestion also requires removing the definition of `thisFlashCenterZ` and `thisFlashCenterZ` from above (rule: declare as close as possible to where it is used).
Even more compact, but less readable:
```suggestion
        // Find index of flash that minimizes barycenter distance...
        double const thisFlashCenterZ = flash.ZCenter();
        double const thisDistance = fMinimizeZDistance
          ? std::abs(thisFlashCenterZ - fChargeCenterZ); // ... along Z
          : std::hypot( (flash.YCenter() - fChargeCenterY), (thisFlashCenterZ - fChargeCenterZ) ); // ... in the YZ plane
        if ( thisDistance < minDistance ) {
          minDistance = thisDistance;
          matchIndex = m;
        }
```

icaruscode/PMT/OpReco/TPCPMTBarycenterMatchProducer_module.cc
```diff
@@ -240,22 +243,24 @@ class TPCPMTBarycenterMatchProducer : public art::EDProducer {
   double CentroidOverlap(double center1, double center2, double width1, double width2) const; ///< Return overlap between charge and light centroids OR distance apart if no overlap
   double CalculateAsymmetry(art::Ptr<recob::OpFlash> flash, int cryo);                        ///< Return the east-west asymmetry of PEs in a given OpFlash
   void updateChargeVars(double sumCharge, TVector3 const& sumPos, TVector3 const& sumPosSqr, std::array<double, 2> const& triggerFlashCenter); ///< Update slice-level data members with charge and trigger match info
-  void updateFlashVars(art::Ptr<recob::OpFlash> flash, double firstHit);                      ///< Update slice-level data members with best match info
-  void updateMatchInfo(sbn::TPCPMTBarycenterMatch& matchInfo);                                      ///< Update match product with slice-level data members
- 
+  void updateFlashVars(art::Ptr<recob::OpFlash> flash, double firstHit, int matchedFlashClassification); ///< Update slice-level data members with best match info
```
PetrilloAtWork: I _suggest_ to preserve the original data type as long as possible:
```suggestion
  void updateFlashVars(art::Ptr<recob::OpFlash> flash, double firstHit, sbn::crt::MatchType matchedFlashClassification); ///< Update slice-level data members with best match info
```

icaruscode/PMT/OpReco/TPCPMTBarycenterMatchProducer_module.cc
```diff
@@ -306,10 +312,12 @@ TPCPMTBarycenterMatchProducer::TPCPMTBarycenterMatchProducer(fhicl::ParameterSet
   fOpFlashLabel(p.get<std::string>("OpFlashLabel")),
   fPandoraLabel(p.get<std::string>("PandoraLabel")),
   fTriggerLabel(p.get<std::string>("TriggerLabel")),
+  fCRTPMTMatchingLabel(p.get<std::string>("CRTPMTMatchingLabel")),
```
PetrilloAtWork: I _recommend_ to put the new feature as optional, and _suggest_ that the default behaviour disables it to make this not breaking.
A few changes are needed for that to happen.

PetrilloAtWork (changes requested):
Two topics for your consideration:
 * _recommendation_ to make the extraction of CRT matching optional (e.g. by specifying an empty CRT/PMT matching label): let's not force CRT on people who do not care about it (because they just don't care, because it takes too long to execute, because it was dropped from the input, because there was no CRT component in that data run, ...)
      * _suggestion_ that it be disabled by default
 * _suggestion_ to preserve the `MatchType` type in the data product, and in most of the code

Also a couple of code style notes.

icaruscode/PMT/OpReco/TPCPMTBarycenterMatchProducer_module.cc
rtriozzi: This makes sense -- implemented. Thanks!

jas1005:
@francescopoppi and @PetrilloAtWork

I'm going through old PRs and it looks like this one never got approved. However, there seems to be a lot of substantial progress made in this PR. Do y'all have time to review this PR in its current form or should I ask someone else?

Gianluca, I hear you're an often-requested reviewer. I'm going to try to keep your workload light. But I think you are well-positioned to review this specific code given your expertise and your already-proposed changes.

francescopoppi:
I expressed my concerns to Riccardo on Slack. I feel like once this is approved by @PetrilloAtWork it can be merged. I would suggest a bump up in priority list.

PR #873, opened 2025-12-02
Update CI FHiCLs in production branch

vitodb:
This PR is to update the set of FHiCL files used by the CI for the three MC workflow we currently test with the CI, these are:
- single
- nucosmics
- intimecosmic
This also update some standard FHiCL to update them with current code.

With this PR we should be able to run regression test for the production branch.

PR #896, opened 2026-04-17
Updated Continuous Integration DetSim configurations to refactored G4

PetrilloAtWork:
Fixes the broken DetSim stage of the C.I. tests.
This change was tested only on the single particle tests; the other two test workflow are (almost) identical though.

This pull request is against `release/SBN2025A` branch and it was developed against `v10_06_00_06p04` tag.
**Release managers: please duplicate, cherry-pick and change target branch as needed.**

Last force-push: I adopted the changes in `develop`. Still reluctant to just cherry-pick the fix in `develop` (308484a6) because of a statement about a discontinued parameter in `MergeSimSources` which [does not seem to have actually been discontinued](https://github.com/LArSoft/larsim/blob/4e1000511eaaad4e3287d44f2786ab7596cbf080/larsim/MergeSimSources/MergeSimSources_module.cc#L88). @vitodb may comment on that.

### DIFF
--- a/test/ci/icarus_ci_intimecosmic_detsim_quick_test_icaruscode.fcl
+++ b/test/ci/icarus_ci_intimecosmic_detsim_quick_test_icaruscode.fcl
@@ -1,1 +1,1 @@
-#include "detsim_2d_icarus.fcl"
+#include "standard_mc_all_detsim_icarus.fcl"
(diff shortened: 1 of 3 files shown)


jas1005:
I got turned around with what CI work made it into this production branch, release/SBN2025A, versus develop. PR #873 makes all the changes this PR does and some others. I'm leaving this PR open so Vito can comment on Gianluca's question above. But we can likely close this PR since it duplicates PR #873. My apologies. I closed PR #873 in the past because I did not think it was necessary to have this work live on release/SBN2025A alongside develop.

PR #904, opened 2026-06-02
Updates after upstream drop of legacy LArG4

PetrilloAtWork:
The changes proposed here are aimed mainly to pass the FHiCL-parsing unit test.
A half-hearted attempt to correctly update the configuration is sometimes tried, but no test was done except in one case, where the test _failed_.

List of changes:
 * `larg4_icarus.fcl` now refers to definitions in `larg4_icarus_defs.fcl` instead of having the same definitions inline. This was tested by verifying that the dumps of the configuration before and after the change are identical.
 * `photonlibrary_builder_icarus.fcl` is a failed attempt to update the photon library creation: the resulting tree is empty (maybe because the physics list is specified in a different way now).
 * There are in-time and dirt configurations that take some work to be fixed. They have been "archived" in `fcl/archives/LegacyLArG4` as they were, until somebody cares and fixes them.
 * Many configurations included `largeantmodules.fcl` and its ICARUS twin just because it was fashionable. The inclusion has been removed, and they still parse.


fcl/g4/larg4_icarus.fcl
```diff
@@ -30,69 +19,28 @@ physics:
...
-    genericcrt: @local::icarus_genericCRT
-
-    # simplemerge for ML
-    simplemerge: @local::simplemerge
+    @table::icarus_larg4_standard_producers  # from larg4_icarus_defs.fcl
 
   }
 
   # All producers and filters modules for this path, order matters
   simulate: [ rns
```
jas1005, line 27: Should we leave 'rns' here since the other entries in simulate were deleted?

icaruscode/PMT/prova_source.fcl
```diff
@@ -140,3 +142,6 @@ physics.analyzers.pmtresponse.MakeAllPhotonsTree: true
 physics.analyzers.pmtresponse.MakeDetectedPhotonsTree: true
 physics.analyzers.pmtresponse.MakeSimPhotonssTree: true
 physics.analyzers.pmtresponse.MakeEventsTree: true
+
+# new LArG4 optical products are produced by `pdfastsim`, not `largeant`
+physics.analyzers.pmtresponse.InputModule: [ "pdfastsim" ]
```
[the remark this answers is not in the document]
PetrilloAtWork: `SimPhotonCounter` module [accepts both forms](https://github.com/LArSoft/larana/blob/f6699d338b3d1e0a83e3b185a1868cf80f8dc7e4/larana/OpticalDetector/SimPhotonCounter_module.cc#L175-L180).
The sequence form is the preferred one, the other being for backward-compatibility.

jas1005 (commented):
Everything looks good to me, but I'm really only able to give approval at the level of how Gianluca has proposed moving fcl file things around. I don't know anything about the underlying configurations and how they manifest physically. Barring the syntax comment I made and GitHub Copilot's two comments, I'm ready to approve.

cerati (approved):
Looks good me, with definitions moved upstream to a new fcl file and the other files updated accordingly. I especially enjoyed Gianluca's bullying Copilot.

fcl/configurations/larg4_icarus_defs.fcl
```diff
@@ -0,0 +1,146 @@
...
+# ---  ICARUS "standard" configuration
+# --- 
+#
+# (note overrides at the bottom)
+#
+# The service sets include:
+#  - `icarus_larg4_services`: only the services needed only by LArG4
+#  - `icarus_all_larg4_services`: all services needed for a job including LArG4
+#
+# Note that currently `icarus_g4_services` is directly taken from `larg4_services_icarus.fcl`.
```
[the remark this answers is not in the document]
PetrilloAtWork: Fixed.

fcl/g4/larg4_icarus.fcl
PetrilloAtWork: I think we should, since `rns` is in the original `larg4_icarus.fcl` (i.e. if I removed it, it would be different from the original).
Should it be included in `icarus_larg4_standard_path` instead? My opinion is it should not, since it is not related to LArG4 execution. For example, it is present in many configurations¹ and it is custom to set it as first module. If we put it in the sequence, other configurations that have `[ rns, generator, @sequence::icarus_larg4_standard_path ]` would list `rns` twice (which I _believe_ has no actual consequence beside being not very elegant).

--- 
[1] For no good reason, in my opinion: I was never able to take advantage of it.

PetrilloAtWork:
To be true, a lot of this PR was orchestrated with CoPilot.
In the end I think it was slower, but it may have avoided some mistake.

fcl/g4/larg4_icarus.fcl
jas1005: Makes sense to me! Formally approving now.

jas1005 (approved):

PetrilloAtWork:
Superseded by #906.

PR #907, opened 2026-06-18
Add CMakePresets.json needed by spack mpd 

gartung:
CMake defines need to be set in CMakePresets.json so spack mpd can use them for development builds.


gartung:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_21_00

PR #911, opened 2026-07-09
Updates to Enable ICARUS WireMod Closure

gputnam:


francescopoppi (approved):
I have stared at this code for weeks, I played with it extensively and I think we added all possible existing diagnostic tools. In this review I went through everything again and I do not see residual issues. One thing I tried to make sure is that all the different fcls are pointing to the correct configuration.
I will leave 3 open taks with this PR which are implicitly assigned to myself:
1) huge cleanup of the fcls, right now 80% of the newly added fcls were used to debug wireMod. I think they will be very useful in the near future, hence they do not need to be removed, but in the future a cleanup is needed, e.g. remove the unused ones, remove duplicated config declarations.
2) the YZ corrections have not been tested, code is there, ready to apply them, but it needs to be double checked once we have the respective ratios.
3) UseChannelROIMode is to be favored wrt to what it is used in the standard fcls. Results should be identical, but it is favored as it removes some iffy conversions along the procedure. FCLs will be updated after extensive closure tests will confirm the similar results.

A note for the release manager and the readers: all these fixes are related to WireMod and WireMod only. They will not impact other processing stages, unless they actively call for wire modification.

Approved by me

SFBayLaser:
The companion sbn merge was done and sbncode is available so will merge this to the SBN2025A branch and start the process...

PR #916, opened 2026-07-22
Temporarily disabling DNN ROI to enable CI tests in short term

SFBayLaser:
This PR is meant to temporarily disable the DNN ROI finding in both data and MC. At issue is that the latest versions of the Wirecell toolkit, including the version that gives us the updated yz simulation, are crashing in the DNN ROI finding and this is preventing us from debugging the rest of the CI test system. We don't know when we will see an updated wirecell toolkit so hopefully "unsticking" the problem with this temporary update.

### DIFF
--- a/fcl/reco/Stage0/data/stage0_run2_wcdnn_icarus.fcl
+++ b/fcl/reco/Stage0/data/stage0_run2_wcdnn_icarus.fcl
@@ -1,3 +1,3 @@
 #include "stage0_run2_wc_icarus.fcl"
 
-#include "enable_dnn_sp.fcl"
+#####include "enable_dnn_sp.fcl"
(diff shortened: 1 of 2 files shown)


SFBayLaser:
I am not sure why this is referencing the develop branch as I had explicitly checked out v10_20_09p02... I will try to change the base.

nvmnd... can't base to a tag...

SFBayLaser:
trigger build

cerati (approved):
Approving, assuming this is a temporary (<1 month) patch. If this becomes more permanent, then I'd prefer the changes go into standard_mc_all_stage0_icarus.fcl and standard_data_all_stage0_icarus.fcl.

SFBayLaser:
Closing since there is now a wirecell fix that should become available soon.

PR #919, opened 2026-08-11
fhicl file to run purity monitor for keep up processing

SFBayLaser:
We would like to pull back the keep up processing to only run the code necessary to maintain the argon purity measurement. This fhicl file, produced and tested by Christan Farnese will do that job.


SFBayLaser:
Uh... was not intending to change 177 files, only add 1 so clearly this PR is in error.

PR #924, opened 2026-09-11
Update requested resource for quick CI tests only

jas1005:
Upper/Lower bounds updated via resource warnings in Jenkins CI dashboard. New reference files introduced as of icarus_ci build 15201, and these bounds were tuned via builds 15211, 15218, and 15222.


cerati (approved):
I do not see anything wrong, and the change in requirements seem reasonable. Vito might have more insights. Also, I am not sure what the Copilot comment means. But as far as I am concerned this is approved

